Issue #4, opened 2017-02-11
No matter what tags I'm looking at, I get -> Too few images, aborting message


LarryKanionga:
Using the tester.py provided in instaPy-master folder and no matter what tags I'm looking at, I'm getting following error output:

```
$ python tester.py 
It is recommended to install PIL/Pillow with the desired image format support so that image resizing to the correct dimesions will be handled for you. If using pip, try "pip install Pillow"
Logged in successfully!
Tag [1/1]
--> veganfoodshare
Too few images, aborting

Session ended
-------------
```

Am I missing something?

timgrossmann:
@diegoloop Hey, I'm sorry you are encountering this problem.
I also had the problem and got it working again. (I haven't updated it on github yet).

**Just edit "instapy.py" and add the line**

```python
self.browser.implicitly_wait(10)
```
**to the constructor.
If it stills says 'Too few images', just increase the number of seconds.**

> This probably comes from your internet connection and the speed the instagram page is loading

##### Also, please make sure to Star the Repo if you're using my script

timgrossmann:
@diegoloop Also, better don't ignore the **recommendation** to install PIL
```
pip3.5 install pillow
```

LarryKanionga:
Hi @timgrossmann thanks for reply :)

So in the **InstaPy-master/instapy/instapy.py** file (please note that I didn't find the instep.py as you mentioned) I added the following, but still not working:

```
class InstaPy:
  """Class to be instantiated to use the script"""
  def __init__(self, username=None, password=None):
    self.browser = webdriver.Chrome('./assets/chromedriver')
    self.browser.implicitly_wait(30) # new line. I played around from 10 to 100 
```
I also installed pillow

timgrossmann:
Hey @diegoloop ,
Sorry, I changed the "instep" to "instapy", sorry for the typo. Thank you for mentioning!

I'm currently encountering some of the same problems. I'll get back to you (and push it to master) once I found the Bug 👍 

Sorry for the Problems.

timgrossmann:
@diegoloop Just saw, that they changed their pagelayout a bit. (I already fixed that on my local repo, but haven't updated it only, yet)

I'll push it to github this evening (~8pm GMT+1)

timgrossmann:
Hey there @diegoloop. 
I updated the code base to the version which works (at least on my system).
I hope it works for you, too.

If not, feel free to contact me again.
Best Regards

[...]

PR #5, opened 2017-02-13
Bug fix

timgrossmann:
#### Updated to current version of the Instagram Page
- Added implicit wait, to cover the problem with "Too few images"
- Preventing "Friends" to be unfollowed if unfollow is used
- Added logging to a file
- Added File with unicode emojis that can be used
- improved quickstart.py file


timgrossmann:
Should fix Issue #4

PR #20, opened 2017-04-04
Add base setup.py

seemethere:
Adds a setup.py script so that users can either `pip install .` or do a `python setup.py install`


timgrossmann:
Awesome!, thank you! 👍

PR #21, opened 2017-04-05
Fixed like, comment and follow references

eduardocalazansjr:
Instagram changes the class of elements for some users/locations.

I’ve changed the like, follow and comment references for XPATH, instead
of Class.

Also added coding header on comment_util for avoiding errors when user
comment contains special characters/accents.


timgrossmann:
Looks good, Thank you!

Issue #28, opened 2017-04-05
Make InstaPy Language Independant


timgrossmann:
#### Right now, InstaPy only works if your Instagram language is english. Make sure it works with any language (or at least, most)

- [x] change check if already likes from "Like" to something more generic
- [x] make sure there are no other areas of language problems

eduardocalazansjr:
The comment is working perfectly because of the text inside the field.

Instead of turning everything to an language independent solution, why can we just use the chromedriver locale flag pointing to the en-US language?

Check my comment under https://github.com/timgrossmann/InstaPy/pull/31#issuecomment-292020518

PR #31, opened 2017-04-05
Multilanguage Like

eduardocalazansjr:
I've tested with multiple languages, resolutions, users and locations
and seems that the coreSpriteLikeHeart class never changes.
When the image is not liked yet, the class is coreSpriteLikeHeartOpen,
when the image is already liked the class changes to
coreSpriteLikeHeartFull.
This should fix the language/locale issue with the like function, but
the comment function must be reviewed.

### DIFF
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -68,19 +68,20 @@ def check_link(browser, link, dont_like, ignore_if_contains, username):
 
 def like_image(browser):
   """Likes the browser opened image"""
-  link_elem = browser.find_element_by_xpath('//article/div[2]/section[1]/a[1]/span[1]')
+  like_elem = browser.find_elements_by_xpath("//span[contains(@class, 'coreSpriteLikeHeartOpen')]")
+  liked_elem = browser.find_elements_by_xpath("//span[contains(@class, 'coreSpriteLikeHeartFull')]")
 
-  span_elem_text = link_elem.text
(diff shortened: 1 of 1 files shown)


timgrossmann:
fixes part of #28

timgrossmann:
@vlntdds Just a quick question, how do you test stuff like different languages and locations?

eduardocalazansjr:
@timgrossmann I have multiple machines with different regional/locale settings on OS (tested on Windows and OSx)

But you can try adding an argument to chromedriver and call chrome with another locale config:

`options.add_argument('--lang=es')`
[ref](https://developer.chrome.com/extensions/i18n#locales-testing)

timgrossmann:
@vlntdds Oh, wow that's amazing!
What's the reason you have them?

Really glad you're interested in committing to this project.

eduardocalazansjr:
@timgrossmann in my work i use windows (en-US language w/ Brazil locale), at home i use a mix of OSx (en-US language w/ en-US locale + pt-BR language w/ Brazil locale).
Also i have a bunch of VMs on Azure with a lot of mess setups, hahah.

I dive into Selenium documentation and realize that the following line will set the chromedriver locale and change instagram language

`chrome_options.add_experimental_option('prefs', {'intl.accept_languages': 'pt-br'})`

Give it a try.

PR #32, opened 2017-04-05
add Docker support :)

stephane-archer:
Added Dockerfile with default linux64 chromedriver

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM ubuntu
+
+RUN apt-get update
+RUN apt-get install -y python wget unzip python-pip
+
+RUN mkdir /InstaPy
+WORKDIR /InstaPy
+COPY . .
+
+WORKDIR /InstaPy/assets
+RUN wget https://chromedriver.storage.googleapis.com/2.29/chromedriver_linux64.zip
+RUN unzip -o chromedriver_linux64.zip
+
+RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - 
+RUN sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google-chrome.list'
+RUN apt-get update 
+RUN apt-get install -y google-chrome-stable
+
+WORKDIR /InstaPy
+RUN pip install .
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,9 +14,8 @@ RUN mkdir /InstaPy
 WORKDIR /InstaPy
 COPY . .
 
-WORKDIR /InstaPy/assets
-RUN wget https://chromedriver.storage.googleapis.com/2.29/chromedriver_linux64.zip
-RUN unzip -o chromedriver_linux64.zip
+RUN wget https://chromedriver.storage.googleapis.com/2.29/chromedriver_linux64.zip \
+    && unzip -o -d /InstaPy/assets chromedriver_linux64.zip \
+    && rm chromedriver_linux64.zip
 
-WORKDIR /InstaPy
 RUN pip install .
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,14 @@
-FROM ubuntu
+FROM ubuntu:16.04
+MAINTAINER ARCHER Stéphane <stephane.archer@epita.fr>
 
-RUN apt-get update
-RUN apt-get install -y python wget unzip python-pip
+RUN apt-get update \
+    && apt-get install -y \
+        chromium-browser \
+        python \
+        python-pip \
+        unzip \
+        wget \
+    && rm -rf /var/lib/apt/lists/*
 
 RUN mkdir /InstaPy
 WORKDIR /InstaPy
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM ubuntu
+
```
jaycegregg0: Add a MAINTAINER line with your name and email.

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+COPY . .
+
+WORKDIR /InstaPy/assets
+RUN wget https://chromedriver.storage.googleapis.com/2.29/chromedriver_linux64.zip
+RUN unzip -o chromedriver_linux64.zip
+
+RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - 
+RUN sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google-chrome.list'
+RUN apt-get update 
+RUN apt-get install -y google-chrome-stable
```
jaycegregg0: Since InstaPy doesn't use pepperflash, the Hangouts plugin, or other video codecs there's no reason to use the proprietary Chrome over the free Chromium.

Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM ubuntu
+
+RUN apt-get update
+RUN apt-get install -y python wget unzip python-pip
+
+RUN mkdir /InstaPy
+WORKDIR /InstaPy
```
jaycegregg0: Changing WORKDIR numerous times in a Dockerfile creates substantial layer bloat so try and reduce how many times you change directory. Same goes for switching users.

Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM ubuntu
```
jaycegregg0: Specify a specific version of the base image. Use something like ubuntu:xenial or ubuntu:16.04.

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+RUN apt-get update
```
jaycegregg0: Optimize your layer usage with something like the following:

```bash
RUN apt-get update && apt-get install -yqq --no-install-recommends \
  python \
  python-pip \
  unzip \
  wget \
&& rm -rf /var/lib/apt/lists/*
```

Also, putting your dependencies into alphabetical order makes it easier to scan them and add/remove from the list.

jaycegregg0 (changes requested):
Great way to pitch in by Dockerizing the project Stephane!

Hope you don't mind some suggestions on how to take your Dockerfile to the next level by implementing some of Docker's official best practices.

https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/

stephane-archer:
Dear @jaycegregg0,
I'm really happy to get feedback and advice :)
I will made these changes :)

Dockerfile
stephane-archer: What is the point of using -yqq? Just reduce the log? :) if there is error we need log for the user so is it ok? :)

stephane-archer:
Dear @jaycegregg0,
I hope you will love my changes
If you have any feedback please tell me :)

timgrossmann:
@stephane-archer First of all, thank you for participating!

Could you do one more thing?
- Add a comment a description to the PR on what you did. (It could be as short as "Added Dockerfile with default linux64 chromedriver".)

Did you test if this works?
I saw you only install a few dependencies, does it work with the pyvirutaldisplay?
Had to do quite some tinkering on my server to get it running: https://medium.com/@hoppy/how-to-test-or-scrape-javascript-rendered-websites-with-python-selenium-a-beginner-step-by-c137892216aa

stephane-archer:
Are you already using in your script pyvirutaldisplay?
Actually I'm using VNC to launch the script because the script need to launch Chrome.

I will have a look to: http://fabiorehm.com/blog/2014/09/11/running-gui-apps-with-docker/
I hope to improve the process :)

timgrossmann:
@stephane-archer yes, when running on e.g. a server, pyvirutaldisplay is used.

I don't have enough experience on how it works with docker.

Dockerfile
jaycegregg0: -y to not require interaction to confirm the install and -qq to quiet the output. If you buy into the UNIX philosophy as described in TAOUP silence should be preferred and output should only happen when there's something exceptional going on or when the user requests verbose output.

Log size usually isn't an issue with containers and VMs since they are starting to log to the stderr console rather than a file in many environments.

There's no problem with dropping the -qq though if you want to see the output but keep in mind that when using Docker via a CI/CD system like Travis, Jenkins, etc that all of the extraneous output can make quickly scanning the logs for the build or deployment more tedious.

jaycegregg0:
It works pretty much the same way with Docker. pyvirtualdisplay or xvfb are usually how you deal with the GUI bit when working on headless CI/CD environments.

@stephane-archer You could easily use xvfb via [this Docker image](https://hub.docker.com/r/markadams/chromium-xvfb/).

timgrossmann:
@jaycegregg0 Yes, I just realised, that you can go through the whole process just as if it was a Droplet on DO. 
Let's see if stephane will take care of it, else I'll add the Dockerfile soon

stephane-archer:
Sorry I was busy these few days :/
Hope I can help you more next time :)

timgrossmann:
@stephane-archer You helped a lot!
The only thing is, you're not listed as collaborator by now... 
Maybe you could add another PR 😉

stephane-archer:
I don't care about collaborator, just want to contribute to cool project :D 
Probably soon ;)

Issue #33, opened 2017-04-05
Invalid Page Log


stephane-archer:
I have the following log message:
`Invalid Page: Message: no such element: Unable to locate element: {"method":"xpath","selector":"//article/div[2]/div[1]/ul[1]/li[1]"}`
Every time a page is loaded.
I have a small screen, maybe the "phone version" of the page is loaded?
Any idea?

eduardocalazansjr:
What OS, chromedriver version and resolution are you using?

stephane-archer:
Ubuntu 16.04, ChromeDriver 2.29 :)

mateuspontes:
```
Invalid Page: Message: no such element: Unable to locate element: {"method":"xpath","selector":"//article/div[2]/div[1]/ul[1]/li[1]"}
  (Session info: chrome=57.0.2987.133)
  (Driver info: chromedriver=2.29.461585 (0be2cd95f834e9ee7c46bcc7cf405b483f5ae83b),platform=Mac OS X 10.12.4 x86_64)
```

mateuspontes:
Instagram has renamed attrs, I fixed using:

```
user_div = browser.find_element_by_xpath('//*[@id="react-root"]/section/main/div/div/article/header')
```

and for like
```
def like_image(browser):
  """Likes the browser opened image"""
  like_elem = browser.find_elements_by_xpath("//span[contains(@class, 'coreSpriteHeartOpen')]")
  liked_elem = browser.find_elements_by_xpath("//span[contains(@class, 'coreSpriteHeartFull')]")
```

eduardocalazansjr:
@mateuspontes 

this is not the right user_div for the mostly of accounts, some accounts have different layouts.

there are two variants of user_div
`//article/div[2]/div[1]/ul[1]/li[1]`
and for some users
`//article/div[2]/ul[1]/li[1]`

I'm gonna try to find the pattern/relation between user accounts and the elements class.

eduardocalazansjr:
@mateuspontes could you open [this photo](https://www.instagram.com/p/BSUNC_ugUA5F-oCjArV_iczK5DwU_TSQLFTdAE0/) on chrome, open the JS Console and show me the value of `window._sharedData.qe.ebd.p.use_new_styles` variable?

mateuspontes:
`undefined`

eduardocalazansjr:
@mateuspontes for me it has the "true" value, did you have the `window._sharedData ` element in this page?

[...]

PR #34, opened 2017-04-06
Make like_util work independent of 'use_new_styles' 

eduardocalazansjr:
With the #33 I realized that there’s a JS variable that contains all
data that we need and that the UI of Instagram Web differs from users
(use_new_styles flag).

Changed the XPATH/DOM references for JS functions to retrieve the data,
independent of ‘use_new_styles’.

Also made a change on InstaPy.py that forces the Chrome to run with the
“en-US” location to avoid the Locale/Language problem with the like
button and the comment placeholder.


timgrossmann:
Perfect, thank you!
Fixes #28

PR #39, opened 2017-04-06
Check if the post/comment exists

eduardocalazansjr:
Just realized that if the user left the caption empty and after the
post date write an comment, it doesn't count as a caption. Fixed that
by getting the first comment from the post.

Also added a check if “PostPage” variable exists in “check_link”.


timgrossmann:
@vlntdds Looks good, thank you!
Could you please check which of the open issues are solved by this and close those?

PR #40, opened 2017-04-06
fixed problem with no follower count on completely new account

timgrossmann:
> Fixes the issue with crashing when the script is run with an entirely new account where there is no follower amount on the profile page

### DIFF
--- a/instapy/print_log_writer.py
+++ b/instapy/print_log_writer.py
@@ -1,14 +1,19 @@
 """Module only used to log the number of followers to a file"""
 from datetime import datetime
+from selenium.common.exceptions import NoSuchElementException
 
 def log_follower_num(browser, username):
   """Prints and logs the current number of followers to
   a seperate file"""
   browser.get('https://www.instagram.com/' + username)
+  follower_num = 0
 
-  follower_elem = browser.find_element_by_xpath\
-    ('//a[@href="/' + username.lower()  + '/followers/"]')
+  try:
+    follower_num = browser.find_element_by_xpath\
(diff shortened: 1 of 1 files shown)


eduardocalazansjr:
If you get the followers number from `window._sharedData.entry_data.ProfilePage[0].user.followed_by.count`, the issue is gone and we avoid the DOM differences between profiles/accounts.

timgrossmann:
@vlntdds Thank you, i'll change that.
I'll really have to look into what this _sharedData is...

Hmm, well when trying this in the console I get:
```
VM219:1 Uncaught TypeError: Cannot read property '0' of undefined
    at <anonymous>:1:42
```

eduardocalazansjr:
@timgrossmann you have to use it on your profile page, i guess.

timgrossmann:
@vlntdds Hmm I tried it in chrome and safari, but both times with an account without any posts, that might be the problem. Does it work for you?

timgrossmann:
@vlntdds Could you test it with this print_log_writer again, please?

eduardocalazansjr:
@timgrossmann just realized that if you click on the "person" icon on the top to go to your profile, the entry_data.ProfilePage doesn't exist, but entry_data.FeedPage exists, but doesn't hold the followers count.
If you go to the profile url, the entry_data.ProfilePage exists.

It seems to work on a brand new (and empty) account and for my principal account.

timgrossmann:
@vlntdds Ahh alright, now I got it.
Thanks for checking it out. Works pretty good now!

Issue #42, opened 2017-04-07
Element is not clickable at location


razvanilin:
I get this error and the script crashes after about 5-10 likes:

```
selenium.common.exceptions.WebDriverException: Message: unknown error: Element <span class="_soakw coreSprite...HeartOpen">Like</span> is not clickable at point (445, 463). Other element would receive the click: <div class="_mghjk">...</div>
  (Session info: chrome=57.0.2987.133)
  (Driver info: chromedriver=2.29.461571 (8a88bbe0775e2a23afda0ceaf2ef7ee74e822cc5),platform=Linux 4.4.0-72-generic x86_64)
```

timgrossmann:
@razvanilin I also got this error a few times already... I think this is due to the footer bar in the Instagram webpage :🙁
We'll see if we can fix this.

Thank you for reporting!

razvanilin:
At first I thought it was because I was using a business account. I then reverted back to personal account and it still happens. So business accounts are not causing this.

timgrossmann:
@razvanilin Thank you, I really think it's the footer bar in Instagram, since Selenium just moves to the element and clicks the mentioned element.

caltv:
yes I'm seeing the same

selenium.common.exceptions.WebDriverException: Message: unknown error: Element <span class="_soakw coreSpriteHeartOpen">...</span> is not clickable at point (684, 678). Other element would receive the click: <div class="_mghjk">...</div>

caltv:
Update: it seems if you manually click that footer at the beginning to dismiss it, the script continues successfully.

razvanilin:
I tried that in the beginning but it didn't work for me. What finally worked is to switch my account back to business account.

eduardocalazansjr:
@caltv what footer is that? i don't have it.
could you post a screenshot of the footer and tell me what resolution are you using?

caltv:
<img width="882" alt="screen shot 2017-04-07 at 15 03 21" src="https://cloud.githubusercontent.com/assets/219204/24803510/604d53de-1ba3-11e7-934f-6ee997d6f95b.png">

caltv:
I'm on a retina 2880 x 1800 res

timgrossmann:
Fixed by #47

PR #47, opened 2017-04-07
Fix like when the "Get the App" bar is visible

eduardocalazansjr:
Instead of calling Selenium method Click, get the Class of the Like
Button and do a Click() with Javascript.

If there’s an element above the Like Button, Selenium doesn’t hit it
and throw an error, JS don’t.

### DIFF
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -82,7 +82,7 @@ def like_image(browser):
   liked_elem = browser.find_elements_by_xpath("//a[@role = 'button']/span[text()='Unlike']")
 
   if len(like_elem) == 1:
-    like_elem[0].click()
+    browser.execute_script("document.getElementsByClassName('" + like_elem[0].get_attribute("class") + "')[0].click()")
     print('--> Image Liked!')
     sleep(2)
     return True


eduardocalazansjr:
Fixes #42

timgrossmann:
@vlntdds Insanely smart fix for this problem!
Thank you.

blesson3:
This change was overwritten in the current `master` commit 7ab3e0042c3c8ed9c7f5abad9d201ed55b972b54 (file w/ line: https://github.com/timgrossmann/InstaPy/blob/master/instapy/like_util.py#L538).

I suspect the overwrite is causing issues that this PR solved to come back up.

timgrossmann:
@blesson3 Thank you for pointing out!
Can you double check if it works again now?

blesson3:
@timgrossmann I ran the script for a while and it seemed to fix the problem with clicking the like button! 

I see that you overwrote that line above to what the PR changed it to, see below why I don't think the solution is that easy in this case. https://github.com/timgrossmann/InstaPy/commit/ad6e579b56b464698cd13b01faca18291f639659
(https://github.com/timgrossmann/InstaPy/blob/master/instapy/like_util.py#L538)

Now I'm running into another issue similar, but different:
```
except: Message: unknown error: Element <a href="/explore/tags/sun/?max_id=1691136460033415146" class="_1cr2e _epyes">...</a> is not clickable at point (381, 340). Other element would receive the click: <div class="_eyflt">...</div>
  (Session info: chrome=63.0.3239.132)
  (Driver info: chromedriver=2.34.522913 (36222509aa6e819815938cbf2709b4849735537c),platform=Linux 4.4.0-53-generic x86_64)
```

See the following issue for discussion: https://github.com/timgrossmann/InstaPy/issues/1232

PR #60, opened 2017-04-09
Introduced option to ignore pictures from certain users and include owner comments when checking for tags

pekapa:
Usage:
    .set_ignore_users([usernames])

Also changed the behaviour when checking for ignores based on keywords, now it includes all the comments made by the owner (sometimes the owner place tags in their first comment, not necessarily the first comment of the picture)

Finally, added some verbosity to track why an image has been flagged as inappropriate (which tag was responsible or the user name if it’s blacklisted).

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -129,6 +129,18 @@ class InstaPy:
 
     return self
 
+  def set_ignore_users(self, users=None):
+    """Changes the possible restriction to users, if user who postes
+    is one of this, the image won't be liked"""
+    if self.aborting:
+      return self
+
+    if users is None:
+      users = []
+    self.ignore_users = users
+
+    return self
+
   def set_ignore_if_contains(self, words=None):
     """ignores the don't likes if the description contains
     one of the given words"""
@@ -221,7 +233,8 @@ class InstaPy:
         try:
           inappropriate, user_name = \
             check_link(self.browser, link, self.dont_like,
-                       self.ignore_if_contains, self.username)
+                       self.ignore_if_contains, self.ignore_users,
+                       self.username)
 
           if not inappropriate:
             liked = like_image(self.browser)
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -38,15 +38,9 @@ class InstaPy:
     self.logFile.write('Session started - %s\n' \
                        % (datetime.now().strftime('%Y-%m-%d %H:%M:%S')))
 
-    if username is None:
-      self.username = environ.get('INSTA_USER')
-    else:
-      self.username = username
+    self.username = username or environ.get('INSTA_USER')
+    self.password = password or environ.get('INSTA_PW')
 
-    if password is None:
-      self.password = environ.get('INSTA_PW')
-    else:
-      self.password = password
 
     self.do_comment = False
     self.comment_percentage = 0
@@ -100,10 +94,7 @@ class InstaPy:
     if self.aborting:
       return self
 
-    if comments is None:
-      comments = []
-    self.comments = comments
-
+    self.comments = comments or []
     return self
 
   def set_do_follow(self, enabled=False, percentage=0, times=1):
@@ -123,9 +114,7 @@ class InstaPy:
     if self.aborting:
       return self
 
-    if tags is None:
-      tags = []
-    self.dont_like = tags
+    self.dont_like = tags or []
 
     return self
 
@@ -147,9 +136,7 @@ class InstaPy:
     if self.aborting:
       return self
 
-    if words is None:
-      words = []
-    self.ignore_if_contains = words
+    self.ignore_if_contains = words or []
 
     return self
 
@@ -158,9 +145,7 @@ class InstaPy:
     if self.aborting:
       return self
 
-    if friends is None:
-      friends = []
-    self.dont_include = friends
+    self.dont_include = friends or []
 
     return self
 
@@ -174,7 +159,7 @@ class InstaPy:
 
     if secret is None and self.clarifai_secret is None:
       self.clarifai_secret = environ.get('CLARIFAI_SECRET')
-    elif secret is not None:
+    elif secret:
       self.clarifai_secret = secret
 
     if proj_id is None and self.clarifai_id is None:
@@ -207,8 +192,7 @@ class InstaPy:
     commented = 0
     followed = 0
 
-    if tags is None:
-      tags = []
+    tags = tags or []
 
     for index, tag in enumerate(tags):
       print('Tag [%d/%d]' % (index + 1, len(tags)))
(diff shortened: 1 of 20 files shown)


timgrossmann:
@pekapa Thank you for collaborating!
Really like the *include owner comments when checking for tags* part of this, also the ignore users is useful, but I don't really get the difference between this and the already implemented 
```python
set_dont_include([friend1, friend2 ...])
```

Please also add the "documentation" to the Readme and the examples/example.py, so people can see the feature

pekapa:
@timgrossmann Done.
The difference from `set_dont_include` is that it only avoids commenting and unfollowing while my version ignores it completely.
In my case, I'm always checking for a tag that I know a few people use it and I don't want to like the pictures form those.

eduardocalazansjr:
Just got a strange usage of ram/cpu and sometimes chrome just hangs/force crashes while fetching "media.comments.nodes" on photos with a lot of comments.

OSx 10.12, 28gb ram, latest Chrome Stable (also tried with Chromium)

pekapa:
@vlntdds I'll have a look on that,
My first try was to actually filter it via js before sending it to the browser but I kept getting an unexpected error.

Do you have any link with tons of comments? How many are we talking about? 1k, 100k?
I'm working on OSx 10.12 with 16gb of ram and didn't ran into that kind of problem.

EDIT:
From what I can tell, there should be no difference as "media.comments.nodes" always has length smaller than 36, which is also a problem when trying to grab the first comment of them all!

eduardocalazansjr:
Just tested on two droplets w/ 2gb (centos + ubuntu)
The same crash occurs, always if the logged account have `window._sharedData.qe.ebd.p.use_new_styles = true`. 

With accounts that don't have this attribute, the crash didn't happen.
Happened on different posts, with less comments (i.e: 2 or 50)

pekapa:
The comment filtering and mapping is running via js now reducing a lot the amount of data that needs to be sent back to python.
I was unable to reproduce so, if you could please check if it works now, I would appreciate.
If it's not working, could you show me the traceback?

eduardocalazansjr:
```
Uncaught TypeError: Cannot read property 'filter' of undefined
    at <anonymous>:2:8
```
The `owner_comments `value is always `"\nundefined"`

Only occurs in the account that have the `use_new_styles` flag

(No hangs for now, just the error)

pekapa:
New fix...
My account has the `use_new_styles` flag and I can't reproduce, must be something else.
Anyway, I treated the possibility of `comments.nodes` being `undefined`.
Let me know if the error persists (or if there's a new one)

eduardocalazansjr:
Always return an empty string now, doesn't filter/get the creator's comments/replies.

Edit: got the error again:

```
VM669:2 Uncaught TypeError: comments.filter is not a function
    at <anonymous>:2:3
```

pekapa:
Just introduced some console.logs to check
Could you please tell me if `latest_comments` is of type object in those checks?

I really can't reproduce this, is it related to the logged in user account or to the post?
If it's related to the post, can you send me the post link?
If it's related to the user, how can I create a user with / without that flag?
Do the problem persist on your OSx? Should I try to get a droplet to test it there as well?

timgrossmann:
@vlntdds Any updates on the errors?

eduardocalazansjr:
@timgrossmann will check asap.

@pekapa nice, will send you the output in a few hours!
Not post related, i think (sometimes it passes the posts that the have previously crashed)

It happens on two accounts that i use for testing (logged in), but not on the account that doesn't have the layout flag, as i said.

Also in OSx and on the droplets.

eduardocalazansjr:
With this PR i couldn't start InstaPy without the "ignore_users" setting, currently i can.

Script:
```
InstaPy(username='9', password='9')\
  .login()\
  .set_do_comment(True, percentage=10) \
  .set_comments(['Cool!', 'Awesome!', 'Nice!']) \
  .like_by_tags(['#programming', '#reactjs', '#webdevelopment', '#angularjs', '#nodejs', '#java', '#developer', '#coding', '#javascript', '#webdesign'], amount=100) \
  .end()
```

Error:
```
  File "/Users/eduardocjr/Desktop/devel/InstaPy/instapy/instapy.py", line 236, in like_by_tags
    self.ignore_if_contains, self.ignore_users,
AttributeError: InstaPy instance has no attribute 'ignore_users'
```

pekapa:
@vlntdds could you please check again?

instapy/instapy.py
```diff
@@ -129,6 +130,18 @@ def set_dont_like(self, tags=None):
 
     return self
 
+  def set_ignore_users(self, users=None):
+    """Changes the possible restriction to users, if user who postes
+    is one of this, the image won't be liked"""
+    if self.aborting:
+      return self
+
+    if users is None:
```
timgrossmann: Please use:

```python
self.ignore_users = users or []
```

here...

instapy/like_util.py
```diff
@@ -56,7 +58,29 @@ def check_link(browser, link, dont_like, ignore_if_contains, username):
   user_name = browser.execute_script("return window._sharedData.entry_data.PostPage[0].media.owner.username")
   image_text = browser.execute_script("return window._sharedData.entry_data.PostPage[0].media.caption")
 
-  """If the image has no description gets the first comment""" 
+  owner_comments = browser.execute_script('''
```
timgrossmann: @vlntdds This looks pretty error prune, does it work with different OS and devices?

instapy/like_util.py
```diff
@@ -56,7 +58,29 @@ def check_link(browser, link, dont_like, ignore_if_contains, username):
...
+      .reduce((item, total) => item + '\\n' + total, '');
+    return owner_comments;
+  '''.format(username))
+  if owner_comments == '':
+    owner_comments = None
+
+  """Append owner comments to description as it might contain further tags"""
+  if image_text is None:
+    image_text = owner_comments
+  elif owner_comments is not None:
```
timgrossmann: You could also do:

```python
elif owner_comments:
...
```

timgrossmann (changes requested):

pekapa:
I've updated this to include the changes and solve conflicts but it keep showing "Changes requested", I'm not sure what I should do to solve it.

timgrossmann (approved):

timgrossmann:
@pekapa Sorry, this was on my side.
After I request it, I have to approve it.

Did you test your code?
After the problem with the last merge I'd like to make sure you tested it locally before merging it.

Can't test on my own right now

pekapa:
Yes, I've tested.
All changes were made locally.

The other one was made directly on GitHub as it was supposed to be an "easy" one...

timgrossmann:
@pekapa  Well, lets see how it works in use then 😉

Issue #55, opened 2017-04-08
Filter by media type


tibor:
Since Instagram has both videos and photos now, it would be nice to filter for media types. Otherwise, comments like “Great photo!” might seem akward when posted on videos.

timgrossmann:
@tibor You're right, this could be a small check-feature... Like to only comment on photos

foname:
I'll work on this ^^

pekapa:
working on it; will make a PR soon.

eduardocalazansjr:
@tibor @timgrossmann  what do you guys think of two distincts comments arrays? one for photos and one for videos?

pekapa:
If we make two comments arrays we might break backward compatibility.
I believe there should be three arrays, a general one and two specific ones (photos e videos).

I'll wait for a definition to implement as a whole and only then I'll submit a PR.

eduardocalazansjr:
Why three?
Keep the original and add another for vids.
If the second is empty, use the first for all cases.

pekapa:
Suppose I have the following comments:

"Amazing" - for both classes
"Great video" - for videos
"Nice shot" - for photos

how would you arrange them in just two arrays? I can't see a way, unless I duplicate the "Amazing" into both arrays, or is there any other way?

eduardocalazansjr:
#56 solves

timgrossmann:
@vlntdds @pekapa I also think three would be better, since you don't want to have like 100's of comments duplicated at two different places

pekapa:
Just submitted PR #68 
Just plain comments, didn't get to work with Spintax for now

eduardocalazansjr:
It's better wait for the spintax implementation for this, isn't?

timgrossmann:
mostly solved by #68

PR #68, opened 2017-04-10
Implement #55

pekapa:
Also improved how we find links so we can get the proper amount of links for the appropriate media type.


timgrossmann:
@pekapa @vlntdds With such an amount of changes, this definitely needs some testing.
Did you move the chrome window out of focus while scrolling up and down to get all the links to the images?

Currently, if you move to chrome window out of focus, it won't get all the links since the automatic load more upon reaching the pages end won't work anymore. The while loop seems pretty dangerous combined with that problem...

Will it get the exact amount of posts?

pekapa:
@timgrossmann thanks for pointing that out. Added a method to bring it the focus and guarantee we can actually get the posts.
I've tested while out of focus and it works (can be a little annoying to have the window popping in your face while scrolling but I guess it's a necessary evil! Let me know if you have another solution.

Yes, it loads up to 11 more posts than needed (if the last load needs just one and it gets 12 of the right media type). And passes through the exact amount.

eduardocalazansjr:
Get the focus is not the recommended.
I think that it needs a in-depth test, will do ASAP.

eduardocalazansjr:
Sorry for the delay, @pekapa!
I will start testing this PR and #60 just now and will write some feedbacks and ideas for you, ok?

Thanks.

timgrossmann:
@vlntdds Any updates on this and #60 ? 😉 👍

timgrossmann:
@vlntdds Any updates?

eduardocalazansjr:
@timgrossmann as soon as i get home (today) i will test (i'm on a unplanned trip :( )

eduardocalazansjr:
Made some tests, seems to work fine with all of my accounts.
@pekapa fix the conflicts, thks!

pekapa:
Just fixed the conflicts.
Also fixed #118 with this and improved the `while` to avoid loading more than there's available.

It is working for me but might be worth to test again.

eduardocalazansjr:
Why not split the commits?
Too much changed, it's better test and review them separately

pekapa:
Because the point he raised in that issue is valid and affects the `while` loop directly. User could be really stuck in it so I was not comfortable letting this goes live without that fix.

If you want I can split them, but the logic to fix without the `while` is different than with the `while` so I think it would be just extra trouble.

If you need tags to help you text, I used `['#landscapeasdf', '#landascapesicily', '#landscape']`
The first has no image, the second, about 130 images and the latter, all the images you might need.

PR #74, opened 2017-04-11
Some new features and improvements (w/ bug fixes)

sagielevy:
Using single logging.logger instance to print and handle log file
instead of printing using ‘print’ and then logging to file via file
opening
Unfollowing starting from the bottom of the list rather than the top,
thus working in FIFO manner rather than LIFO. Otherwise you’re going to
get stuck with the same people InstaPy follows at the beginning.
Option to select if to unfollow from top or bottom of list (default is
top)
Loading enough users from following list prior to unfollowing session
Added scheduled job to run the task of the script to quickstart.py at a
predefined time of day so that the script can essentially run foreve


timgrossmann:
@sagielevy Thank you for the collaboration, we will check it out ASAP.
Once tested, I'll be merged.

eduardocalazansjr:
One question: why use 'logging.logger'?

I've take a look on the commits, maybe it's time to get rid of all classes element and find all by XPATH (most reliable).

sagielevy:
It's better to use logging.logger for a couple reasons:

1. Single handler of output both for printing to console and file
(duplicated code is never good practice)
2. It's thread safe, unlike normal file writing, which could be useful if
multi threading features would be introduced (e.g. unfollowing users while
running like by tag)

eduardocalazansjr:
I didn't get why the `sleep(600)` after the 10th unfollow.
I've cleaned one of my accounts unfollowing everyone (+/- 400 users) without sleep and didn't notice nothing wrong.

Also didn't get why `body_elem.send_keys(Keys.HOME)` is commented, it's very necessary depending of the cenario (slow hardware of the guest, slow internet)

sagielevy:
That is irrelevant for my PR, I didn't add that line

eduardocalazansjr:
Tested and it breaks the "infinite loading" on the photos feed, as i said.
And i think that the pause between unfollows is irrelevant (but just for unfollows)

sagielevy:
What do you mean by breaking the infinite loading?

I did receive an exception there in the unfollow_util and came up with a
solution. I can re-request a PR with my updated code if that's the issue

eduardocalazansjr:
The combination of END + HOME key up is necessary for the infinite scroll loading on selenium.
But it just applies on slow internet connections and slow machines, as i said.
If you JUST use the END keypress, while the instagram page still loading the next photos the next HOME keypress is ignored, for that the HOME + END function.

The logging util is ok, despite of that selenium and webdriver ARE NOT thread-safe and we can't do any multithreading task with that guys for some reasons (instagram bot detection, another chrome instance would eat a considerable amount of ram and a sort of issues).

And i think that the delay on unfollows is realy unnecessary.

@timgrossmann, can you review that?

timgrossmann:
@sagielevy @vlntdds 
The delay was added, because I noticed some weird behaviour if I unfollow more than 10 accounts after each other. 
The 11th and so on just stayed there and my unfollowing was blocked for ~10min.
That's what the delay was about.
But if you can't reproduce that behaviour, we could throw that out.

If totally fine with logging instead of printing and writing separately. This just had very low priority on my list, but since it's there now, why not...
The only thing I don't like about it is that it adds an extra parameter to a lot of functions...

@sagielevy It would really be way easier if you'd make like feature PR's.
One for the logger, one for the unfollowing and so on, because we can then simply review that one feature and merge the single PR's on their own. 
The way it is right now, we'd take everything or nothing and that's not too cool for as well as for us.

We'll definitely keep the HOME and END keystrokes, since this is needed to get all of the posts.

Also, get rid of the logs, we want to keep the repo's logs clean.

Especially the instapy.py is really hard to review, since there are so many changes...

Also, we need to keep Display, since it's needed for running the script on a server.
```bash
from pyvirtualdisplay import Display
```

<strike>And it looks like all the chrome_options are deleted...</strike>
Just saw the function for that 😉 

I know this might be quite some work, but would it be possible to split the feature implementations into separate smaller PR's? 👍 

Thank you so much for collaborating!

sagielevy:
I'll see what I can do about those multiple PR requests... This really does seem like a lot of work

I'll return the HOME and import and clear the logs.

timgrossmann:
@sagielevy Thank you so much!
Yes, see what you can do. That's all I'm asking for. 
Looking forward to your changes.

sugionodjojo:
hello,

  File "/InstaPy/instapy/instapy.py", line 34, in __init__
    self.browser = webdriver.Chrome('./assets/chromedriver', chrome_options=chrome_options)
  File "/usr/local/lib/python3.5/dist-packages/selenium/webdriver/chrome/webdriver.py", line 69, in __init__
    desired_capabilities=desired_capabilities)
  File "/usr/local/lib/python3.5/dist-packages/selenium/webdriver/remote/webdriver.py", line 90, in __init__
    self.start_session(desired_capabilities, browser_profile)
  File "/usr/local/lib/python3.5/dist-packages/selenium/webdriver/remote/webdriver.py", line 177, in start_session
    response = self.execute(Command.NEW_SESSION, capabilities)
  File "/usr/local/lib/python3.5/dist-packages/selenium/webdriver/remote/webdriver.py", line 236, in execute
  File "/usr/local/lib/python3.5/dist-packages/selenium/webdriver/remote/errorhandler.py", line 192, in check_response
selenium.common.exceptions.WebDriverException: Message: unknown error: Chrome failed to start: exited abnormally
  (Driver info: chromedriver=2.29.461571 (8a88bbe0775e2a23afda0ceaf2ef7ee74e822cc5),platform=Linux 4.4.0-66-generic x86_64)

is the script instapy.py line 34 i have to change or my droplet setting still not right?

AlainPilon:
This is turning into a giant monster of a PR. I suggest to split it into multiple PR, each one tackling a single issue.

timgrossmann:
@AlainPilon @sagielevy 
Yes, that's the only problem...
I really like a lot of the changes, but the size of the PR makes it untestable and really hard to read...

Let's see if there will be any updates

timgrossmann:
@AlainPilon @vlntdds Looks like we'll have to dissect the PR and take out the best of it.

I think there are some cool features in this PR, sad, that @sagielevy couldn't do it...

sagielevy:
I have to get back to my university tasks, I don't see when I'll have the time to split this into multiple PRs. You're still welcome to take any of the features included

AlainPilon:
kids are crazy sick since yesterday (noticed, I did not commit anything since! ) and with crunch week coming at work it will be challenging to do anything for the project. I will try to review everything and at least extract the easiest part.

PR #75, opened 2017-04-11
Description encoding problem solved.

hicay:


timgrossmann:
@hicay This won't fix the solve the problem in the terminal. If the terminal output is set to be ASCII even encoding it with utf-8 won't help.

Did it work for your machine?

timgrossmann:
Reverted, since it didn't work

hojabbr:
Just add these lines to `quickstart.py` and `instapy.py`:

```
# -*- coding: UTF-8 -*-
import sys
reload(sys)
sys.setdefaultencoding('utf-8')
```

timgrossmann:
@hojjabr Feel free to submit a PR with these changes 👍

Issue #18, opened 2017-04-04
Add docker


apertureless:
Thought about dockerizing it ? Would be way easier to deploy.

timgrossmann:
Yes, I just lately thought about that.

I don't have enough knowledge in this area by now but I'm open to PR's 😉 
I mean, writing a Dockerfile shouldn't be too hard and time consuming, but right now I sadly don't have the time to cover any of the open Issues...

stephane-archer:
Done ;) see pool request :D

timgrossmann:
@stephane-archer Awesome thank you!
Fixed by #32 

Edit: Still open till merge

timgrossmann:
#77 will resolve this issue...
Just the encoding isn't working... Does someone have an idea?
@apertureless @stephane-archer @jaycegregg0

timgrossmann:
Fixed by #77

apertureless:
@timgrossmann Maybe because of wrong locale?

timgrossmann:
@apertureless  Yes, but it can't set it. 
Or more like, it set's it but still throws the same error

PR #77, opened 2017-04-11
added Dockerfile

timgrossmann:
> Build yourself an Docker image which will pull in your current quickstart.py.

Not working by now -> Problem with encoding...
Does someone know an answer? 

> Edit: Fixed by removing the printing of the description


timgrossmann:
Fixes #18

PR #78, opened 2017-04-12
Update dockerfile

apertureless:
# Update Dockerfile

## Cleanup dockerfile

I removed the unused env variables for the locale and language and added env variables for the chrome and chrome driver urls. This way it's cleaner to change them later.

Also added a `apt-get clean` to cleanup the cache. And removed the lists.

## Env Variables
Added the username and password as env variables to the quickstart.py.
This way you can launch multiple docker containers for multiple instagram accounts.

## Update Readme

Added a docker section to the readme with instructions how to build the image and run the container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,6 +45,10 @@ RUN apt-get update \
     && pip3 install --upgrade pip \
     && apt-get -f install
 
+ENV LANG en_US.UTF-8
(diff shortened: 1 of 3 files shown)


apertureless:
Should be working now 💯

timgrossmann:
@apertureless Perfect, thank you!

PR #82, opened 2017-04-13
Added Mint Install Script

binhonglee:
`linux64.sh` is tested on a **Mint** and a **Raspbian** which both worked as intended but not sure about `linux32.sh` though I assumed they should be the same just aside from the 'chromedriver' file being different.


timgrossmann:
@binhonglee Nice!
Thank you

PR #88, opened 2017-04-14
Simplify boolean logic

AlainPilon:
Going over the code, I saw some needless IFs which could be simplified into one liners.
Also, I cleaned up the code to use `any` function with a generator. `any` is pretty cool, especially when used in conjunction with a generator.


timgrossmann:
@AlainPilon Awesome, thank you!

Issue #14, opened 2017-03-14
Add check for follower number


timgrossmann:
#### Would be nice if you could define an upper bound of followers an author of the posts you like should have. The more followers the account, the less likely it is that they will follow you back, so you can simply save that interaction

Could look somewhat like this:
```python
.set_upper_follower_count(1000) # default is INT Max
```

- [x] implement account check for each post
- [x] add check for more than given upper bound

7ruth:
This would be a great feature to have!

Do you have plans to make a feature for similar upper check but on likes? (for example on liking posts that have below a certain amount of likes...for the same reasons, its more impactful for people that don't have mega accounts). I am brand new to Python and just got to play around with your tool today, its really awesome!

timgrossmann:
@7ruth yes this feature actually is meant to exclude these accounts completely, no liking, no commenting, no following... Accounts with a lot of followers won't do any interaction with your account and that means it'll just cost Instagram interactions which will have no impact on your profile

littleswany:
@timgrossmann I will work on this issue

thecolorblue:
Would it be better to add in generic filtering for any attribute on an account or post?

timgrossmann:
@thecolorblue That's an interesting question. 
What besides the follower count and the following count could be interesting to check?

pekapa:
@timgrossmann I believe the number of likes and comments on the current picture should be used as a filter.

timgrossmann:
@pekapa Well if you'd do that, the filter would be really dependent on the timing... Even a profile with thousands of followers could have a post with only few likes and comments directly after they posted them.

littleswany:
I have submitted a pull request adding this

pekapa:
@timgrossmann that's very right indeed, but one can use that together with the follower count.

timgrossmann:
solved by #91

PR #91, opened 2017-04-14
Addition of issue #14 - Checking the follwer count

littleswany:
Adding functionality that checks the numbers of followers a user has
and compares this to a user set value and if this exceeds the value no
further interaction happens.

Command:
.set_upper_follower_count(limit = 250) \

Addition of this to documentation and readme also


instapy/instapy.py
```diff
@@ -184,6 +186,13 @@ def clarifai_check_img_for(self, tags=None, comment=False, comments=None):
 
     return self
 
+  def set_upper_follower_count(self, limit=100):
+    """Used to chose if a post is liked by the number of likes"""
+    if limit is None:
+      self.like_by_followers_limit = 10000000000
+    self.like_by_followers_limit = limit
+    return self
```
AlainPilon: 190 to 193 could be replaced with:
`self.like_by_followers_limit = limit or 1000000000`

pekapa:
If I might make a suggestion...

Instead of using a huge number (10000000000), use 0 and add and `if like_by_followers_limit`
Also, the way your are parsing the `.set_upper_follower_count` the user can only set it to `None` if he explicitly do so (change de default `limit` value to `None`). Finally, if it is `None`, the code does break as you set it to `None` and not the huge number (int > None always evaluate to True).

littleswany:
Thanks @pekapa.

@pekapa so are you suggesting to have the default value as 0 and then checking for a 0 value or setting the default to None?

instapy/instapy.py
littleswany: Thanks @AlainPilon will look into this

pekapa:
I'm suggesting to have the default as 0 and in than:

```python
if like_by_followers_limit and (num_followers > like_by_followers_limit):
    return True, user_name
```

no need to use `None`

littleswany:
Thanks, yeah Im with you, gimme a sec lemme just update and test

littleswany:
@pekapa Have updated

AlainPilon:
BTW a little bit off topic but, what was the thought process to have the functions return `True, user_name` when they should not be liked? I know the variable is `inappropriate` but shouldnt we reverse the boolean logic to make it more intuitive?

README.md
```diff
@@ -1,9 +1,9 @@
...
+> Feel free to contribute to the project in whatever way! <br /> If you're not familiar with python, you could build a github page for this project (Just head over to the issues, there might be a task for you). You're a marketer? Perfect! hit me with a message on contact.timgrossmann@gmail.com. <br/> If you don't have the time or skills to contribute, you can also support us through Patreon! <br /> **Become a part of InstaPy!**
 
 > **Disclaimer**: Please Note that this is a research project. I am by no means responsible for any usage of this tool. Use on your own behalf. I'm also not responsible if your accounts get banned due to extensive use of this tool.
 
-<img src="http://i.imgur.com/9ZjtveL.png" width="150" align="right">
+<img src="https://s3-eu-central-1.amazonaws.com/centaur-wp/designweek/prod/content/uploads/2016/05/11170038/Instagram_Logo-1002x1003.jpg" width="200" align="right">
```
timgrossmann, line 6: Please pull the master branch again and rebase on it...
We don't want to alter the readme more than your insertions 😉

instapy/like_util.py
```diff
@@ -56,15 +57,29 @@ def check_link(browser, link, dont_like, ignore_if_contains, username):
...
+  sleep(1)
+  num_followers = browser.execute_script("return window._sharedData.entry_data.ProfilePage[0].user.followed_by.count")
+  browser.get(link)
+  sleep(1)
+
+
   print('Image from: ' + user_name)
   print('Link: ' + link)
   print('Description: ' + image_text)
+  print "Number of Followers: ", num_followers
```
timgrossmann, line 78: We only want to print this if we have set the `like_by_followers` to true, don't we?

timgrossmann (changes requested):

timgrossmann:
@littleswany This is really nice! Thank you for the collaboration!
Could you please also add lower_follower_count?
I think that shouldn't be too much work now, since you already added the upper one 😉 

> Make sure to rebase :P

README.md
littleswany: Done, will update shortly.

instapy/like_util.py
littleswany: I have removed as it is not really necessary just had it in there for my testing.

littleswany:
@timgrossmann Glad you like it and your welcome, glad to be able to get involved!
Have committed the requested updates.

README.md
timgrossmann: ``


sorry, I probably communicated it wrong:
This is the new image link

Also, could you please add  your additions to the index.md in /docs/?
Would be awesome!

timgrossmann (approved):

littleswany:
@timgrossmann No worries, have updated!

timgrossmann:
solved #14

PR #94, opened 2017-04-15
add switch to check if all clarifai tags are present or only one

AlainPilon:


timgrossmann:
@AlainPilon could you please shortly describe, why this could be useful?
Did you ever check their site?
They return a huge list of possible tags for given image 😄 I'm not sure if you want to check for a full match.

But I'm open for discussion.

AlainPilon:
The full match would only check if MY tags are included in their (huge) list, not if they are identical lists. Basically, from the experiments I have done so far with IG search, it puts more importance on result freshness or popularity of the poster than accuracy. Ex: searching for `#zombie` returns a bunch of memes or tattoos, two things I am not interested in. So now I could refine the search for `zombie`, `group`, `red`, etc... And like pictures which are closer to what I am looking for, increasing my odds of a like back.

timgrossmann:
Well, that sounds really nice 👍 
Could you maybe add documentation to the readme as well to describe what this is useful for?

Thank you

AlainPilon:
@timgrossmann Done. I am planing to do a lot more changes to the clarifai module to make it more flexible (to fit my requirements) but it is going to take a few days.

timgrossmann:
@AlainPilon Sounds really nice!
Looking forward to that 😉

PR #100, opened 2017-04-15
Introduced random times to the script.

pekapa:
Times that were fixed before now falls in a range +/- 50% from the original value selected from a normal distribution.
e.g. `sleep(2)` can now be anywhere in-between `sleep(1)` and `sleep(3)`, normally distributed in this range, centered around `2`.

### DIFF
--- a/instapy/time_util.py
+++ b/instapy/time_util.py
@@ -1,0 +1,22 @@
+"""Helper module to handle time related stuff"""
+import random
+import time
+
+
+# Amount of variance to be introduced
+# i.e. random time will be in the range: TIME +/- STDEV %
+STDEV = 0.5
+
+def randomize_time(mean):
+  allowed_range = mean * STDEV
+  stdev = allowed_range / 3  # 99.73% chance to be in the allowed range
+
+  t = 0
+  while abs(mean - t) > allowed_range:
+    t = random.gauss(mean, stdev)
+
+  return t
+
+
+def sleep(t):
+  time.sleep(randomize_time(t))
--- a/instapy/time_util.py
+++ b/instapy/time_util.py
@@ -1,6 +1,6 @@
 """Helper module to handle time related stuff"""
-import random
-import time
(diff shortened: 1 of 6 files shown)


instapy/time_util.py
```diff
@@ -0,0 +1,22 @@
+"""Helper module to handle time related stuff"""
+import random
```
timgrossmann: Could you please do specific imports?

We don't need the whole random module here.

Thank you for the PR, these are some really nice changes 👍

timgrossmann (approved):

AlainPilon:
Shouldnt we also do the same thing for the numbers of likes, comments, etc.. ? Obviously, this would create some confusion is someone asks for 100 likes and does 127 but this would make it look more natural. 

Or I am going overboard with this?

instapy/time_util.py
pekapa: Done.
And thx!

timgrossmann:
@AlainPilon You can simply handle this in the quickstart.py

```python
tag1 = randint(200, 300)
tag2 = randint(50, 350 - tag1)
tag3 = 450 - tag1 - tag2

.like_by_tags(['vegan'], amount=tag1)\
.like_by_tags(sample(like_tag_list, 1), amount=tag2)\
.like_by_tags(sample(like_tag_list, 1), amount=tag3)\
```

timgrossmann:
And for starting on random time:

```python
if randint(0,10) < 9:
  try:
    sleep(randint(0, 60 * 10))
    startScript()
```

This way you get quite random behaviour if you're running it on a server started by a cron job

eduardocalazansjr:
Instagramweb seems to gather all the mouse/keyboard events via javascript.
I think that simulating the page movement is more efficient than random sleep times (also with the page loading times, it will be equivalent to a random sleep time)

timgrossmann:
I think we can improve this further later...

PR #109, opened 2017-04-15
Fix parentheses and quotes in like_util.py

maxinedotdev:
Syntax error: Missing parentheses in call to 'print' in like_util.py, line 80
Coherency: Change double quotes to single quotes in like_util.py, line 80


timgrossmann:
@hillockink Thank you

eduardocalazansjr:
The output for me is:
`('Number of Followers: ', 40490)`
I think that using string.format is better.

And it would be nice if the print was "--> Number of Followers: {x}", since it was below the comments, and sometimes the comment are just "big enough", hahah.

Edit: Or maybe just put this information bellow the "Image from: {x}" output and before the link, it would be nice too (and more clean, i guess)

PR #114, opened 2017-04-15
Only get number of followers if actually needs to.

pekapa:
Avoid going to the image poster profile page and back to the image if we are not filtering the likes by number of followers.

This is not about the logged in user (the one running the script). This is about the owner of the picture being liked.

### DIFF
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -65,24 +65,24 @@ def check_link(browser, link, dont_like, ignore_if_contains, username, like_by_f
   if image_text is None:
     image_text = "No description"
 
-  """Find the number of followes the user has"""
-  userlink = 'https://www.instagram.com/' + user_name
-  browser.get(userlink)
-  sleep(1)
-  num_followers = browser.execute_script("return window._sharedData.entry_data.ProfilePage[0].user.followed_by.count")
-  browser.get(link)
-  sleep(1)
-
-
   print('Image from: ' + user_name)
   print('Link: ' + link)
   print('Description: ' + image_text)
-  print('Number of Followers: ', num_followers)
 
-  if like_by_followers_upper_limit and num_followers > like_by_followers_upper_limit:
-    return True, user_name, 'Number of followers exceeds limit'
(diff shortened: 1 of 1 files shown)


eduardocalazansjr:
The profile page needs to be opened for two reasons:
 - For comparison and write on FolowerNum.txt (it's kinda of a log/performance metter)
 - For future uses of the _userData, that only resides on the user profile page.

pekapa:
@vlntdds, this is the poster user profile, from the image being liked.
Not sure if this was clear enough but this is not the logged user profile.

eduardocalazansjr:
Oh, sorry, my bad.
Could you please explain better on the PR?

Also, take a look on https://github.com/timgrossmann/InstaPy/pull/109#issuecomment-294316590

pekapa:
Yeah, I fixed that in this commit as well. Used string concat as the previous prints to keep it consistent.

timgrossmann:
@pekapa Thank you, I'll edit the position of the follower get just a tiny bit

instapy/like_util.py
```diff
@@ -65,24 +65,24 @@ def check_link(browser, link, dont_like, ignore_if_contains, username, like_by_f
...
-    return True, user_name, 'Number of followers does not reach limit'
+  """Find the number of followes the user has"""
+  if like_by_followers_upper_limit or like_by_followers_lower_limit:
+    userlink = 'https://www.instagram.com/' + user_name
+    browser.get(userlink)
+    sleep(1)
+    num_followers = browser.execute_script("return window._sharedData.entry_data.ProfilePage[0].user.followed_by.count")
+    browser.get(link)
+    sleep(1)
+    print('Number of Followers: ' + num_followers)
```
littleswany, line 80: @pekapa This causes an issue when printing the follower number:

`  File "./instapy/like_util.py", line 100, in check_link
    print('Number of Followers: ' + num_followers)
TypeError: cannot concatenate 'str' and 'int' objects`

I will submit a PR to undo this and fix the issue unless there is another reason?

timgrossmann: should be fixed by #122 @pekapa

PR #121, opened 2017-04-16
improved like_from_image and like_by_tags documentation

calvinclaus:
When I first used this I had to read the source to figure out some details of the main methods:

- does tag = hashtag or is hashtag just a special kind of tag?
- can I omit '#'?
- are `like_by_tags` and `like_from_image` completely separate or do they work together? Is `like_from_image` some kind of starting point the bot starts looking for the tags specified in `like_by_tags`? (This question came up as there was just one comment for both methods and 'from' implies a starting point. Maybe should be called `like_by_tags_of_image`).
- why is the example url `www.instagram.com/image` not actually pointing to an image but a user? should I be passing an image or a user?

I tried to answer these questions in the README with an improved explanation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,10 +70,12 @@ from instapy import InstaPy
 session = InstaPy(username='test', password='test')
 session.login()
 
-#likes 100 posts of dogs
+#likes specified amount of posts for each hashtag in the array (the '#' is optional)
+#in this case: 100 dog-posts and 100 cat-posts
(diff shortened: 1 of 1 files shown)


timgrossmann:
@calvinclaus Awesome, thank you this probably helps a lot of people!
Would be awesome if you could go through the rest of the readme and clear up anything that is not clear to you!

Thank you!

PR #122, opened 2017-04-16
Changed print style to newer `.format` one.

pekapa:
Also, fixed issue where parenthesis were being printed together with the text.


AlainPilon:
Can't test right now but looks good to me.

PR #145, opened 2017-04-18
Fixed login_util

eduardocalazansjr:
Changed the elements and make it work on every login scenario.

Also added the support to ActionChains to avoid Instagram anti-bot
protection and simulate mouse movements


pekapa:
working fine for me

About the anti-bot, isn't it moving way to fast? Is there a way to slow it down?

timgrossmann:
@vlntdds Really nice and clean. Love it!

eduardocalazansjr:
@pekapa 
No, for the js gather it doesn't count the speed, but the movements on the page.
You will doesn't see the movement, of course, but it will track as an valid x/y movement through the elements.

For the submit time between pages i will keep the sleep after the click.

timgrossmann:
@vlntdds So... do you think xpath with text comparing is the least error prune solution?

eduardocalazansjr:
@timgrossmann i'm testing it in a couple of scenarios and the only thing that remains the same are the text on the elements.

Since they're using react to build the elements/css with random classes and the JS is kind of dubious, i think that text comparing is the best way out (at least for now)

Issue #142, opened 2017-04-18
Ignore accounts I have followed myself


raffg:
Ignore all accounts I followed manually. Only automatically unfollow accounts which were followed by the bot.

AlainPilon:
So obvious I dont understand why we did not think about it earlier!

timgrossmann:
I think it would be also really nice if you could unfollow LIFO instead of only FIFO... like #74 already suggested.
That way we would not instantly unfollow the just followed accounts

AlainPilon:
@timgrossmann I was planing to extract that feature tonight, if time allows.

timgrossmann:
@vlntdds Would be awesome!

timgrossmann:
#349

ghost:
How do I activate this feature?

PR #151, opened 2017-04-19
unfollow via FIFO (extract from #74)

AlainPilon:
This branch is a complete Cut&Paste from feature #74 from @sagielevy to only extract the FIFO unfollowing feature. 

I did not port the option to pick a FIFO/LIFO strategy since now that we have FIFO there is no logical reason why someone would want to use LIFO. 

I dont have enough followers to test it, so could someone check it out and confirm I did not miss anything from the original code.

### DIFF
--- a/examples/example.py
+++ b/examples/example.py
@@ -67,7 +67,7 @@ session.like_from_image(url='www.instagram.com/image', amount=100)
 # media filtering works here as well
 session.like_by_tags(['#test'], amount=10, media='Video')
 
-session.unfollow_users(amount=10) #unfollows 10 of the accounts your following -> instagram will only unfollow 10 before you'll be 'blocked
+session.unfollow_users(amount=10) #unfollows 10 of the oldest accounts your are following -> instagram will only unfollow 10 before you'll be 'blocked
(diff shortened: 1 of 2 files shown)


eduardocalazansjr:
I will make an PR to fix all the XPATH/JS refs and test this ASAP.

AlainPilon:
Now that I think about it, one of the side effect of using the FIFO may be that we would unfollow users who have been added by the actual account owner. This wouldnt be an issue once #142 is done. Until then, it is caveat emptor!

timgrossmann:
@AlainPilon could you add a parameter for the user to decide whether he wants to use FIFO or LIFO unfollowing? like #74 did?
Would be great!

littleswany:
@timgrossmann @AlainPilon Do we want this for the master branch or for the develop branch as at some point master will be replaced?

timgrossmann:
@littleswany Master, the one for V1 should have more feature... E.g. only unfollowing people the script followed, smart unfollowing after special amount of time...

RealMakAttak:
I would love this feature

PR #161, opened 2017-04-19
Fixed comment_util emoji output

eduardocalazansjr:
Added unicode flag to string.format

### DIFF
--- a/instapy/comment_util.py
+++ b/instapy/comment_util.py
@@ -14,6 +14,6 @@ def comment_image(browser, comments):
   comment_input.send_keys(rand_comment)
   comment_input.submit()
 
-  print('--> Commented: {}'.format(rand_comment))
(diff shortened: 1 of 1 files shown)


instapy/comment_util.py
```diff
@@ -14,6 +14,6 @@ def comment_image(browser, comments):
   comment_input.send_keys(rand_comment)
   comment_input.submit()
 
-  print('--> Commented: {}'.format(rand_comment))
+  print(u'--> Commented: {}'.format(rand_comment))
```
pekapa, line 17: You probably need a `.encode('utf-8')` added to the `rand_comment`

eduardocalazansjr: https://docs.python.org/2.5/ref/strings.html
.

Issue #97, opened 2017-04-15
Switch from Clarifai API v1 to v2


timgrossmann:
#### It would be nice, if we could switch our clarifai API from version 1 to the newer version 2 to make sure the script is supported longer

saphuttu:
When trying to use Clarifai I'm constantly getting
> Image check error: '{"detail": "Request was throttled."}'  Wait for 10 seconds before retrying.

No new users to v1 if I'm getting it right.
https://community.clarifai.com/t/request-was-throttled/402
https://community.clarifai.com/t/accessing-v1-video-analysis-throttled/464

timgrossmann:
@saphuttu Just looks like you already used your 5k free requests per Month...

But thank you for the links, this is new to me...
Good thing, V1 will have clarifai APIv2

saphuttu:
@timgrossmann No, i didn't get any requests because of APIv1. Clarify is throttling all new v1 users. I created my account few days ago. My Clarify account says in billing
> V1 API
> You are currently on the Unverified plan.

and in usage
> There is currently no activity for this account

CodeMaster1:
What is the solution to this issue? I keep getting the request throttled error for image checking. My set up includes this package for clarifai. "clarifai-2.0.20-py3.6.egg". Any helpful pointers here?

timgrossmann:
Oh ok, didn't know this is only for new users...

Did you select v1 in the usage tab? @saphuttu 
For v2 it says 
> There is currently no activity for this account

for me as well

saphuttu:
> V2 API
> You are currently on the Free Plan.

In the usage tab I can't do anything.

CodeMaster1:
@timgrossmann cool product. Let me know if rectify this issue. 

Thanks for quick reply

timgrossmann:
@saphuttu You should be able to switch the api version with a slider...

saphuttu:
@timgrossmann can you take a screenshot. I'm seeing no slider here. :D

CodeMaster1:
This is screenshot from my clarifai account. If someone can suggest what changes I need to do to get this API used for image identification, please let me know.

![image](https://cloud.githubusercontent.com/assets/7813918/25331179/adf701fc-28ff-11e7-9c46-b17216e0a41b.png)

saphuttu:
Where is the slider?

timgrossmann:
@saphuttu

Just click on usage, you can switch between v1 and v2 there.
[...]

PR #171, opened 2017-04-20
Refactor preview.

pekapa:
Include breaking changes to make it more logical and reduce confusion

Working:
* Basic like by tags
* ignore_users
* follower_limit
* dont_like
* like_anyway
* like_tags
* like_media

Need testing:
* Follow utilities
* Comment utilities

Not working:
* like_from_image
* follow_by_list
* follow_restrictions (only follow an specific amount of times)
* log_follower_num
* unfollow (make it so to only unfollow those followed by the script?)
* clarifai (make it v2?)

Need to do:
* Documentation and examples


pekapa:
@timgrossmann @AlainPilon @vlntdds 

I started the work on the refactoring, please have a look and let me know what you guys think.
Kept some of the old files for reference only, while it's not done.

AlainPilon:
Wow, you coded like a beast! Lots of code to review. 

I think ClarifAI has to be part of v1, but I will try to generalize its integration as an extension. Still sick but I am planning to work on this over the weekend.

pekapa:
> I think ClarifAI has to be part of v1

I mean, use ClarifAI v2, just like the issue #97

timgrossmann:
@pekapa Wow, you really did code like a beast 😄 
Really awesome! Even though this will take a while to review

timgrossmann:
@pekapa sorry for the close 😉 
Could you change the branch to `develop_v1`, please?

timgrossmann:
@pekapa We should either use `print` and `file.write` **OR** `logger.log`... Please don't mix the two things.

I'll look into the Page Object Pattern today so I'm ready to start developing Version1... @vlntdds @AlainPilon @pekapa How does your days look like? Do you have time to contribute?

@pekapa Please make smaller PR's so it's easier to review.

pekapa:
Oops, missed one `print` there. I'll fix it tonight.

My days are quite rushed until Wednesday at least.
After that I'll be able to help a little bit more. Let me know what you need.

About the big PR, the first one needed to be big as it involver changes to the very core. The next ones can be smaller ones as they are features and improvements to existing ones.

timgrossmann:
@pekapa Yes, once you fixed the print we'll merge that into develop.
From then on well go on and fix stuff and implement the rest of the features/fixes 😉

AlainPilon:
I am afraid I wont have much time to do anything, between work and 4 out of 5 family members (including me! :-( ) sick, there is not much free time. But I try to keep up to date on the dev branch and be able to hit the ground running once things clear up.

eduardocalazansjr:
In two day i will check It.
But for what i see, i think It will turn into a mess soon, why not use MVC?

AlainPilon:
@vlntdds Actually, it should make things cleaner because whenever you code an interaction with a page, you just have to call the method associated with that given page. 

It totally separate the business logic of the system (logic of liking/following) from the way we get the information (the browser). It also means that if you want to make an app to scrap images of IG, you can use our code, totally reusability. 

In the end, we should have one class per IG page inside a module "browser", a script to coordinate all the browsing and business logic, and another module with the various files used for the business logic (ex: ClarifAI).

I really home to have some time tonight to review the code.

timgrossmann:
I also think, that the, recommended (even from Selenium itself), PageObject Design Pattern probably will be a good fit.

timgrossmann:
@vlntdds Why do you think this will turn into a big mess?
Could you explain that a bit, please?

timgrossmann:
Well, let's just start with this as a first codebase...
Thank you @pekapa for your commitment.

Please resolve the conflicts and we'll merge this to get started.
Also check the issues in the project tab to move the ones you already worked on

eduardocalazansjr:
@timgrossmann maybe if there's an user w/ prefs elem and share it thru the classes will be more efficient and easy to maintain in a large timespan of this tool, since we don't know if we will use XPATH forever.

And i saw some strange boilerplates, why there's an function to return the length of an xpath if there's already an selenium and a native method for that? Same i see for filling forms.

timgrossmann:
@vlntdds These are some good points...
Maybe we really have to consider changing the design pattern to something other.

pekapa:
@vlntdds for the functions you mentioned, no special reason.
Please, feel free to improve to code wherever you believe it can be improved! I'm definitely not an expert on Selenium (in fact, I've never really used it before). I just wrapped the few ones that I knew to make it easier to maintain.

The whole idea centres on what we've discussed before, making classes for pages.
I added one extra class to hold the main browser actions and the page classes should be able to either get some state (e.g. username) or perform some action (e.g. follow a user).

pekapa:
Shall we merge this and actually start working on v1?
I think those new features that are starting to pop should be done directly into v1 and not in the current version to be re-written again after.

littleswany:
@pekapa I like that idea, I am waiting for that to start working on some of the features. But there are conflicts that need resolving first.

pekapa:
Fixed the conflicts.

Issue #112, opened 2017-04-15
Fix OSx and Linux scripts


eduardocalazansjr:
I've tried to use the automate installation scripts and the OSx one didn't work OOB.
And i think that the Linux scripts could be merged in only one, since there's how to distinguish if is x86 or x86_64.

- [x] Fix the OSx script and make sure that it works with OOB OSx installations
- [x] Merge the two linux scripts into just one
- [x] Check if is possible to move them to another folder (just to cleanup)

binhonglee:
> Check if is possible to move them to another folder (just to cleanup)

Pretty sure it is. Just the pathing needs to be updated.

timgrossmann:
@binhonglee Could you submit a PR for the cleanup and update the pathing?

PR #172, opened 2017-04-21
Merged both linux scripts to one

binhonglee:
Addressing #112 . If this script works, the other two can be removed.

@vlntdds


timgrossmann:
@vlntdds Could you do some quick tests? Please

pekapa:
Maybe we can merge OSx script as well.
I think `uname` and `uname -m` should give us all the information we need, ain't that right?

nijaru:
Yeah uname output gives you specific kernel (linux or darwin) plus architecture

timgrossmann:
@binhonglee Could you add the case for OSX as well, please?

binhonglee:
`linux.sh` should work for Mac now.

timgrossmann:
Awesome! Thank you @binhonglee

PR #175, opened 2017-04-21
Add nogui instapy constructor parameter

lanquemar:
I add an optional nogui parameter to the instapy class in order to use the project on a server without modifying the code.


pekapa:
Could you please change de documentation to include information about this new parameter?

PR #208, opened 2017-04-27
Added crontab section

tibor:


eduardocalazansjr:
But crontab isn't related to instapy.
In the same way that Windows scheduled tasks isn't.

And also navigating to the path works, but isn't recommended (it's a part of the chromedriver docs)

timgrossmann:
@tibor @vlntdds Yes, that's not part of InstaPy, but since it's just in the docs and not in the library, I think this will help a lot of people.

Thank you @tibor

Issue #186, opened 2017-04-23
Exact match for negative Keywords


tibor:
It seems as if the function ```set_dont_like``` not only skips the selected hashtags but also skips all hashtags that starts with the selected hashtags. While this might be useful, too, I guess we should implement an exact match feature as well.

I stumbled upon this issue when I set "#me" as a negative keyword and images with hashtags like "#meindeutschland" wouldn’t be liked anymore.

pekapa:
I had that with the word "hot", it would skip every post tagged with "photo" or something related. I changed to skip " hot" (with a space before).
I think the solution would be to do a split on spaces to find words and on hashes to find hashtags

tibor:
Yeah, but I think it would be nice to have both options (broad match and exact match).

PR #211, opened 2017-04-29
Resolve exact match / starts with for negative keywords

tibor:


tibor:
Resolves the problem mentioned in #186 like suggested [here](http://stackoverflow.com/posts/3271667/revisions)

Tested with `#me` in `#selfie`
Tested with `#me` in `#meinvorarlberg`

To do: Add option for `starts with`

eduardocalazansjr (commented):
Why use dictionary in that case?
Isn't more pratical /optimal to look for the string directly, with an START/END space without trim?

tibor:
What if the string would start with a negative keyword (no space before) or would end with a negative keyword (no space beyond)?

eduardocalazansjr:
regex solves this in a clean and better way, with just one interaction.

with this commit you do a total of four interactions on the image_text (lowercase, split, add to dictionary and scan the dictionary for the occurrence)

pekapa:
some people use all the hastags together, you might need to split by "#" as well, to cover those cases

e.g. #using#hastags#like#this

timgrossmann:
@tibor Please revisit and refactor for regex like @vlntdds and @pekapa mentioned

tibor:
@timgrossmann changed it to regex and tested it again. Works with hashtags with and without spaces between.

However, I’d like to generate the regex expressions just once per session instead of each loop / image. Where would I have to place it then?

tibor:
Not tested yet, but I think we could extend this to broad match, exact match, starts with and ends with functionality like this:

```
dont_like_regex = []

for dont_likes in dont_like:
  if dont_likes.startswith("#"):
    dont_like_regex.append(dont_likes + "([^\d\w]|$)")
  elif dont_likes.startswith("["):
    dont_like_regex.append("#" + dont_likes[1:] + "[\d\w]+([^\d\w]|$)")
  elif dont_likes.startswith("]"):
    dont_like_regex.append("#[\d\w]+" + dont_likes[1:] + "([^\d\w]|$)")
  else:
    dont_like_regex.append("#[\d\w]+" + dont_likes + "[\d\w]+([^\d\w]|$)")

# loop over dont_like_regex
```

You could then use them like this:

`.set_dont_like('#exactmatch', '[startswith', ']endswith', 'broadmatch')`

What do you think?

timgrossmann:
@tibor Yes, I like the idea of being able to decide... However, I don't think we think that many types of differentiation.

I'd rather just do a "exact match" if I type "food" I only want to match "food", not "asianfood"   
and the "broad match" you mentioned where if I type "food" I want "food" but also  "asianfood" and, questionably, also "notfood"

tibor:
Yeah, it would be more interesting for matching Insta-prefixes and suffixes.

E. g.`[ig` would match `#ig_houses`, `#ig_history` and so on but not `#biggestfan`

timgrossmann:
@tibor Ahh now I get what you would like to do with it!
Is it something that is used? Never saw that, but I'm not really looking for other tags than mine 😅

tibor:
Yeah, I guess that are some common prefixes. Also, in germany “mein” is used in combination with regions a lot (meinbawue, meinberlin, meindeutschland).

As a suffix, otd would be an often used example (potd, ootd)

However, I’ve adjusted the regex matches now so that starts_with and ends_with won’t match exact_match. My tests haven’t shown any errors.

Nevertheless, I guess it’d be useful to move the regex part outside the loop. Unfortunately, this exceeds my knowledge of python at the moment.

timgrossmann:
@tibor Could you please also add documentation?

tibor:
@timgrossmann sure. Do you mean as comments directly in code or do we have another documentation?

timgrossmann:
Well, I mean the Readme for Github, as well as the index.md in docs.
You can add it to the wiki page as well. 

Just show off, how your new feature can be used like you did above a `.set_dont_like('#exactmatch', '[startswith', ']endswith', 'broadmatch')`

timgrossmann:
@tibor well, let's see how this works in use 😉

PR #233, opened 2017-05-06
Fix issues: Return correct following user list and checking if User has comments disabled before trying to comment.

wordythebyrd:


tibor:
I guess, the xpath `//a[contains(concat(' ', normalize-space(@class), ' '), ' _4zhc5 ')]` could be simplified to `//a[@class="_4zhc5"]`

What @timgrossmann meant with using xpath, was more something like this: `//div/div/div/div/a`
The idea is that instagram will more likely change the class names rather than the structure of the web interface. So when we target the structure, our code will be more error prone.

tibor:
Tested the unfollow feature and it worked for me. Thanks!

timgrossmann:
@wordythebyrd Thank you!

PR #266, opened 2017-05-20
Update login_util.py to avoid errors

TGlide:
When the instagram website is in, for example, Brazillian Portuguese, there are errors because of the change of the name of certain html elements. So, this simple change fixes it.

### DIFF
--- a/instapy/login_util.py
+++ b/instapy/login_util.py
@@ -6,6 +6,11 @@ def login_user(browser, username, password):
   """Logins the user with the given username and password"""
   browser.get('https://www.instagram.com')
 
+  # Changes instagram language to english, to ensure no errors ensue from having the site on a different language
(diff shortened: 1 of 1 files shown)


timgrossmann:
@TGlide Thank you for your contribution!

Is this actually a problem?
InstaPy sets the chrome settings to all english when starting up...

TGlide:
@timgrossmann  Thank you for your reply!

I did not know that Instapy did that, that's interesting. 
While on my main computer there were no problems, when I used it on another computer, Instagram displayed itself in brazillian portuguese (I live in Brazil), even after I tried changing the chrome settings manually. It could be a problem on my end though. 

If you wish to implement this little change anyways, it can be a quick fix if anyone faces some issue of this kind.

timgrossmann:
@TGlide Just wanted to clarify on the problem 😉

PR #268, opened 2017-05-21
Added a section describing that order is important 

sunnypatel165:


timgrossmann:
@sunnypatel165 Thank you!

Issue #143, opened 2017-04-18
Emojis in comments


raffg:
Better handling of emojis in comments. I have added emojis in utf-8, but it would be much easier to import emoji and type :thumbs_up_sign:

timgrossmann:
@AlainPilon @vlntdds @pekapa This would really be a nice to have... Especially since some of the issues were due to people using the actual emojis instead of the unicodes.
As long as the comments are part of the clients setup we should make sure to ease the creation up...

What do you guys think?

pekapa:
I like the idea. I was having a look at the `emoji` module and it's just a big dictionary of emojis.
Unfortunately, I could only use selenium with 1 byte emojis (i.e. 167 out of 2388 available in that module)


(just for reference #2 )

shawoozy:
Use geckodriver (firefox) to use all emojis. Chrome supports just 4 character emojis

littleswany:
Looks like there is something here, using JS to set the value of the comment filed opposed to using sendKeys. Maybe this could work?

[https://groups.google.com/forum/#!topic/selenium-users/SYr0Xlzds30](https://groups.google.com/forum/#!topic/selenium-users/SYr0Xlzds30 )

timgrossmann:
@littleswany This sounds really interesting! Looks like it is worth giving a try.
I experimented with the "fill with JS" idea a bit when i started to add the commenting feature, but couldn't get it working how i thought that might work.

timgrossmann:
#281

PR #281, opened 2017-05-26
Emoji Code Names

atomhax:
PR addressing #143 raised by @raffg

This will let users type emojis in a text style such as --> :thumbs_up:


timgrossmann:
@atomhax Soooo awesome! Thank you

PR #285, opened 2017-05-29
likes photos using a number of tags from total tags

converge:
commit bedfb2c was deleted

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ session.login()
 session.like_by_tags(['#dog', 'cat'], amount=100)
 
 #likes 50 photos using 2 tags from total tags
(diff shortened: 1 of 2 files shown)


converge:
don't know why it's conflicting

timgrossmann:
@converge Thank you for your contribution!

Even though this is a nice Idea, I don't think this should be part of the core functionality.

If we add this, we should also add using a "window" to give a list of tags and then always take the next 2 and so on...

Since this feature can simply be added by using:

```python
tagList = [..., ..., ..., ...]

.like_by_tags(tags=sample(tagList, 2), amount=50)
```

This gives the user somewhat more control.
Also, `sort` is probably not the right word for this feature.

Thank you very much!
Keep contributing!

PR #295, opened 2017-06-03
Prevent crash when 0 people following, raise warning instead and brea…

MoritzGruber:
Prevent Crash when 0 people following, raise RuntimeWarning instead and break unfollow loop.


timgrossmann:
@MoritzGruber Thank you!

PR #328, opened 2017-06-19
Like from last post

jungfish:
Hi this is the new feature like from last post where I made an issue about. Hope you enjoy it !


timgrossmann:
@jungfish Thank you, looks nice!
Let's see this working 😉

Issue #338, opened 2017-06-24
selenium.common.exceptions.WebDriverException: Message: chrome not reachable


FBosler:
New bug, stopped my script twice now. I am running it on an ubuntu server. Script crashed after roughly 400 (for the first time) and 450 (for the second time) likes. Please find below the traceback. Cheers
Fabian

```
Traceback (most recent call last):
  File "LikeSCript.py", line 16, in <module>
    .like_by_tags([<My Tags>], amount=300, media='Photo')\
  File "/root/InstaPy/instapy/instapy.py", line 386, in like_by_tags
    liked = like_image(self.browser)
  File "/root/InstaPy/instapy/like_util.py", line 279, in like_image
    browser.execute_script("document.getElementsByClassName('" + like_elem[0].get_attribute("class") + "')[0].click()")
  File "/usr/local/lib/python3.5/dist-packages/selenium/webdriver/remote/webelement.py", line 111, in get_attribute
    resp = self._execute(Command.GET_ELEMENT_ATTRIBUTE, {'name': name})
  File "/usr/local/lib/python3.5/dist-packages/selenium/webdriver/remote/webelement.py", line 461, in _execute
    return self._parent.execute(command, params)
  File "/usr/local/lib/python3.5/dist-packages/selenium/webdriver/remote/webdriver.py", line 236, in execute
    self.error_handler.check_response(response)
  File "/usr/local/lib/python3.5/dist-packages/selenium/webdriver/remote/errorhandler.py", line 192, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.WebDriverException: Message: chrome not reachable
  (Session info: chrome=59.0.3071.109)
  (Driver info: chromedriver=2.29.461571 (8a88bbe0775e2a23afda0ceaf2ef7ee74e822cc5),platform=Linux 4.4.0-81-generic x86_64)
```

timgrossmann:
@FBosler Could you give some stats about how and where you're running it?

FBosler:
Sure. It just crashed for the third time in a row, this time at the first tag.

```
excludePpl = ['list of my friends']

InstaPy(username='myName', password='myPass!', nogui=True)\
  .login()\
  .set_dont_include(excludePpl)\
  .set_upper_follower_count(limit = 500)\
  .set_do_follow(enabled=True, percentage=10, times=1)\
  .set_do_comment(True, percentage=15)\
[...]

PR #378, opened 2017-07-08
Added to fix issue #338

Skullmaker90:
chromedriver binary path wasn't set the options, causing selenium to fail to find it.


timgrossmann:
@Skullmaker90 Perfect, thank you!

Issue #398, opened 2017-07-12
AttributeError: 'InstaPy' object has no attribute 'aborting'


Cas7or:
Hi Guys,

After reinstalling windows10 and reinstalling instapy I get the following output:

C:\Users\usrname\Documents\GitHub\InstaPy>quickstart.py
Traceback (most recent call last):
  File "C:\Users\usrname\Documents\GitHub\InstaPy\quickstart.py", line 3, in <module>
    session = InstaPy(username='usrname', password='pass')
  File "C:\Users\usrname\Documents\GitHub\InstaPy\instapy\instapy.py", line 40, in __init__
    self.set_selenium_local_session()
  File "C:\Users\usrname\Documents\GitHub\InstaPy\instapy\instapy.py", line 79, in set_selenium_local_session
    if self.aborting:
AttributeError: 'InstaPy' object has no attribute 'aborting'

Can you please help because I've been trying for 20h now and no luck.
Thank you in advance!

botpadawan:
I also got the same error updating the latest code

onyxim:
I've created [the pull request](https://github.com/timgrossmann/InstaPy/pull/399) to fix it.

[Pull the update from my master branch fork](https://github.com/onyxim/InstaPy)  if you want to fix it right now.

SameerKhan:
I get the same error after a pull


python quickstart.py
Traceback (most recent call last):
  File "quickstart.py", line 14, in <module>
    InstaPy() \
  File "/Users/sameer/OuzelSystems/Experiments/InstaPy/instapy/instapy.py", line 40, in __init__
    self.set_selenium_local_session()
  File "/Users/sameer/OuzelSystems/Experiments/InstaPy/instapy/instapy.py", line 79, in set_selenium_local_session
    if self.aborting:
AttributeError: InstaPy instance has no attribute 'aborting'

timgrossmann:
#400

PR #400, opened 2017-07-12
Fixed self.aborting issue. #398

onyxim:
Fix issues with self.aborting and small style fix.

[#398](https://github.com/timgrossmann/InstaPy/issues/398)

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -36,8 +36,6 @@ class InstaPy:
             self.display.start()
 
         self.browser = None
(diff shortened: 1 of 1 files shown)


onyxim:
It'll be good practice to use unit tests in future to avoid this kind of bugs:-) I think I'll make some soon.

timgrossmann:
@onyxim Thank you!
Yes, the test would really be necessary!

Would be awesome if you could contribute a few.

PR #457, opened 2017-08-02
Firefox as an option and more "user-configurable" parameters 

BernardoGO:
**Firefox as an option for the user**

> Most of the operating systems distributions already include Firefox and geckodriver. Chrome and ChromeDriver might be an issue on some ARM based systems.
> 
> The idea here isn't to replace Chrome as default browser. Instead, my idea was to provide an option for the user while keeping Chrome as default since it indeed runs smoother.
>
> Following the Chrome setting, Firefox is also capable of disabling the image load to use less bandwidth.

**A way to disable language switch**

> The language switch function can switch to a random language on some systems that already defaults to English.
>
> A simple way to disable this function without having to dig into the code can help less experienced users

**The implicitly_wait always set to 25 might be too much for some users**

> Following the idea of the _language switch option_, this is more of a quality of life parameter. Sometimes it is useful to change the implicitly_wait value.

**Fixed a bug on login_user function where both username and password were typed in the password textbox**

> The _ActionChains_ for both username and password was creating a situation where both was being typed in the password textbox on slower or under high load systems. Sometimes only the first letter of the username was correctly written in the correct box.
>
> Solved with a sleep between username and password send function.

The added parameters are optional and defaults to the original options which guarantees compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,6 +281,26 @@ crontab -e
 45 */4 * * * cd /home/user/InstaPy && /usr/bin/python ./quickstart.py
 ```
 
+## Switching to Firefox
+
+Chrome and ChromeDriver might be an issue on some ARM based systems. Alternatively, InstaPy provides support for Firefox as well. Most of the operating systems distributions already include Firefox and geckodriver.
+
+### Example with Firefox browser
+
+```python
+from instapy import InstaPy
+
+InstaPy(username='test', password='test', use_firefox=True, page_delay=25)\
+    .set_switch_language(False)\
+    .login()\
+    .set_do_comment(True, percentage=10) \
+    .set_comments(['Cool!', 'Awesome!', 'Nice!']) \
+    .set_dont_include(['friend1', 'friend2', 'friend3']) \
+    .set_dont_like(['food', 'girl', 'hot']) \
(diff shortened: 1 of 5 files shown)


timgrossmann:
@BernardoGO Awesome!
Could you please add documentation for this as well?

BernardoGO:
@timgrossmann Sure, I did a few changes in How_to_Raspberry.md since the modifications in the instapy.py code are no longer necessary. An example was added and the Readme.md was changed.

BernardoGO:
**Updated for the new Instagram login page**
> Instagram changed the position of the divs in the login page. The commit d7af29d fixes this issue.
> Sleep time during login increased from 2 to 5 to compensate the implicitly_wait.

GregoryHupkins (approved):
Fixes the issue IndexError: list index out of range

RobertVallew:
Replacing instapy/login_util.py by this one fixes the login issue.

RobertVallew (approved):

Naramsim:
Can we get this merged?

jpesce (approved):

mihikadutta:
@timgrossmann We need to merge this change into the master. If not the whole feature then atleast the fix for the DOM updates from the instagram.

Issue #456, opened 2017-08-02
Everything was working fine until...


n0sw34r:
Script was working fine (Ubuntu 14.04) until I saw that it stoped working and when I try to run it I got this...
Anybody help please?
![image](https://user-images.githubusercontent.com/30392165/28857181-b8f1fc74-7747-11e7-96ca-c932442713bf.png)

lgr:
I've observed thae same issue, working on a fix now

JMuffin:
There is a simple solution that I tried:
In instapy/unfollow_util.py in line 52 change dialog = ... to
dialog = browser.find__element_by_xpath("//div[contains(@Class, '_4gt3b')]")

It should work ;)

lgr:
The CSS class names seem to have hashed/random names, so IMHO it's better to refer to the DOM elements in a relative manner - as it was done so far.

JMuffin:
If instagram is changing so fast... there should be a better way to find this dialog box :/
I'm using XPath Helper for Chrome and it give me that path or
/html/body/div[4]/div/div[2]/div/div/div/div - as You done it.

lgr:
Yeah... another option is to search for the label "Following" in the dialog itself, however I guess not everyone has an English version of the UI

JMuffin:
I've tried it, but with no luck as the box is under div with label :/

timgrossmann:
#458

PR #458, opened 2017-08-02
fix unfollow errors

Onatcer:
This pull request fixes two problems with the unfollow feature

1. the path of the dialog box changes
2. the script did not unfollow if the first users were not in the pool because it didn't scroll

This closes #456 and closes #453

PR #460, opened 2017-08-02
FIX Problems with Unfollowing #456

lgr:


timgrossmann:
@lgr Thank you for contributing. Unfortunately you've been too late...
#458

PR #465, opened 2017-08-03
Fixed the issue that caused IndexError: list index out of range

mihikadutta:
The issue occured due to change in html structure in the instagram login page. @timgrossmann please review the change and let me know if it looks good. thanks.

### DIFF
--- a/instapy/login_util.py
+++ b/instapy/login_util.py
@@ -20,7 +20,7 @@ def login_user(browser, username, password):
 
     # Enter username and password and logs the user in
     # Sometimes the element name isn't 'Username' and 'Password' (valid for placeholder too)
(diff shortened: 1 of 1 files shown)


brandonkovacs (approved):
This seems to fix the issue! 👍

BernardoGO:
Nice quick fix.
The problem I see with this is that unlike the one seen in d7af29d of #457 this one is very sensible to future changes. Problems related to changes in the HTML structure will probably happen again.

I think that #457 is more of a "permanent" fix for this.

mihikadutta:
@BernardoGO you are right, targeting the username field in the form is the permanent fix. I am closing this PR. 👍

FBosler:
really dumb question. How do I get that installed on ubuntu? I only know how to do the initial setup.

PR #467, opened 2017-08-03
Add div

Bigu93:
Added one more div


BernardoGO:
The PRs #465 and #457 fixes this issue already.
As I've posted in the other one. The problem I see with this is that unlike the one seen in d7af29d of #457 this one is very sensible to future changes. Problems related to changes in the HTML structure will probably happen again.

I think that #457 is more of a "permanent" fix for this.

jpesce (changes requested):
#457 is more of a "permanent" fix for this

PR #470, opened 2017-08-03
Fixes login bug due to instagram change in login screen

jpesce:
Instagram apparently changed the HTML of the login screen resulting in the following error:
```
  File "somefile.py", line 5, in <module>
    session.login()
  File ".../InstaPy/instapy/instapy.py", line 113, in login
    if not login_user(self.browser, self.username, self.password):
  File ".../InstaPy/instapy/login_util.py", line 24, in login_user
    action = ActionChains(browser).move_to_element(inputs[0]).click().send_keys(username) \
IndexError: list index out of range
```
This commit fixes the problem.


jpesce:
Just saw that there are other PR that fixes this.

PR #475, opened 2017-08-04
Fixes the extract problem in How_to_Raspberry.md

BernardoGO:
As seen in #473, the `tar` command was trying to extract an already extracted file if the user had previously extracted another driver.

> gzip: stdin: unexpected end of file
> tar: Child returned status 1
> tar: Error is not recoverable: exiting now


timgrossmann:
@BernardoGO Looks good, thank you!

PR #476, opened 2017-08-04
Removing Firefox "Capabilities" as it seems useless in newer versions

BernardoGO:
Firefox Capabilities doesn't seem to make any difference at all in most of Firefox newer versions.
Furthermore, as seen in #473, it throws an error when used in the newest version.


timgrossmann:
@BernardoGO Looks good, awesome! 
Thank you for the contributions

Issue #444, opened 2017-07-27
Interact with the followers of given list of users


timgrossmann:
> This is the first Issue that will get a bounty on it. (More to come)
This has a high priority.
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=47675897)](https://www.bountysource.com/issues/47675897-interact-with-the-followers-of-given-list-of-users?utm_source=47675897&utm_medium=shield&utm_campaign=ISSUE_BADGE)

The idea is to interact with the followers of a given account. This should be implemented so that I can provide a list of accounts and a number of followers I want to interact with.

E.g. I provide the feature with the list of `['natgeo', 'geo']`, I then want the script to go to these accounts and interact with e.g. 50 of their followers. (Following is highest prio)
The followers can either chosen randomly or simply the top X followers.

>In addition, it would be great to allow the liking+commenting features of InstaPy to interact with this specific “followers of specific users” list as well.

This could looks something like this:

```python
.follow_from_account(['natgeo', 'geo'], amount=50, random=True) 
```

**DoD**:
- I can provide a list of users from which the followers will be followed
- I can provide a number of followers which will be used for each provided account
- I can either choose to do it top-down or "random", which will simply follow some random followers
- The feature can be "easily" extendable with the next "interact with user" feature

gitpatrickhub:
Not sure this is open to additions/feedback, but just in case.

> In addition, it would be great to allow the liking+commenting features of InstaPy to interact with this specific “followers of specific users” list as well."

Support for the following behavior would be great;
For specified IG account scroll follower list for specified number, like 3 most recent posts from each follower and post a comment on 33% and follow 10% --> result for your example would be for 50 random natgeo followers --> 150 posts liked (3 posts per 50 followers) --> 50 posts commented --> 5 accounts followed

timgrossmann:
@gitpatrickhub Yes it is, I just wanted to make several smaller issues instead of one big one.
I think the "interaction" with other users should be a separate method that can then simply be called in this method.

Good to see people interested!

bardascat:
+1 this would be a great functionality..

bardascat:
Also it would be nice to automatically interact with followers of somebody who you are following...

jeffreylm:
Didn't realize that this is what this function does. What would make this function more valuable is if it follows the people that Like or that commented on  a given post or last post.

[...]

PR #477, opened 2017-08-05
Follow someone else's followers

BernardoGO:
**Follow the followers of a given user or follow everyone that a given user is following**
- Follows everyone or an specified amount of users
- Built-in delays following the ones we see in the unfollow function
- Keep track of followed users using the existing log handler
- Sets follow_restrict
- Skips users that are already being followed
- Sleeps for 10 minutes after 10 users
- Makes use of the dont_include list to skip users

**follow_through_dialog function**
- A centralized way to handle all the following tasks done in the dialog list
- Currently being used by both functions above

**formatNumber function**
- Converts to integer numbers like: 2.2K, 1K...
- Added to util.py

**Example for following followers**
- Examples were added

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,28 @@ accs = ['therock','natgeo']
 session.follow_by_list(accs, times=1)
 ```
 
+### Follow someone else's followers
+
+```python
+# Follows the followers of each given user
+# The usernames can be either a list or a string
+# The amount is for each account, in this case 30 users will be followed
+# If random is false it will pick in a top-down fashion
+ 
+session.follow_user_followers(['friend1', 'friend2', 'friend3'], amount=10, random=False)
+```
+
+### Follow users that someone else is following
+
+```python
+# Follows the people that a given users are following
+# The usernames can be either a list or a string
+# The amount is for each account, in this case 30 users will be followed
+# If random is false it will pick in a top-down fashion
+ 
+session.follow_user_following(['friend1', 'friend2', 'friend3'], amount=10, random=False)
+```
+
 ### Unfollowing
 
 ```python
--- a/examples/example.py
+++ b/examples/example.py
@@ -66,6 +66,11 @@ session.like_from_image(url='www.instagram.com/image', amount=100)
 # media filtering works here as well
 session.like_by_tags(['#test'], amount=10, media='Video')
 
+# follows the followers of a given user
+session.follow_user_followers('friend1', amount=10)
+# follows the people that a given user is following
+session.follow_user_following('friend2', amount=10)
+
 session.unfollow_users(
     amount=10)  # unfollows 10 of the accounts your following -> instagram will only unfollow 10 before you'll be 'blocked
 #  for 10 minutes' (if you enter a higher number than 10 it will unfollow 10, then wait 10 minutes and will continue then)
--- a/examples/example.py
+++ b/examples/example.py
@@ -67,9 +67,13 @@ session.like_from_image(url='www.instagram.com/image', amount=100)
 session.like_by_tags(['#test'], amount=10, media='Video')
 
 # follows the followers of a given user
-session.follow_user_followers('friend1', amount=10)
+# The usernames can be either a list or a string
+# The amount is for each account, in this case 30 users will be followed
+# If random is false it will pick in a top-down fashion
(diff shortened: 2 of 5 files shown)


BernardoGO:
Ok, I just realized that @timgrossmann has set a bounty for this feature in #444.
I'll implement the other requirements.

BernardoGO:
**I can provide a list of users from which the followers will be followed**
- Both functions can now accept either list or str for the accounts.

**I can provide a number of followers which will be used for each provided account**
- Already done in the first comment.

**I can either choose to do it top-down or "random", which will simply follow some random followers**
- Both functions will pick the users to follow in a random or top-down fashion, selectable.

**The feature can be "easily" extendable with the next "interact with user" feature**
- Both functions have their core functionality implemented in a single method, making it easy to extend. Since it contains a list of the picked users it is easy to implement callbacks, export a list or add another feature into it. The callbacks can be used to like or comment for a specific user.


Example of simple callback to extend the function:
```
    def printName(name):
        print("Callback: " + str(name))
    followNum = follow_through_dialog(browser, user_name, amount, dont_include, login, follow_restrict, allfollowing, random, callbacks=[printName])

```

So this will basically call the function `printName` for each user it follows. The output will be:

> [bernardo@foobar-pc InstaPy]$ python followExample.py 
> Logged in successfully!
> --> Ongoing follow 1, now following: b'friend2'
> Callback: b'friend2'
> --> Ongoing follow 2, now following: b'friend1'
> Callback: b'friend1'
> --> Total people followed : 2 
> 
> Session ended
> -------------


This is probably all for #444

timgrossmann:
@BernardoGO Awesome! 
I'll try to find time today to test this feature and will get back to this ASAP.

Thank you very much for your contributions.
(Maybe you'll also be interested in working on the next bounty feature as well :P )

timgrossmann:
@BernardoGO What I unfortunately also have to ask for (even though not explicitly stated in the requirements) is some documentation of the feature in the README.

Please add that

BernardoGO:
@timgrossmann Documentation added.
(I'm down for more bounties.)

Another cool thing about this implementation that I forgot to say is that instead of getting the list of everyone in the dialog (as seen in the unfollow function), it get only the ones that are not being followed so there is no risk to pick someone that is being followed already.
It uses a random sample instead of following a top-down fashion with a chance to follow. This provides a better distribution.
It is also guaranteed to follow the specified amount even in random unless there are no users left to follow. If it cannot follow or if the picked user is in the dont_include list it will repick.

The follow_through_dialog() function can be modified to handle the unfollow too. It will fix some of the problems related to the unfollow itself and also make the maintenance easier. If you think it's cool I can make it in another PR later.

timgrossmann:
@BernardoGO I only need to do some local testing to make sure it works and look over your code really quickly.

Thank you so much for making InstaPy a little bit better, I really appreciate all the support and contributions to this project 👍 

The improved unfollow sounds awesome!
If you could make a PR for that as well, I'd be very thankful!

BernardoGO:
The way I'm seeing this is to use that 'follow_throught_dialog" function only to get the users and use the callbacks to do the other interactions.
In this case, even the following part would be moved to a separate callback. 

Tell me what you think about this design. I think it helps to implement new interactions and also select which one to do.
Maybe this next PR should go to the develop instead?

timgrossmann:
@BernardoGO Sounds good to me. I think this design is, as you already stated, a nice decision for the develop branch.

Your PR works like charm. Do you already have a bounty source account?

This fixes #444

timgrossmann:
@BernardoGO Please visit https://www.bountysource.com/issues/47675897-interact-with-the-followers-of-given-list-of-users and "Claim Bounty" 👍 

Thank you for contributing

BernardoGO:
Thanks, @timgrossmann. I have an account, just claimed the bounty.

timgrossmann:
@BernardoGO Should already be paid out 😉

jeffreylm:
This is really a nice system this github. Learning python now and from reading all the fixes and using the scrpts, it really is helping with the learning curve. Just last week had no idea what import did and why it was used, now it's totally clear.  It will be funny some day when I look back and tell my learning story, I will say I learned python through learning to use a opensource bot created by @timgrossmann , hahahahahahaha

thonyhenrique:
Do you add this line of code in quickstart.py?

Issue #484, opened 2017-08-05
Like different pictures of a user's profile


timgrossmann:
> This again has high priority, therefore this will be another **BountyHunt**
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=47972754)](https://www.bountysource.com/issues/47972754-like-different-pictures-of-a-user-s-profile?utm_source=47972754&utm_medium=shield&utm_campaign=ISSUE_BADGE)

### Single use 
*Key to this feature* is that it includes all the `.set_...` calls, therefore it will interact with different pictures of given users but still make sure that nothin inappropriate is liked or commented. (e.g exclude friends, filter for keywords, etc.)

```python
# this will move to the given users profiles and 
# randomly like 10 of his posts taking into account
# restrictions set and e.g. the comment ratio
.like_by_users(['natgeo', 'geo'], amount=10, random=True)
```

### Combined use
In addition to that, we need to extend the `.follow_user_followers` and `.follow_user_following` with a new parameter called `interact` => `.follow_user_followers(..., interact=True)`.
This will, for every user in the list of followers, move to their profile and like the given amount of pictures (also taking into account settings like the `comment`-rate).
To make sure the method call won't get too big, we will introduce a new `.set_...` method which will work as a setup for the `.follow_user_followers` interaction.

> Take into account, that this could also be interesting to be used with e.g. the `.like_by_tag(...)` feature <= make it easily integrateable with the other `.like_by_...` methods

```python
.set_user_interact(percentage=10, amount=10, random=True)
# percentage defaults to 100 so you could also do
.set_user_interact(amount=10, random=True)
```

this will give us the opportunity to do something like this:

```python
# For 10%  of the 20 randomly selected users (~2 users), move to their profile
# and randomly choose 10 pictures to be liked. 
# Take into account the other set options like the comment rate
# and the filtering for inappropriate words or users
.set_user_interact(percentage=10, amount=10, random=True)
.follow_user_follower(['natgeo', 'geo'], amount=20, random=True, interact=True)
```

**DoD**:
- There is a example given in the examples file
- There is documentation added in the README
- I can extend my session with the `.like_by_users()` method
- I can set up my session with the information for the `.follow_user_follower()/following()`
- Both features include the restrictions and settings set up before calling the method (including e.g. comment-ratio and like restrictions)

[...]

PR #491, opened 2017-08-07
Like different pictures of a user's profile

BernardoGO:
This pull request adds the features requested in #484.

**like_by_users(['natgeo', 'geo'], amount=10, random=True)**

- Moves to the user's profile page and like the given amount of photos
- The image picking can be either random or top-down.
- Includes the restrictions and settings set up before calling the method (including e.g. comment-ratio and like restrictions)
- Clarifai is being used
- Uses built-in follow_user and like_image
- Uses built-in log system

**set_user_interact(amount=10, random=True)**
- Set interaction parameters for `follow_user_follower(['natgeo', 'geo'], amount=20, random=True, interact=True)` and `follow_user_following(['natgeo', 'geo'], amount=20, random=True, interact=True)`
- In this case, for every user of the 20 randomly selected, move to their profile and randomly choose 10 pictures to be liked. 

**get_posts_by_user()**
- Moves to a user profile and returns the posts.
- Scrolls automatically 

**follow_user_followers**
- Now includes interact parameter, calls like_by_user

**follow_user_following**
- Now includes interact parameter, calls like_by_user

**There is a example given in the examples file
There is documentation added in the README**

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ session.like_by_tags(['#cat'], amount=15, media='Video')
 session.like_from_image(url='www.instagram.com/image', amount=15, media='Video')
 
 #Likes 10 random pictures of user's profiles
-session.like_by_users(['natgeo', 'geo'], amount=10, random=True)
+session.like_by_user(['natgeo', 'geo'], amount=10, random=True)
 
 session.end()
 ```
--- a/README.md
+++ b/README.md
@@ -108,6 +108,9 @@ session.like_from_image(url='www.instagram.com/image', amount=50, media='Photo')
 session.like_by_tags(['#cat'], amount=15, media='Video')
 session.like_from_image(url='www.instagram.com/image', amount=15, media='Video')
 
+#Likes 10 random pictures of user's profiles
+session.like_by_users(['natgeo', 'geo'], amount=10, random=True)
+
 session.end()
 ```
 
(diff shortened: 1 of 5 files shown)


ahmadudin:
you're fast

timgrossmann:
@BernardoGO You guys really are fast!
Perfect, thank you for this. I'll look into it and get back to you.

Thank you for contributing.

timgrossmann:
@BernardoGO I just tested it and it already works very well.

Here are a few things I noticed:

- If random is True for the user interaction, it will load more and more without taking into account how many pictures I actually want to like (check out the get_links_for_username method at the other PR) This will take very long with accounts like natgeo which have 14k posts.
- change the like_by_user() to like_by_**users** to keep the interface a bit more consistent
- While testing I ran across a "Sorry this page is not available"-Page which led to the script aborting instantly. Please make sure that this will skip the current user and continue on with the next one.

This is definitely my fault,  but we need to add another parameter to the `set_user_interact`-method, I updated the issue for the requirements.
- add another param which is `percentage=...` like on a lot of the other methods.
Me might want to follow 250 of the given users followers but only interact with ~10% of them.

Thank you very much for contributing!

ahmadudin:
Very well. For adding percentage param, it just came to my mind when i was about to use the feature for real purpose, not just testing.

timgrossmann:
@ahmadudin Haha, yes the same happened to me 😅 
I just then realised that this would be important.

timgrossmann:
@BernardoGO Thank you for contributing, I hope, even though @ahmadudin PR was merged, you'll consider contributing again in the future.

stephane-archer:
is the action `like_by_users` available? @timgrossmann @ahmadudin @BernardoGO

Issue #506, opened 2017-08-08
Limit the number of items for feedlike


sojan-official:
currently the feed like function only has one parameter which takes the number of likes and the script will keep on executing until that number is reached.
what about about another parameter which will abort the function after scrolling through n number of images in feed

use case 
------------------------
- i want to automate like pictures on my feed  by setting up a crontab every hour
- the  parameter i passed is say 10 
- but if in the half hour less than 10 images were posted, the script will keep getting executed forever

sojan-official:
i think with the latest implementation this shouldn't be an issue. so closing this

PR #508, opened 2017-08-08
possible solution for  #506

sojan-official:
This is a quick fix i use for #506 .
Let me know if we are planning to take this upstream and if there is a better way to do this. 
will be happy to contribute the code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,9 +235,9 @@ Example:
 
 ```python
 #This is used to perform likes on your own feeds, amount specifies how many total likes you want to perform
-
-session.like_by_feed(amount=100)
+session.like_by_feed(amount=100, scrollLimit=None)
 ```
+`.like_by_feed` keeps scrolling through your feed until the likes are performed. You can optionally pass `scrollLimit` parameter to limit the number of scrolls performed.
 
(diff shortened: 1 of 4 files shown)


timgrossmann:
@sojan-official Hmm... Now I understand what you're requesting.

If you can add the right documentation for this, this could be interesting.
Add an example in the example.py
Add documentation in the Readme
Add the param to the method header where the two other parameters are described

I think key to this feature will we that it's documented to be understood.
Make sure people understand what the scroll is for.

Also make sure that the parameter stays optional, meaning that if you leave it out, the script keeps working as it is right now.

Thank you for contributing

sojan-official:
@timgrossmann the mentioned changes are performed. Let me know if i missed something.

timgrossmann:
@sojan-official Thank you for the changes! I'm really sorry that there are quite a few conflicts now, but the feed interaction changed a bit...

sojan-official:
@timgrossmann  no problem. will look through the changes and see what can be done.

sojan-official:
i think with the latest implementation this shouldn't be an issue. so closing the request

Issue #530, opened 2017-08-12
User customizable delays


paunocu69:
Hello everyone,

Today I wanted to run a quick demonstration to my clients and the long delays was a pain in the ass.
I know it's useful to avoid being banned and shit but sometimes we just need to test or to show the thing working and we will have to wait too long for it.

My idea is to have something like a method called set_delay_reduction so that we can set a percentage to reduce every sleep in the bot.
This would really make my weekend if implemented

I'm willing to pay for this to whoever comes with a solution

sojan-official:
InstaPy has option for passing `page_delay` as a param. you can reduce the default 25 sec delay using that.

`session = InstaPy(username='username', password='password', page_delay=5)`

paunocu69:
This is not the delay I've meant. The one you've said is relative to the time the browser will wait to least I think.
The one I'm asking for is relative to the sleeps

ahmadudin:
@paunocu69 good idea

converge:
now, follow_user_follower, follow_user_following and unfollow_user has this option sleep_delay.

PR #547, opened 2017-08-13
User customizable delays

ahmadudin:
**add feature requested on #530 by @paunocu69**
- Add` sleep_delay` parameter for custom user defined sleep duration
- Input value in second, default `sleep_delay=600`
- example and README updated

### DIFF
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -1,6 +1,7 @@
 """Module which handles the follow features like unfollowing and following"""
 import json
 import csv
+from math import ceil
 from .time_util import sleep
 from random import randint
 from .util import delete_line_from_file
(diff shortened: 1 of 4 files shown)


timgrossmann:
@ahmadudin Awesome! Thank you.
I'll try this out later today.

Thank you for contributing!

instapy/unfollow_util.py
```diff
@@ -179,7 +180,7 @@ def follow_given_user(browser, acc_to_follow, follow_restrict):
         sleep(3)
         return 0
 
-def follow_through_dialog(browser, user_name, amount, dont_include, login, follow_restrict, allfollowing, is_random, callbacks = []):
+def follow_through_dialog(browser, user_name, amount, dont_include, login, follow_restrict, allfollowing, is_random, delay, callbacks = []):
```
timgrossmann, line 183: <strike>I think we should keep a default value so the users don't need to provide a value if they don't want to</strike>
Sorry just realised that it's passed through 😉

timgrossmann (approved):

instapy/unfollow_util.py
ahmadudin: :)

Issue #509, opened 2017-08-08
Question re: Interact with someone else's followers/following


gitpatrickhub:
Does this feature support only interacting with someone else's followers? I.E., I don't want to follow any of the provided user accounts followers, I only want to interact with the followers by liking & commenting on posts. If so, how do you configure the lines below?

```
session.set_user_interact(amount=5, random=True, percentage=50, media='Photo') 
session.follow_user_followers(['friend1', 'friend2', 'friend3'], amount=10, random=False, interact=True)
```

timgrossmann:
@gitpatrickhub Unfortunately, this feature is not supported yet...

gitpatrickhub:
@timgrossmann can this be added to the list of features to be created? This would be very helpful feature, as not all InstaPy users want to systematically follow or unfollow users.

timgrossmann:
@gitpatrickhub Hmm... I think we can at least keep this issue open.
If you posted this earlier, we could've made this part of #492

gitpatrickhub:
@timgrossmann thanks! I requested it, but failed to explicitly state that following should not be required to support the "interact with".

datteba:
Really interested in this too! I only want to like posts of user followers without follow them. Possibly I would like to always like also the last uploaded photo. (ignoring private profiles)

@timgrossmann I could pay 10$ for this implementation

gitpatrickhub:
@timgrossmann @datteba I'd be willing to contribute, as well, to get this delivered soon

timgrossmann:
@datteba @gitpatrickhub Well, then let's open up a bounty to get this implemented ASAP 😉 

I think with more than 10$ we'll get this implemented in no time.
I'll create a bounty issue for that soon.

timgrossmann:
@BernardoGO <strike>Would you be interested in refactoring your `follow_user_followers/following` a little bit? 
25$ Bounty? 😉 </strike>

BernardoGO:
Yes, I'm interested. 
I'll start working on it later today. Thanks

timgrossmann:
@BernardoGO <strike>Oh perfect, I'll write an Bounty Issue for it.
This will be the first issue where some of the interface will be changed.
We will move from a `.like_by...` and `.follow_...` to a global `.interact_with`

I think this will be way better since people want to choose on what they do and don't, e.g. like in this example where they only want to get all the users and like their stuff without following.
Or the people that want to only follow from tags instead of always liking and maybe following.</strike>

[...]

PR #552, opened 2017-08-14
Fix for the math.ceil problem #509

BernardoGO:
It seems like in Python 2 math.ceil actually returns a float instead of int, which causes problems for the random.sample function.

This fixes the error mentioned in #509 by parsing the returned value to int.


timgrossmann:
@BernardoGO Awesome, thank your for the fix.

PR #553, opened 2017-08-15
Paunocu's fix the sleep ceil

paunocu69:
Its a very nice a pull request it fixes the problem "can't multiply sequence by non-int of type 'float'" while adding the option for non integer sleeps.
There is no reason to force the sleep to be a integer. it also doesnt work on some operational sysmtem


paunocu69:
i di not fix the same fix as #552 it is another one for another part that isnt related along eachother

Issue #352, opened 2017-06-29
Unfollow not followers


emKaroly:
How can i achieve this functionality?

emKaroly:
If i think about it, it can be done with adding current followers to set_dont_include and let InstaPy unfollow as much it can. It would be nice if the [instagram-profilecrawler](https://github.com/timgrossmann/instagram-profilecrawl) could return the followers list also without going through all my post for advanced statistics.

timgrossmann:
@emKaroly Well that's a hard and dirty get around 😄 
will work though

wesleyabbey:
In order to get around this, I went to my followers on Instagram on my computer and then pushed all of them into a comma delimited string.

I pushed this string into a file called `followers.txt` and in my quickstart.py I read from that file. I can recreate this and post a solution or possibly try to create a script for this if enough people want it.

Edit: Actually it looks like we might linearly read from the python list of friends. I'm not sure about what sort of performance impact this might have for large enough friends list but maybe it could be worth exploring creating a dictionary instead of a list of friends?

deronsizemore:
>I went to my followers on Instagram on my computer and then pushed all of them into a comma delimited string.

Can you explain how you went about this?

datteba:
I'm also interested in this feature! I would like to unfollow users that are not following back **after more than 24 hours**.

Is there anyone who wants to contribute to a bounty to develop this feature quickly? 

@emKaroly @deronsizemore @wesleyabbey @timgrossmann

converge:
#695 #581 #221 #707 -> all about this feature.

I'm working on it, and need some help ( #707 )

this is my progress:  https://github.com/converge/InstaPy/commit/c7897b9ddf376ecbf2ce6b5dd8df3530c8f5eba2

converge:
1. I tried to scroll down until the bottom at "Followers" list, but it takes too long to complete and it does a lot of server requests. bad approach.
2. Now I'm getting just some pages at "Followers" list and adding it to a file, and telling unfollow() function to now unfollow usernames in this file.
3. I still need to update this file to remove users that aren't following me anymore, how could I do that ?

pleum:
i try to get follower and following list by api that scroll down fetch, need some test: https://github.com/pleum/InstaPy/commit/883c36cdea73f7018a4b7d2e1febc3525e9a469a

PR #560, opened 2017-08-16
Ensuring the script gets the amount needed to follow

ahmadudin:
- Add simple algorithm to check if amount of user needed to follow is sufficient
- Expand the amount when `random=True` for better sampling distribution

### DIFF
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -185,6 +185,10 @@ def follow_through_dialog(browser, user_name, amount, dont_include, login, follo
     sleep(2)
     person_followed = []
 
+    if random:
+        # expanding the popultaion for better sampling distribution
+        amount = amount * 3
+
     # find dialog box
     dialog = browser.find_element_by_xpath('/html/body/div[4]/div/div[2]/div/div[2]/div/div[2]')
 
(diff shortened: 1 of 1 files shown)


timgrossmann:
@ahmadudin Thank you! Will check this ASAP

timgrossmann:
@ahmadudin Do you think there is a way to incorporate a new parameter which will let people decide if they want to include all of their followers/following in the `random=True` selection?

e.g. if I want 10 random of my 500 followers, I can simply provide that param and it will scroll until all are loaded? And only then select the 10?

ahmadudin:
@timgrossmann wait, what method this kind of `random=True` for?

timgrossmann:
@ahmadudin It would be interesting for the `.interact_user_...` and `.follow_user_...` the used method is called `scroll_bottom`.

ahmadudin:
@timgrossmann Oh i see, so the parameter will prevent bananas scrolling on account with millions follower/following if including all the followers/following function implemented by default(?)

timgrossmann:
Yes, right now, there is a check if the amount is greater than 50, if it is, it will simply set it to 50 (the scroll_bottom method in utils.py) 

This is bad default behavior since it will simply stop scrolling even if you provide a large enough number like 500.

So e.g. I want to do a follow_user_followers with 500 (random=False) it will simply stop at 270 and say that you've requested more than there are followers. 

This limit is great for accounts with lots and lots of followers, since loading all of them would take hours, therefore it would be nice if we could add a param that forces the scroll_bottom to really go as far as the amount states and not cut it off.

I hope this is somewhat understandable.
I'll try to come up with a good example soon.
Sorry if there are any confusions

timgrossmann:
@ahmadudin Here are a few examples of what I mean:

1. The user wants to use the `.interact_user_followers` feature at a big account with 2Mio followers, in this case the current behaviour is nice because it makes sure the script won't take forever to load all the followers. (It still will only do 50 scrolls to the bottom, because this is the max amount decided in the `scroll_bottom` method <= this definitely needs to be changed) 

2. The user wants to use his own followers/following, in this case he might want to get e.g. 50 random people of his whole followers list. Right now, when providing 50 in the `.interact_user_...` methods, it will simply get a few more and select 50 of them to not only take the latest. But it will never aggregate all the followers/followings before selecting the 50 people.
He himself might only have 500 followers, therefore he wants to use all of his followers instead of only 50 of the 50*3 aggregated.

It might be a good idea to simply create a followers and following list once in a while which will also help us to create features like #352 (unfollow not followers)

ahmadudin:
@timgrossmann sorry, i've been away for a while. For the first example, i think my pull request is already solves it. And for the second i agree, its a good idea to involve the whole followers/following into the population to be sampled, but i think it will be on different method.

PR #563, opened 2017-08-17
Change click() to .send_keys("\n")

westondeboer:
If the button is covered for some reason. The error "Element ... is not clickable at point" shows up. This bypasses that and clicks no matter what. I am running this on OS X, and Ubuntu. Don't know if it will work elsewhere.

### DIFF
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -9,6 +9,7 @@ from .util import scroll_bottom
 from .util import formatNumber
 from .print_log_writer import log_followed_pool
 from selenium.webdriver.common.keys import Keys
+from selenium.webdriver.common.action_chains import ActionChains
 import random
 
 def set_automated_followed_pool(username):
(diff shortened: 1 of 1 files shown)


instapy/unfollow_util.py
```diff
@@ -168,7 +169,7 @@ def follow_given_user(browser, acc_to_follow, follow_restrict):
     follow_button = browser.find_element_by_xpath("//*[contains(text(), 'Follow')]")
     sleep(10)
     if follow_button.text == 'Follow':
-        follow_button.click()
+        follow_button..send_keys("\n")
```
timgrossmann: This won't work, since there are two .'s there.

Also, if you want this to be merged, please change this on all the methods in the file.
Thank you

timgrossmann (changes requested):

timgrossmann (approved):

timgrossmann:
@westondeboer We still can't only let this be there for a single method 😉

timgrossmann:
#589

PR #578, opened 2017-08-23
Develop

petriichuk:
Want to add an extra method for getting results to third-party apps


petriichuk:
I think I did wrong Pull request

PR #579, opened 2017-08-23
Added a new method to get results for third-party apps

petriichuk:


timgrossmann:
@petriichuk What is the use case for this method if it's not written out to a file?

Also, I think it would be more convenient to return an object instead on an array so we can get the information we wan't by key and don't have to take the index of the array.

petriichuk:
@timgrossmann 
1) because you can call it from another application ( I use Django), and save results (answer) to database.
2) You are right, it can be done as an object, it was just quick solution, gonna rewrite it

petriichuk:
@timgrossmann or better to make just those internal variables as attributes of the class (I mean how many likes/followers ...etc per session)? 
I need it because I created an application using Django... I created tasks there, every task is a new session, so I need to get back results and show in my interface.

timgrossmann:
@petriichuk Could you show that Django application?

petriichuk:
@timgrossmann yes, will post it to github this weekend

timgrossmann:
@petriichuk Could you also add documentation for this feature? 😉

petriichuk:
@timgrossmann there are two things I modified in the original instaPy file to use it in my django project.
1) An extra method, or get as properties 
2) There is a path to log file is written as double "../", if I call it from django, I try to access log file from a wrong place... 

If you have any suggestions how we can solve, feel free to write it here

timgrossmann:
#839

Nocturnal-2:
@petriichuk I am trying to integrate InstaPy into Django for the same reason you have mentioned (ease of saving into database). Did you manage to do it successfully? I would like to know about the changes you have mentioned in your [latest comment above](https://github.com/timgrossmann/InstaPy/pull/579#issuecomment-325158580) :) Especially the extra method that you have added to the instapy.py

PR #586, opened 2017-08-24
"Too few images" is not a good reason to abort

iSimonWeb:
As it is for "like_by_users", if a tag or a location has too few images the script should just "continue" iterating the array.


timgrossmann:
@iSimonWeb Thank you! That was on my Todo for a pretty long time...
Would you be interested in refactoring this for the other methods as well?

iSimonWeb:
@timgrossmann I've checked every file and the only other method that needs to "continue" (like_by_users) already does it.
Is there anything more I can do for this awesome project?

PR #585, opened 2017-08-24
Fixed error with searching 'Follow' and 'Following' buttons

dbknv:
I found issue when I'm using like by tag. For example:
`session.like_by_tags(['hashtag'], amount=1)`

The main problem was with "find_element_by_xpath", after my fix it starts work properly.

PR #587, opened 2017-08-24
follow_user and follow_given_user 	Click on button via js

iSimonWeb:
I think I've solved the issue with the un-clickable Follow button by just trying to find it by text (as it yet was for follow_given_user).
I hope I've done right with the code since its my first step into Python.

### DIFF
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -131,7 +131,7 @@ def follow_user(browser, follow_restrict, login, user_name):
     try:
         follow_button = browser.find_element_by_xpath("//*[contains(text(), 'Follow')]")
         sleep(2) # Do we still need this sleep?
-        follow_button.click()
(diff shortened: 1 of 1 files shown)


timgrossmann:
@iSimonWeb Thank you for contributing, there are a few PR's now that do quite the same, I'll check them all out and merge then 😉

timgrossmann:
@iSimonWeb I just don't think that clicking the buttons with JavaScript is such a good idea 😅

timgrossmann:
#585

iSimonWeb:
@timgrossman I'm using it this way and fortunately my account is still up. I give you that it could be a way to get busted, you can remove it if you thinks so.

timgrossmann:
@iSimonWeb Yes, I already merged your PR and replaced it with the `send_keys('\n')` version 😉 

Thank you for helping out!
Looking forward to future contributions

PR #589, opened 2017-08-25
Replace instances of click() with send_keys

Jayphen:
This may fix the potential issues with the script trying to click DOM elements that are obscured.


Jayphen:
Note this is the same logic as #563, but applied across the whole file

timgrossmann:
@Jayphen I'll take a look at it soon.

Thank you very much for contributing.

PR #591, opened 2017-08-26
Fix is_private verification

TGlide:
Change the h2 class attribute to match new class naming in instagram.com, as to mantain functionality of verification of a user's private status.


timgrossmann:
@TGlide Awesome, thank you!

PR #600, opened 2017-08-28
Inverse order of users of automatedFollowedPool

thomasuno:
I think it would be better to invert the csv of automatedFollowedPool to remove before older followed users.

### DIFF
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -16,7 +16,7 @@ def set_automated_followed_pool(username):
     try:
         with open('./logs/' + username + '_followedPool.csv') as followedPoolFile:
             reader = csv.reader(followedPoolFile)
-            automatedFollowedPool = [row[0] for row in reader]
+            automatedFollowedPool = [row[0] for row in reversed(list(reader))]
 
         print("Number of people followed automatically remaining: {}".format(len(automatedFollowedPool)))
         followedPoolFile.close()


timgrossmann:
@thomasuno Could you introduce a new parameter from `.unfollow_users` here?
So I could do something like:

```python
.unfollow_users(amount=10, onlyInstapyFollowed = True, method='fifo') # or lifo for "last in first out (how it is currently)
```

Also, would be awesome if you could add a check that the method only works when you enabled the `onlyInstapyFollowed`

Thank you for contributing!

sionking:
Thanks

dbknv:
I think your code will not work. The main reason is **automatedFollowedPool** array is not using as list of users in **unfollow** function. It's only checking in 
`if onlyInstapyFollowed == True and person in automatedFollowedPool:`

So you need to sort users in the right order **person_list** witch are populated from **person_list_a**.  And after that it will be possible use FIFO or LIFO in the cycle of each user in **person_list**

timgrossmann:
#643

thomasuno:
Sorry if I havent answered.
@timgrossmann I did not find time to develop this feature but I saw @sterid did.
I also find it a bit strange to unfollow people you have just followed, I mean, why you cam choose between FIFO and LIFO? For me it was just a development mistake.

For @sterid My code is working and I'm using it.

Issue #606, opened 2017-08-30
On following followers error in Chromedriver


y471n:
`('following_link error \n', 'Message: unknown error: cannot focus element\n  (Session info: chrome=60.0.3112.113)\n
  (Driver info: chromedriver=2.29.461571 (8a88bbe0775e2a23afda0ceaf2ef7ee74e822cc5),platform=Linux 4.9.0-3-amd64 x8
6_64)\n')`

`instapy/unfollow_util.py", line 395, in follow_given_user_followe
rs`

`/unfollow_util.py", line 190, in follow_through_dialog
    dialog = browser.find_element_by_xpath('/html/body/div[4]/div/div[2]/div/div[2]/div/div[2]')`

`local/lib/python2.7/site-packages/selenium/webdriver/remote/e
rrorhandler.py", line 194, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.NoSuchElementException: Message: no such element: Unable to locate element: {"method":"x
path","selector":"/html/body/div[4]/div/div[2]/div/div[2]/div/div[2]"}
  (Session info: chrome=60.0.3112.113)
  (Driver info: chromedriver=2.29.461571 (8a88bbe0775e2a23afda0ceaf2ef7ee74e822cc5),platform=Linux 4.9.0-3-amd64 x8
6_64)`

timgrossmann:
@iSimonWeb Could you submit a better way to find that element as well? 😉

iSimonWeb:
@timgrossmann I'm on it 👍

y471n:
@iSimonWeb - Probably the div with `role="dialog"` could serve as a point for finding the element.

iSimonWeb:
@y471n I've tried selecting by attribute but it wasn't working.
Anyway I've found an even simpler and lightweight solution with #607 😄

timgrossmann:
@y471n Could you please do a git pull and test again 😉 Seems to be working fine on my side

nnvum:
I had the same issue and seems to be working for me as well, thank you.

Edit: I just noticed that "session.set_upper_follower_count()" is ignored for the following action and I cannot tell whether that was already the case at an earlier stage as it seem to be working for the like actions.

PR #607, opened 2017-08-30
Fixed XPATH selectors for followers/following and dialog boxes

iSimonWeb:
As issued on #606, fixed unfocusable elements when following other users' followers/following.


timgrossmann:
@iSimonWeb So awesome, thank you very much!

PR #610, opened 2017-08-30
Use send_keys on like buttons too

iSimonWeb:
Following the idea from the pull request #589 we should get rid of the click triggered via execute_script.
I'm actually asking myself why we were getting the className from the xpath element and then using javascript to get it again by className to finally click on it 😆

### DIFF
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -425,7 +425,7 @@ def check_link(browser, link, dont_like, ignore_if_contains, ignore_users,
 
 def like_image(browser):
     """Likes the browser opened image"""
-    like_elem = browser.find_elements_by_xpath("//a[@role='button']/span[text()='Like']")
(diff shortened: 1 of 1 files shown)


iSimonWeb:
Sorry for the open/close/reopen mess 😢
I was missing a parent selector `/..` to get the anchor and correctly send_keys:
`"//a[@role='button']/span[text()='Like']/.."`

Committed the fix to my patch-1 branch.

timgrossmann:
@iSimonWeb I'll be on vacation for a few days and won't have the time to test this...

@ahmadudin @BernardoGO Could you quickly test if this is working so I can merge it, please?

ahmadudin:
@timgrossmann i have tested it and it works great, haven't found any problem with it

timgrossmann:
@ahmadudin Thank you!

PR #611, opened 2017-08-31
Fix following link

aellispierce:
The unfollow method is being sent `username` as a param but inside the method we're looking for `user_name` creating an error `('following_link error \n', "global name 'user_name' is not defined")`.

This bug was introduced today in https://github.com/timgrossmann/InstaPy/commit/a6f93227c4ea8cb5cae90b4523b021c632273d21


iSimonWeb:
Pardon, my fault 🙏
I used the multiple selection to replace every similar xpath selectors and I forgot to check the valid variable to use.

PR #616, opened 2017-09-01
Fix random on follow through dialog

y471n:
Fix random variable reference in follow_through_dialog.


timgrossmann:
@y471n Good catch, thank you!

PR #663, opened 2017-09-15
Update and rename How_To_DO_Ubuntu.md to How_To_DO_Ubuntu_on_Digital_…

alx-a:
…Ocean.md

clarified that the guide is for DO in the file title, commented that chrome browser is only available 64-bit.


timgrossmann:
@alxnorden Thank you

PR #669, opened 2017-09-18
Removed /requirements.txt from dockerignore

brammittendorff-dd:
Removed /requirements.txt from dockerignore or else the docker-compose.yml won't build. The pip install -r requirements.txt command fails when /requirements.txt is ignored.

- Even the COPY command will fail fstat because there is no /requirements.txt.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,7 +4,6 @@
 /docker_conf/
 /assets/
 /setup.py
-/requirements.txt
 /LICENSE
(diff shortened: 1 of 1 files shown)


timgrossmann:
@brammittendorff-dd Thank you!

brammittendorff-dd:
No problem :smile: I really like the work you did on this repository, so thank you!

PR #674, opened 2017-09-19
Updated docker-compose.yml for setting hostname default

brammittendorff-dd:
Docker generates hostnames that will not match:
- http://selenium:4444/wd/hub

So now we set the `container_name: selenium` to a default value so it will work. No need for setting the ip anymore.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,6 +10,7 @@
      depends_on:
        - selenium
    selenium:
+       container_name: selenium
        image: selenium/standalone-chrome-debug
        ports:
          - "5900:5900"


brammittendorff-dd:
To make docker-compose work we needed a correct hostname (on debian it will generate them like `instagramselenium_selenium_1`). On ubuntu it will generate a correct name like `selenium`. With this PR I fixed the hostname (also in debian) `container_name: selenium`.

But the web container needs to wait for the selenium container to be started at port 4444. (a other problem) The web container sometimes starts **BEFORE** the selenium container. The solution in docker-compose 2.1 is to add a healthcheck. 
```version: '2.1'
services:
  web:
    build:
      context: .
      dockerfile: ./Dockerfile
    depends_on:
      selenium:
          condition: service_healthy
    command: 'python /app/start.py'
  selenium:
    container_name: selenium
    image: selenium/standalone-chrome
    shm_size: 128M
    healthcheck:
      test: ["CMD", "wget", "-O-", "http://selenium:4444/wd/hub"]
      interval: 1s
      timeout: 15s
      retries: 15
```

But in docker-compose 3 it is a bit different:
https://docs.docker.com/compose/startup-order/

Conclusion:
I will create a new PR with a bash file to check if selenium:4444 exists.

PR #680, opened 2017-09-20
Added a like-limit via sqlite

snakelab:
hi tim, please check my code.
I would recomend to add sqlite as a replacement for the text-files in the script - this could be rolled out step by step.

## Short description of testing:

1. run your script as usual
2. during the run check the sqlite installation/init like this:
```
pi@SNAKE-RASPBERRY:~/instagram/InstaPy $ sqlite3 db/instapy.db
SQLite version 3.20.1 2017-08-24 16:21:36
Enter ".help" for usage hints.
sqlite>
```
3. Next is to get the count of the likes as every like is saved:
```
sqlite>SELECT COUNT(*) FROM likes WHERE liked = date(date('now'), "-0 day");
538
```
4. Works

### DIFF
--- a/db/instapy.schema
+++ b/db/instapy.schema
@@ -1,7 +1,24 @@
-create table likes (
-	id        	integer primary key autoincrement not null,
-    	insta_user    	text,
-  	insta_name    	text,
-    	link    	text,
-    	liked        	datetime
+CREATE TABLE IF NOT EXISTS likes (
+	id integer primary key autoincrement not null,
+	insta_user	text,
+  insta_name	text,
+	link	text,
+	liked  datetime
 );
+
+CREATE TABLE IF NOT EXISTS comments (
+	id integer primary key autoincrement not null,
+	insta_user	text,
+  insta_name	text,
+	link	text,
+	commented  datetime,
+  comment text
+);
+
+CREATE VIEW IF NOT EXISTS v_today 
+AS 
+    SELECT * FROM 
+        (SELECT COUNT(*) AS likes_today FROM likes b WHERE date('now') = date(b.liked)), 
+        (SELECT COUNT(*) AS likes_hourly FROM likes WHERE strftime('%H', liked) = strftime('%H', 'now') and date(liked) = date('now')),
+        (SELECT COUNT(*) AS comments_today FROM comments a WHERE date('now') = date(a.commented)),
+        (SELECT COUNT(*) AS comments_hourly FROM comments WHERE strftime('%H', commented) = strftime('%H', 'now') and date(commented) = date('now'))
--- a/db/instapy.schema
+++ b/db/instapy.schema
@@ -1,0 +1,7 @@
+create table likes (
+	id        	integer primary key autoincrement not null,
+    	insta_user    	text,
+  	insta_name    	text,
+    	link    	text,
+    	liked        	datetime
+);
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -170,10 +170,6 @@ class InstaPy:
             self.logFile.write('Wrong login data!\n')
             self.aborting = True
 
-        elif likes[0] > self.limit_likes:
-            print("Enough likes for today - EXIT")
-            self.aborting = True
-
         else:
             print('Logged in successfully!')
             self.logFile.write('Logged in successfully!\n')
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -42,7 +42,6 @@ from .unfollow_util import set_automated_followed_pool
 import random
 import os
 
-
 class InstaPy:
     """Class to be instantiated to use the script"""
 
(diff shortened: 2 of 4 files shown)


timgrossmann:
@snakelab Thank you for contributing.

I will check that as soon as I can find some time, looks awesome!
If there are users that might can test it I'd really appreciate it.

timgrossmann:
@snakelab We probably also need to add the python sql driver to the requirements file.

snakelab:
Done

Naramsim:
Would be useful to have it while using Docker, we need to install SQLite in the Dockerfile or add a SQLite container to the compose file

dbrkv:
I'm not sure about `sys.exit(0)` when limit reached. If use InstaPy as standalone - not a big deal, but in integration with other apps it can be a problem.
This feature must be optional I guess.

snakelab:
@Naramsim Wouldn't it be the right way to define a volume 'db' and log it to the host?

@sunthera Well, it was a first shot. What else should/could happen at the stage of 'like-limit exceeded'?

I did work on this PR and added some 'features' like a comment-limit log and one db for user which i want to add soon.

dbrkv:
@snakelab probably script must just exit from current task (break loop and print result)
I'm not sure how `self.aborting` works, but looking at code `like_by_tags` it is placed at beginning of method. Maybe if liked > limit => `self.aborting = True` can work. 
But in the same time need notify about limit reached.

Naramsim:
@snakelab  Yep, we don't need to exit the script. If someone is using Schedule to schedule its jobs, you completely kill everything.

dbrkv:
This feature is pretty good and useful, but it will be much more flexible if it can be placed outside `InstaPy` class code.
I explain why: for example, I have RPC server providing interface for `InstaPy` instance. And in this case other app can control execution and collect information about likes, comments, limits and make desision to stop or continue work.

converge:
In this way the bot can evolute and we can develop a frontend with history and reports.

snakelab:
I did some changes on the exit code, now just self.aborting is set.
The init-db is now idempotent and will be ran at every run to be able to migrate :-)
@sunthera 
What do you mean exactly with "outside of InstaPy classcode"? The logging of this actions must be in the code, the analytic of the numbers can allways be done outside.

converge (commented):
what do you think about not have sql code inside bot files ? we could organize all sql calls inside an util file

dbrkv:
@snakelab I mean good if source code of the Bot can be more flexible for extending.
For example, I extended bot class and did mostly same as you: added code to break loop  after `like_image()` method. But the problem is: I had to override all `like_by_tags()` and other big methods  and now it hard to support my bot code. If move image liking into smaller method, we can avoid overriding big methods

```python
class InstaPy:
 
    def like_by_tags(self, tags=None, amount=50, media=None):
        ....
        for index, tag in enumerate(tags):
            ....

            for i, link in enumerate(links):
                ...
                try:
                    ...
                    if not inappropriate:
                        try
                          liked = self._do_like(self.browser)
                        except LimitReached as e:
                            return
                    ....
                    ....
                        

    def _do_like(browser)
        liked = like_image(browser)
        if no_more_likes_for_today:
            raise LimitReached
```

Code can throw an exception  from `_do_like`, catch it inside methods like `like_by_tags`and make decision by conditions what to do next: exit from method for example.

Then easy:

```python
class MyBot(InstaPy):        

    def _do_like(browser)
        amount = self._lookup_like_amount_sqlite()

        if amount > 42:
            raise LimitReached
        
        liked = like_image(browser)
        

class MyBot(InstaPy):        

    def _do_like(browser)
        amount = self._lookup_like_amount_remote_api_call()
        
        if amount > 42:
            raise LimitReached
        
        liked = like_image(browser)
          
```

UPDATE:

Probably we can write class named `DecisionMaker` (or with better name and functional :) ), inject it as a parameter into `InstaPy` and then use it to control instance behavior.

```python
from abc import ABC, abstractmethod

class DecisionMaker(ABC):

    @abstractmethod
    def add_liked():
        pass
    
    @abstractmethod
    def can_continue():
        pass


class SqliteDecisionMaker(DecisionMaker):

    def add_liked():
        self._conn.execute('INSERT INTO LIKES ...')
    
    def can_continue():
        self._conn.execute('SELECT * FROM ...')


class MysqlDecisionMaker(DecisionMaker):

    def add_liked():
        self._conn.execute('INSERT INTO LIKES ...')
    
    def can_continue():
        self._conn.execute('SELECT * FROM ...')
```

snakelab:
@sunthera that sounds great. thats the missing puzzle part i was searching for. Lets make a first version of this idea....

dbrkv:
@snakelab you mean first version with moving into smaller methods? 
Idea with "State Machine" is very raw, maybe event based system is better.

snakelab:
1. Step would be the smaller methods
2. Integrating state_machine in the relevant methods
3. Adding sqlite on top

The last 2 points could i do, who can do the first one? ;-)

snakelab:
seems too crazy for you folks...

PR #704, opened 2017-10-02
Removed maximze window this causes problems with docker

brammittendorff:
There are issues with docker and the maximizing of your chrome window:
https://github.com/SeleniumHQ/docker-selenium/issues/396

I am having these problems on Ubuntu and Debian and arch.

We are using the standalone version, so I suggest to remove it or set a default.


GOliveira10:
I'm still having this issue when I push to Docker. Anyone else?

PR #713, opened 2017-10-04
Use Travis CI to run flake8 tests on pull requests

cclauss:
The owner of the this repo would need to go to https://travis-ci.org/profile and flip the repository switch __on__ to enable free automated flake8 testing of each pull request.

The output will look like: https://travis-ci.org/cclauss/itinerant-tester/builds/283213087

PR #718, opened 2017-10-05
Fix undefined name found by flake8

cclauss:
__selenium.common.exceptions.NoSuchElementException__ is an undefined name in this context but line 7 does define __NoSuchElementException__ by importing it from __selenium.common.exceptions__.  Undefined names can raise [NameError](https://docs.python.org/3/library/exceptions.html#NameError) at runtime.  Was found via #713

### DIFF
--- a/instapy/feed_util.py
+++ b/instapy/feed_util.py
@@ -33,7 +33,7 @@ def get_like_on_feed(browser, amount):
         abort = False
         try:
             like_buttons = browser.find_elements_by_class_name(LIKE_TAG_CLASS)
-        except selenium.common.exceptions.NoSuchElementException:
+        except NoSuchElementException:
             print('Unale to find the like buttons, Aborting')
             abort = True
 


timgrossmann:
@cclauss Thank you for contributing!

However, I saw that the import statement is at the top of the file, could you please double check that?

cclauss:
That import statement defines __NoSuchElementException__ but does not define __selenium.common.exceptions.NoSuchElementException__.

So, there are two solutions to this problem.  The first is to change the import statement to __import selenium__ and leave the except line alone or the second is to make the changes in this PR.

timgrossmann:
@cclauss Maybe I'm missing your point, but it imports **NoSuchElementException** as **NoSuchElementException** from **selenium.common.exceptions** so it should be exactly the same... What am I missing?

cclauss:
Flake8 is now turned on https://travis-ci.org/timgrossmann/InstaPy

cclauss:
```python
from selenium.common.exceptions import NoSuchElementException

try:
    raise NoSuchElementException()
#except selenium.common.exceptions.NoSuchElementException:
#    print('Long form properly handled')
    print('Short form properly handled')
```

1. Try putting this code into a test.py file and run it.
2. You should see: __Short form properly handled__
3. Now, uncomment the two commented lines and run it again.

Via the import, we have defined the short form but we have not defined the long form.

timgrossmann:
@cclauss Ahh, now I understand that!

Wow, that is one hell of a weird case...
Thank you for the fix!

PR #719, opened 2017-10-05
Fix typo in variable name: user_name --> username

cclauss:
Was found via #713
* Also put multi-line if statement in parentheses to avoid using backslashs like the plague that they are (See PEP8). 
* A problem remains that __checked_img__ is an undefined name in this context.  It is defined as a local variable in several other functions/methods but it is not defined until 50 lines later in this one.


timgrossmann:
@cclauss Thank you for contributing!

Issue #425, opened 2017-07-22
Follow by list not working


sionking:
![image](https://user-images.githubusercontent.com/7851877/28494156-1c175aea-6f2f-11e7-93d9-17332709817b.png)

with the follow:
accs = ['therock','natgeo']
session.follow_by_list(accs, times=1)

RealMakAttak:
yes there seems to be a problem here, @timgrossmann has been made aware by me before there is a bug, just waiting for a fix :) Thanks!

hasangok:
It can be coincidence, but after surounding follow_given_user function with try/except, it started working for me.

ato09:
@hasangok how do you do that surrounding thing? I face the same problem when trying to follow by list. Appreciate your guide.

flothousand:
My following seems also to not work. Everything else is fine. Any updates here?

converge:
@sionking some news about this issue ?

converge:
No interaction for a long time. Please, reopen this issue if you need

Issue #402, opened 2017-07-12
Exception thrown when following a given list of people - seems cant find follow button


RealMakAttak:
On the docker compose logs, the bot keeps failing with this error:

selenium_1  | 15:05:19.361 INFO - Done: [find element: By.xpath: //*[contains(text(), 'Follow')]]
selenium_1  | 15:05:26.343 INFO - Executing: [get text: 7 [[ChromeDriver: chrome on LINUX (740e7b55efacbce37f18d67cb07638dd)] -> xpath: //*[contains(text(), 'Follow')]]])
selenium_1  | 15:05:26.893 WARN - Exception thrown
selenium_1  | org.openqa.selenium.StaleElementReferenceException: stale element reference: element is not attached to the page document

Seems like it cant find the follow button?

RealMakAttak:
@timgrossmann sorry Tim but any ideas on this?

RealMakAttak:
@sionking

Issue #336, opened 2017-06-22
selenium.common.exceptions.StaleElementReferenceException: Message: stale element reference: element is not attached to the page document


sweb1:
Using "Follow from list" in the GUI, it follows first one or two users ok, then script stops with this error message:
selenium.common.exceptions.StaleElementReferenceException: Message: stale element reference: element is not attached to the page document
  (Session info: chrome=58.0.3029.110)
  (Driver info: chromedriver=2.29.461591 (62ebf098771772160f391d75e589dc567915b233),platform=Windows NT 10.0.15063 x86_64)

timgrossmann:
@sweb1 This is probably not a problem with InstaPy itself, I'll leave this open, but also make sure to post this problem on the GUI Repo

doremidom:
Getting this issue as well

Mehnou:
I'm getting this as well, after downloading that last update

timgrossmann:
@Nemixalone Could you look into that, please?

doremidom:
I should add, I'm using the command line program on a mac, not the GUI.

timgrossmann:
@doremidom Ahh good to know, thank you for the information! Well I'll look into it as well then...

Mehnou:
Have you found the problem yet ? I'm using the GUI btw

timgrossmann:
Well, I seriously have no idea why this happens (mostly) with the GUI. 
Couldn't get this error with the normal execution of the script and I'm not familiar enough with the GUI to solve his errors.

komad1na:
I can check that out but my GUI is doing nothing else than generating python lines to execute(maybe bad formatting after InstaPy updates), I mean it has nothing with the background working of InstaPy. But sure I can see whats up. @sweb1 @Mehnou Can you please open up **quickstart.py** file and copy whole line with something like this and paste it here please. Thanks

**accs = ['therock','natgeo']
session.follow_by_list(accs, times=1)**

doremidom:
I was able to fix this by removing sleep(10) at line 66, fyi

Mehnou:
I removed it but it didn't work... It actually crashed sooner

doremidom:
I'm not using the GUI, so my fix only works for the command line program

Mehnou:
Well I think it's actually the same, the GUI just writes the code in the quickstart.py and executes it

Mehnou:
[...]

PR #720, opened 2017-10-05
#402 #425 #336 fix selenium exception on follow button

conradlo:
These issues shared the same exception message:
`element is not attached to the page document`

After looking into the function `follow_given_user` and move the line `browser.find_element_by_xpath(...)` below `sleep(10)` the issue gone away.

Also the xpath selection expression was: `"//*[contains(text(), 'Follow')]"`
which will also match "Following", therefore changed to `"//*[text()='Follow']"`


timgrossmann:
@conradlo Thank you, looks good!

PR #743, opened 2017-10-12
NoSuchElementException fix in unfollow_users function and deleted unu…

converge:
NoSuchElementException fix in unfollow_users function and deleted unused variable that was breaking the build at travis.


timgrossmann:
@converge Is this the same as #744 ?

converge:
yes, did PR first

PR #744, opened 2017-10-12
Update unfollow_util.py

alexauvray:
in order to fix xpath issue : 

`Unable to locate element: {"method":"xpath","selector":"/html/body/div[4]/div/div[2]/div/div[2]/div/div[2]"}`


janlink:
Thanks @alexsonnay, used this fix in my instapy instance and it works

timgrossmann:
@alexsonnay We probably have to find a better solution than simply changing the path to the element every now and then.

But thank you very much for this fix and contribution!

PR #769, opened 2017-10-15
flake8 code style instapy.py

converge:
1. flake8 code style
2. deleted unused variables
3. deleted interact_by_users method that is obsolete and its not in the docs anymore (can we delete it?)

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -794,6 +794,165 @@ class InstaPy:
 
         return self
 
+    def interact_by_users(self,
+                          usernames,
+                          amount=10,
+                          random=False,
+                          media=None):
+        """Likes some amounts of images for each usernames"""
+        if self.aborting:
+            return self
+
+        total_liked_img = 0
+        already_liked = 0
+        inap_img = 0
+        commented = 0
+        followed = 0
+
+        usernames = usernames or []
+
+        for index, username in enumerate(usernames):
+            print('Username [{}/{}]'.format(index + 1, len(usernames)))
+            print('--> {}'.format(username.encode('utf-8')))
+            self.logFile.write(
+                'Username [{}/[]]'.format(index + 1, len(usernames)))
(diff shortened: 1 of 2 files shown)


timgrossmann:
@converge I think it would be better to simply add documentation for it since it's the method that does the same thing as `set_user_interact` only that it's one of the "executing" methods and not just a setter for a setting.

Thank you for fixing the code base a little, looks like the CI is going through as well now!

converge:
@timgrossmann your welcome! it's a pleasure work in this project, so much fun! 

I think it's ready to merge.

instapy/instapy.py
```diff
@@ -100,17 +105,20 @@ def __init__(self, username=None, password=None, nogui=False, selenium_local_ses
             self.set_selenium_local_session()
 
     def set_selenium_local_session(self):
-        """Starts local session for a selenium server. Default case scenario."""
+        """Starts local session for a selenium server."""
+        """Default case scenario."""
```
cclauss: This should be a single multiline comment — that is, remove the second and third triple quotes and leave only the first and last triple quotes.  This will make it like the doc strings in the PEP8 doc.  Repeated below.

instapy/instapy.py
```diff
@@ -753,50 +851,78 @@ def interact_by_users(self, usernames, amount=10, random=False, media=None):
...
+                               commenting):
+
                                 if temp_comments:
-                                     # Use clarifai related comments only!
+                                    # use clarifai related comments only!
                                     comments = temp_comments
                                 elif is_video:
-                                    comments = self.comments + self.video_comments
+                                    comments = self.comments + \
+                                        self.video_comments
```
cclauss: Could we use parens to avoid backslashes?  See PEP8 for reasons to avoid backslashes.  Repeated below.

converge:
thanks @cclauss , I will fix it.

PR #775, opened 2017-10-16
Added Incognito mode

AustineA:
Added Incognito mode for anybody that's interested in that. You can pass it as a parameter like so 

```
from instapy import InstaPy

insta_username = ''
insta_password = ''


session = InstaPy(username=insta_username, password=insta_password, incognito_mode=True)
session.login()
```

The reason why you would want to run instapy in incognito mode is to sign in multiple instagram account simultaneously using virtualenv 

Here is how to set that up

Step 1: Install Virtualenv
`pip install virtualenv `

Step 2:  Install a clean Python installation in a virtual environment 
```
mkdir yourfolder
cd yourfolder
virtualenv venv --python=python3.6
```
Step 3: Activate your virtual environment 
`source venv/bin/activate`

Step 4:  Copy instapy into yourfolder cd into instapy and run as you would normally do. Repeat the process from step 2  for each account

### DIFF
--- a/examples/incognitoMode.py
+++ b/examples/incognitoMode.py
@@ -1,0 +1,18 @@
+from instapy import InstaPy
+
+insta_username = ''
+insta_password = ''
+
+
+session = InstaPy(username=insta_username, password=insta_password, incognito_mode=True)
+session.login()
+
+
+session.set_user_interact(amount=5, random=True, percentage=80, media='Photo')
+session.set_do_like(enabled=True, percentage=80)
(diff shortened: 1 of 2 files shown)


dbrkv:
@amtheguy for each run chrome creates new profile every time unless you provide a path to a profile. It work like incognito mode.

AustineA:
@sunthera I didn't know about that. Anyways having it switch to incognito mode give me peace of mind.

Issue #745, opened 2017-10-12
Can't unfollow more than 10 people


alexauvray:
The bot goes to sleep after 10 people like it's supposed to. But then it stops. Any idea ?

```
sleeping for about 10min
--> Total people unfollowed : 10 

Session ended
-------------
```

converge:
@alexsonnay are you using this method ?
session.unfollow_users(amount=10, onlyInstapyFollowed = True, onlyInstapyMethod = 'FIFO' )
amount is the number of profile you will unfollow

alexauvray:
Yes.
I tried with `session.unfollow_users(amount=50)`

converge:
@alexsonnay how is all your login/actions/options ?

alexauvray:
Everything is fine on this matter :). It just ends the session quicker than it's supposed to
```
[2017-10-15 12:21:47] --> Total people unfollowed : 10
[2017-10-15 12:21:47] 
[2017-10-15 12:21:47] Session ended
```
I just have a session.set_dont_include and a session.unfollow_users in the options.

jasetom:
I have the same problems except the script unfollows 19 people and then stops?

converge:
@alexsonnay @jasetom it will be helpful if you can provide detailed information about it, so we can reproduce it.

- what exatcly setup/options/actions are you using and how is it setup ?
- what is the stacktrace/error log

jasetom:
I'm only using the unfollow method. My whole script contains only these lines:
```
from instapy import InstaPy

insta_username = 'username'
insta_password = 'pass'

session = InstaPy(username=insta_username, password=insta_password)
session.login()
session.unfollow_users(amount=1000)
session.end()
```

Terminal prints out this: (I've changed the names of ig accounts to x for privacy)
```
Logged in successfully!
('set_automated_followed_pool error \n', "[Errno 2] No such file or directory: './logs/username_followedPool.csv'")
--> Ongoing Unfollow 1, now unfollowing: x
--> Ongoing Unfollow 2, now unfollowing: x
--> Ongoing Unfollow 3, now unfollowing: x
--> Ongoing Unfollow 4, now unfollowing: x
--> Ongoing Unfollow 5, now unfollowing: x
--> Ongoing Unfollow 6, now unfollowing: x
--> Ongoing Unfollow 7, now unfollowing: x
--> Ongoing Unfollow 8, now unfollowing: x
--> Ongoing Unfollow 9, now unfollowing: x
[...]

PR #796, opened 2017-10-21
bug fix 745, added sleep_delay to unfollow_users method

converge:


timgrossmann:
@converge Looks good from my Side!
It unfollows more than 10 people mentioned in #745 like this

Thank you very much

Issue #779, opened 2017-10-17
follow_user_followers not respecting Amount neither set_user_interact


mamartins:
I have this configuration:
```
bot.set_lower_follower_count(limit = 10)
bot.set_upper_follower_count(limit = 2550)
bot.set_do_comment(enabled=True, percentage=40)
bot.set_user_interact(amount=6, random=False, percentage=100, media='Photo')

bot.follow_user_followers(['instagood'], amount=30, random=True, interact=True, sleep_delay=120)
```

The script was already following the 50th user, so it did not respected the 30 amount and did not entered any user to like its last 6 pictures.

flo1t:
the "problem" is line 270-272 in unfollow_util.py:
```
if is_random:
        # expanding the popultaion for better sampling distribution
        amount = amount * 3
```

when you set random=True in follow_user_followers it'll triple the set amount. so either you divide your wanted amount through 3 or you set random=False.

mamartins:
Hum, I guess that this should be stated under documentation, no?

Also, the bot will only interact with users after all the following is done? Because it did not enter any profile to interact while I was running it

converge:
@flo1t thanks, just submitted a PR. @mamartins

mamartins:
great!

anyone knows the answer to this question, the bot will only interact with users after all the following is done?

timgrossmann:
@mamartins I'm really sorry for the confusion here... 
There are two features, `follow_user_followers` and `interact_user_followers`... They should both be combined, but somehow they turn ended up being two different methods.

Both first go through the followers of the account given. 
The `follow_user...` will follow given amount of accounts right there before going in and interacting with given amount of the followed.
The `interact_user...` however will only get the given amount of accounts and then start to interact with each of the selected ones, liking their pictures, maybe commenting and maybe following the account. 
 
Hope that clears things up a little. The idea was to have a method that in general does nothing, no like, no comment and no follow.
Then you could simply set up all the interaction with `set_do_like`, `set_do_comment` and `set_do_follow` to make it as customizable as possible...

This still has to be finished...

converge:
@mamartins thank you for your bug report, we just merged a fix, can you test it ? if you find some problem, please, reopen it!

mamartins:
I'll test it. 
Thanks guys!

PR #805, opened 2017-10-24
follow_user_followers not respecting amount to follow

converge:
Is this logic needed ? Removing it, we fix the #779 bug. Random (True/False) continue working.

### DIFF
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -270,10 +270,6 @@ def follow_through_dialog(browser,
     sleep(2)
     person_followed = []
 
-    if is_random:
-        # expanding the popultaion for better sampling distribution
-        amount = amount * 3
-
     # find dialog box
     dialog = browser.find_element_by_xpath(
       "//div[text()='Followers' or text()='Following']/following-sibling::div")
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -266,9 +266,12 @@ def follow_through_dialog(browser,
                           is_random,
                           delay,
                           callbacks=[]):
-    followNum = 0
     sleep(2)
     person_followed = []
+    real_amount = amount
+    if is_random and amount >= 3:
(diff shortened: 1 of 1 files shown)


timgrossmann:
@converge Well, this actually is the logic that creates the random selection (when the random parameter is set). 
Else we will only get the N first followers of that account.

Is there no way of changing that to make it usable?

timgrossmann:
@converge We need a way to keep this but simply do a random.sample() where we pick the actual amount of accounts out of the list of x3 times the accounts

converge:
@timgrossmann pls correct if I'm wrong, that line doesn't help in profiles sorting and doesn't interfere in random.sample here
https://github.com/timgrossmann/InstaPy/blob/8662c9247df4c2a0058a059fef3dad433e03584a/instapy/unfollow_util.py#L321

timgrossmann:
@converge well, I think it does, maybe I get something wrong here...

The `random=True` parameter is supposed to load more accounts until there are 3 times the number of accounts aggregated. 
The next step is to get a randomly distributed sample of the actual given amount of accounts out of the whole list of 3xamount accounts.

What the referenced code snippet does is that it gets random indices which will then be used to create a list of follow buttons that includes the follow buttons of selected indices:

https://github.com/timgrossmann/InstaPy/blob/8662c9247df4c2a0058a059fef3dad433e03584a/instapy/unfollow_util.py#L320-L324

This should lead to having n random follows over n*3 aggregated accounts. (n = amount here)

I'm really sorry if I am misunderstanding your point @converge

converge:
hey @timgrossmann , I'm not 100% sure if it's a good idea, can you check ?

timgrossmann:
@converge Looks promising!

I'll test it right away.

timgrossmann:
@converge Checked it with Linux and MacOS.

Only follows the given amount of people.

Feel free to wait for a little bit more feedback before closing the issues. 
But it definitely works.

Issue #788, opened 2017-10-19
Store User IDs and create blacklist


JCADC:
> [![Bountysource](https://api.bountysource.com/badge/issue?issue_id=50499135)](https://www.bountysource.com/issues/50499135-store-user-ids-and-create-blacklist?utm_source=50499135&utm_medium=shield&utm_campaign=ISSUE_BADGE)

Hi all, 

Is there anyway to store on a data base the users that you have interacted with and then create a blacklist so you won't impact them again? 

Thanks!

JCADC:
Hi there,

I've just posted a Bounty: https://www.bountysource.com/issues/50499135-store-user-ids-and-create-blacklist

Best regards

converge:
@JCADC can you test it ? #812

timgrossmann:
@JCADC Could you specify this feature a little bit closer?

I'm not sure to understand the purpose of it.

Do you really want to store every user you liked a picture from to not be interacted again?

@converge Thank you very much!

JCADC:
Thanks @converge, I will try it in the late evening. 

Sure, @timgrossmann. Let me ellaborate:
- We create a function to comment the pictures with the hashtag "cat" and set the percentage = 50. We are commenting only to the 50% of the people there. In the future the number of posts should increase and we might want to target them again. It may be the case that one of the new pictures is from a user that we have previously commented/liked, therefore the bot could be seen as a bot/spam (assuming that we write the same or similar comment).

My point is to store in a data base all the user IDs that you have interacted with in order to blacklist them in the future campaigns and create a funnel to measure the effectiveness of the action. Example:

First comments to "cat". Let's imagine we only have 8 posts and we set percentage=50. The database would look as follows:
User ID - Hashtag - Comment - Interaction (like back, reply/like the comment, start following)
1 - Cat - Nice Cat - None
2 - Cat - Nice Cat - None
3 - Cat - Nice Cat - None
4 - Cat - Nice Cat - Replied
We have other 4 users left.
Effectiveness: 25%

Two weeks later we decide to send a new round and there are 11 posts, therefore 3 new ones: One of them is from a new user (user 9), other is from user 1 (who have never interacted with us) and the last one is from User 4 who replied our comment.

Here comes in the blacklist function: We set percentage = 100 and enable the blacklist. The blacklist **ideally** could be set for everyone, the ones who never interacted back (or any other condition in a more advanced phase). This time we set it for everyone that received a comment and didn't interact.
The new comment round should impact to:
[...]

PR #812, opened 2017-10-25
Blacklist feature

converge:
If enabled=True, after interact (liked or followed) the user will be stored in the blacklist. Users in the blacklist will not be liked or followed.


phil1995:
This does not work for the command follow_user_followers.

converge:
we have more specifications here #788 , work in progress

converge:
I made a big mess trying to rebase it, I'm closing it, there is a new PR.

PR #824, opened 2017-10-27
locate button xpath only if text is exactly the follow word

converge:
//*[contains(text(), 'Follow')] matchs the "Following" word too, that means that if even we were already Following this profile, it would try to Follow it again, resulting in a unfollow action.

### DIFF
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -194,7 +194,7 @@ def follow_user(browser, follow_restrict, login, user_name):
 
     try:
         follow_button = browser.find_element_by_xpath(
-            "//*[contains(text(), 'Follow')]")
+                "//button[text()='Follow']")
 
         # Do we still need this sleep?
         sleep(2)


converge:
how I tested it:
```
session.set_do_follow(enabled=True, percentage=100, times=1)
session.set_do_like(enabled=True, percentage=100)
session.like_by_tags(['ilovecarrotjuice'], amount=10)
```

**before this patch**, to reproduce the problem:

watch browser window while InstaPy is working, when it opens a post page to like, click in Follow button and watch InstaPy clicks in "Following" button:

**after the patch**, InstaPy wont click at Following button and display the message:
--> Already following

sionking:
Just what I said in my post.
Thanks for commiting.

converge:
@sionking thank you very much! With your information was easier to debug and fix the problem. Can you help testing it ?

timgrossmann:
@converge Excellent! Really minimalistic fix that looks like it can resist a few changes on the general layout of the page!

Just tested it and it looks like it is definitely working on MacOS and Ubuntu

Your Awesome!

timgrossmann:
@converge Let's wait for some feedback and then close the issues if they got fixed 😉

staringispolite:
Testing on Ubuntu as well in headless mode, I'll update later this week.

What would be the best workaround if something like this happens in the future? Catch/ignore exceptions in our own code, or submit a PR to make this a non-fatal error?

converge:
hello @staringispolite , I think it ony happens on layout changes. It´s good to have fatal errors, in this way can always see whats wrong and fix the problem.

staringispolite:
This was a layout change? Looked like it was just failing when it tried to
follow someone you’re already following because it was looking for exactly
“Follow” but the button text was "Following". (I don’t think they changed the word “Follow”)

I agree with the principle of your comment, but I do think errors can be
visible and actionable without crashing the entire script. For instance,
how like_by_tags will say “Load button not found, working with current
images!” Instead of throwing an exception.

Curious what Tim’s philosophy is on this. (And if exceptions are the
preferred way to communicate errors, may be worth adding try statements to
the docs - which I’d be happy to do a PR for btw)

converge:
@staringispolite sure, my bad, was thinking about other things. It would be better to only show errors and dont crash.

staringispolite:
Hm, I still get crashes on this issue. I think it's because the patch only catches NoSuchElementException, but I get ElementNotVisibleException

[...]
  File "/root/instagram/iamaslowtrain/instapy/instapy.py", line 530, in like_by_tags
    followed += follow_user(self.browser, self.follow_restrict, self.username, user_name)
  File "/root/instagram/iamaslowtrain/instapy/unfollow_util.py", line 175, in follow_user
    follow_button.click()
  File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/remote/webelement.py", line 72, in click
    self._execute(Command.CLICK_ELEMENT)
  File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/remote/webelement.py", line 461, in _execute
    return self._parent.execute(command, params)
  File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/remote/webdriver.py", line 236, in execute
  File "/usr/local/lib/python2.7/dist-packages/selenium/webdriver/remote/errorhandler.py", line 192, in check_response
selenium.common.exceptions.ElementNotVisibleException: Message: element not visible
  (Session info: chrome=61.0.3163.100)
  (Driver info: chromedriver=2.33.506092 (733a02544d189eeb751fe0d7ddca79a0ee28cce4),platform=Linux 4.4.0-98-generic x86_64)`

converge:
@staringispolite that issue happens often, not specially related to this issue. It would be great if you could fix it.

https://github.com/timgrossmann/InstaPy/search?q=ElementNotVisibleException&type=Issues&utf8=%E2%9C%93

staringispolite:
Oh, interesting. Is there another issue tracking ElementNotVisibleException? Do you know anything else that could get me started?

converge:
@staringispolite unfortunately I dont have much information about this issue, never had the time to see what's going on. What I do know is that it's hard to catch, cause it's not happening often.

staringispolite:
It happens super often for me, maybe once every 100 follows. My first assumption was Chrome running slowly (or paging memory to disk, etc) because it's running on my low-memory digital ocean server. 

I don't know how to watch what it's doing, since I think D.O. is CLI (and thus headless) only... Any pro tips for debugging?

converge:
@staringispolite the best thing to start is having a local box to test it. I would start googling it, It's a frequent issue related to Selenium.

about debugging, it will be hard to debug as it dont happens often, but python -m pdb is a begginin.

Are you at slack ? There is some people there and I can help you too. (slack invite link at readme)

Issue #839, opened 2017-10-30
[feature] InstaPy statistics


converge:
Hello,

I'm thinking about this idea and wanna share with you.

Record every InstaPy activity (like, comment, follow, unfollow, account followers, account following) in a statisticas.csv file and organize it by day, like:

```
date,likes,comments,follows,unfollows,instagram_server_call
01.xx.xx,30,15,120,150,1000
02.xx.xx,50,25,110,150,1200
03.xx.xx,10,5,110,100,700
```

**Why ?**
To be able to create reports, measure interactions and measure account grow.

This data can be used by external tools to create reports, in this way, we focus InstaPy for the core/engine and let it be extensible as it's today with InstaPy-GUI.

Chart example that can be created using the statistics.csv data:
![Imgur](https://i.imgur.com/npPBQwg.png)

RomeoJuliett:
I think this is a pretty good idea actually! Custom analytics would go a long way in being able to track our activity. I currently use SocialBlade to monitor how far my reach is extending. Not ideal, but it does work.

JCADC:
That'd be awesome @converge !

pilotdude:
I've been brainstorming how to collect and report analytics for a while now. I definitely want to devote some time to it. I've been debating over whether it would be better to use csv or establish a local SQLite Database. Do you have a reason for suggesting a csv file as opposed to SQLite?

converge:
hey @pilotdude ! A "real" database requires heavy modules (and more modules to be installed), a database diagram well documented and it increases complexibility.

I think the csv approach is good to keep it clean and simple, in this way the database complexibility can be used for front ends for example where way more informations are needed/collected/saved.

I'll propose a pull request tonight, it's 90% done, just need to track more server calls and more testing.

pilotdude:
@converge I just skimmed your OP and thought it was exactly what I was thinking about but it wasn't. For what you proposed a csv file is definitely much better. 

I was thinking of a whole lot more analytics and info that could be used to speed up the process. For example noting the follower counts of specific accounts, noting post IDs that had already been interacted with etc. So then when the script comes across a post it first searches the DB to see if we already have seen that this post is for some reason not appropriate before it has to load a page and check. That I guess is my dream of where I would like to see this end up :)

I'm pretty sure using SQLite wouldn't require any extra installation because it already comes with python.  But for sure we should be using a csv file for what you are proposing.

converge:
[...]

PR #853, opened 2017-11-02
Record activity, generate statistics

converge:
I tried to make as simple as possible:

every server call is recorded at logs/activity.csv file, page load, page content, likes, comments, follows and unfollows are recorded by day in this file.

the file layout is:
```
date,likes,comments,follows,unfollows,server_calls
11/01/17,18,14,52,0,84
11/02/17,14,10,32,0,156
```

more informations about this feature: #839 

if you guys wanna keep activity.csv open to analyse it, this frankstein can help:
```python
# osx
from subprocess import call
from time import sleep

while(True):
  with open('activity.csv') as f:
    for line in f:
      print line
      sleep(0.2)
      if 'str' in line:
        break
    else:
      call(['clear'])

### DIFF
--- a/instapy/comment_util.py
+++ b/instapy/comment_util.py
@@ -2,6 +2,7 @@
 """Module which handles the commenting features"""
 from random import choice
 from .time_util import sleep
+from .util import update_activity
 from selenium.common.exceptions import WebDriverException
 import emoji
 
@@ -48,6 +49,7 @@ def comment_image(browser, comments):
         comment_input[0].send_keys("\b")
         comment_input = get_comment_input(browser)
         comment_input[0].submit()
+        update_activity('comments')
     else:
         print('--> Warning: Comment Action Likely Failed:'
               ' Comment Element not found')
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -14,6 +14,8 @@ def get_links_from_feed(browser, amount, num_of_search):
     """Fetches random number of links from feed and returns a list of links"""
 
     browser.get('https://www.instagram.com')
+    # update server calls
+    update_activity()
     sleep(2)
 
     for i in range(num_of_search + 1):
@@ -60,8 +62,9 @@ def get_links_for_location(browser,
         # Make it an array to use it in the following part
         media = [media]
 
-    browser.get('https://www.instagram.com/explore/locations/'
-                + location)
+    browser.get('https://www.instagram.com/explore/locations/' + location)
+    # update server calls
+    update_activity()
     sleep(2)
 
     # clicking load more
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by timgrossmann** [comment]
@converge Awesome, thank you for that PR!

I'll try to look at it later today.

**Comment by timgrossmann** [comment]
@converge Seems to be working like charm!

Do you want to add this before adding #835 ?

**Review by timgrossmann — APPROVED** [review]

**Comment by converge** [comment]
@timgrossmann Just merged this. I'll extend the blacklist feature doc before our merge.

**Comment by Radzam** [comment]
I have updated to latest InstaPy but it does not record activities when Im using docker microservice. Can you please instruct me how to do it?

**Comment by converge** [comment]
hello @Radzam , checks your logs/activity.csv file. maybe we need add it this to readme.

**Comment by Radzam** [comment]
@converge I checked already but it doesnt record anything. Im using docker microservice setup and not on my local machine.

**Comment by timgrossmann** [comment]
@Radzam I think keeping state in a Docker container that always starts over, when it's time to start a new InstaPy session, is pretty hard to do, isn't it?

I only worked with "real" stateless micro services and Docker so I can't tell a lot about this case here...
You probably somehow have to "extract" the recorded statistics out of the container into the host system

(Docker guys correct me if I'm wrong, please)

**Comment by Radzam** [comment]
@timgrossmann I see. It is however outputting the logFile.txt errors and info so perhaps use the same approach for activity.csv?

**Comment by Naramsim** [comment]
Hi @converge, the statistic named "likes" count real likes or even the attempted ones?

**Comment by converge** [comment]
@Naramsim it counts only liked posts, the attempts counts as server_call

**Comment by Naramsim** [comment]
Would it be useful to split the data regarding script runs?
I use Schedule so my script never really stops, and Instapy runs multiple time per day, with different settings. Right now with this PR I have an entire day data all aggregated.

I would prefer something like:
```
date,run,likes,comments,follows,unfollows,server_calls
11/01/17,1st,18,14,52,0,84
11/01/17,2nd,14,10,32,0,156
11/02/17,1st,19,11,31,0,154
```

@converge @timgrossmann @Radzam

converge:
@Naramsim this makes me think we could have a general campaign_name setting, and aggregate it by campaign, what do you guys think ? (blacklist feature uses campaign_name to organize it's csv file).

gitpatrickhub:
I like it! It would allow for more detailed analysis.

riccardorighettiphoto:
I don't know if I missed something, but I cannot find any .csv file in the logs folder about recorded activities.
I am also looking for a way to analyze my runs based on information stored in the general log file.
Did someone develop (even in any fork) anything for storage statistics RAW data?
Thank you.

Issue #797, opened 2017-10-22
Feature: (implemented) MAX LIKE/FOLLOW per HOUR/DAY (works for multiple running also).


sionking:
If it is an interest to someone I will add the code here.

* I used pickle.
* Able to read it at any time since it saved as file.
* wait for next hour/day to return to work if max is reached.

Sion.

VKT16:
@sionking I was wondering about how to set the bot to follow 400 accounts and like 400 posts in 10 hours (I am only aiming to use the bot 10 hours a day to get around Instagram algorithms). 

I think that your code might be of some help, so can you please post it, and a bit of guidance on what each line does as I am a beginner?

Thanks

sionking:
Sure ASAP

VKT16:
@sionking  Patiently waiting as I need it really badly :laughing:

brookton:
I'm interested! 🙂

sionking:
It is not bug free (currently I ignore any errors) since I have no much time for it, But it do works.
I started it since the automation crashed all the time so I was unable to track to total daily likes/follows.
Now it will read the file before every run so it wont start liking from the begginging.
- If the Max is reached I print to console.
- If the Max is reached I wait certain amount of time (instead of doing diff to next hour) since statisticly it will be the same.
- every follow/like it will update the file (not best performance but reliablty)

I am attaching the instapy.py file, But I will explain everything here.
Python is a hobby lang since I use c at work in real time codding, so please fix bugs.

Code(instapy.py):
Added new class:
```

class InstaPyStorage(object):
    TOTAL_START_DAY = '16/10/2017'
    MAX_PER_DAY = 2000
    MAX_PER_HOUR = 90
    def __init__(self):
        self.total = 0
        self.thisDayTotal = 0
        self.thisHourTotal = 0
        self.lastDayExecuted = _datetime.date.today()
        self.lastHourExecuted = _datetime.datetime.now().hour
    def updateStatistics(self):
        # read today's date in 2008-11-22 format, and now time
        today = _datetime.date.today()
        now = _datetime.datetime.now()
        self.total += 1
        if self.lastDayExecuted == today:
            self.thisDayTotal += 1
            if self.lastHourExecuted == now.hour:
                self.thisHourTotal += 1
            else:
                self.lastHourExecuted = now.hour
                self.thisHourTotal = 1
        else:
            self.lastDayExecuted = today
            self.lastHourExecuted == now.hour
            self.thisHourTotal = 1
[...]

PR #859, opened 2017-11-04
Limit daily interactions

converge:
Hey guys, this is a work in progress, it's kind complex control the InstaPy interactions. I will appreciate if anyone could take a closer look on it.

I'll do more tests, and follow/unfollows limit control aren't 100% ok yet.

(the first 2 commits are from my rebase, the PR are the last 2 commits)

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,6 +333,16 @@ session.like_by_tags(['natgeo', 'world'], amount=10)
 session.like_by_feed(amount=100, randomize=True, unfollow=True, interact=True)
 ```
 
+### Limit Daily Interactions
+
+```python
+# You can limit InstaPy actions by day. 
+# likes = 100 means it will stop InstaPy when 100 posts are liked 
+# When some limitation (likes, comments, follows, unfollows or server calls) is 
+# reached, it will stop InstaPy for this day
+
+session.set_interaction_limits(likes=150, comments=25, follows=60, unfollows=30, server_calls=1000)
+
 ### Restricting Likes
 
 ```python
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -25,6 +25,7 @@ from .print_log_writer import log_follower_num
 from .time_util import sleep
 from .time_util import set_sleep_percentage
 from .util import get_active_users
+from .util import check_activity
 from .unfollow_util import get_given_user_followers
 from .unfollow_util import get_given_user_following
 from .unfollow_util import unfollow
@@ -1396,6 +1397,37 @@ class InstaPy:
         for user in active_users:
             # include active user to not unfollow list
             self.dont_include.append(user)
+
+    def set_daily_interactions(self,
+                               likes=None,
+                               comments=None,
+                               follows=None,
+                               unfollows=None,
+                               server_calls=None):
+
+        if (likes or comments or follows or unfollows or server_calls) is None:
+            print('Warning: set_daily_interactions is misconfigured')
+            self.logFile.write(
+                'Warning: set_daily_interactions is misconfigured\n')
+            return
+
+        current_activity = check_activity()
+
+        if current_activity['likes'] >= likes:
+            print('Daily likes limit reached, closing...')
+            self.end()
+        elif current_activity['comments'] >= comments:
+            print('Daily comments limit reached, closing...')
+            self.end()
+        elif current_activity['follows'] >= follows:
+            print('Daily follows limit reached, closing...')
+            self.end()
+        elif current_activity['unfollows'] >= unfollows:
+            print('Daily unfollows limit reached, closing...')
+            self.end()
+        elif current_activity['server_calls'] >= server_calls:
+            print('Daily server calls limit reached, closing...')
+            self.end()
 
     def end(self):
         """Closes the current session"""
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -25,7 +25,7 @@ from .print_log_writer import log_follower_num
 from .time_util import sleep
 from .time_util import set_sleep_percentage
 from .util import get_active_users
-from .util import check_activity
+from .util import check_activity_limits
 from .unfollow_util import get_given_user_followers
 from .unfollow_util import get_given_user_following
 from .unfollow_util import unfollow
@@ -100,6 +100,12 @@ class InstaPy:
 
         self.like_by_followers_upper_limit = 0
         self.like_by_followers_lower_limit = 0
+
+        self.likes_limit = None
+        self.comments_limit = None
+        self.follows_limit = None
+        self.unfollows_limit = None
+        self.server_calls_limit = None
 
         self.aborting = False
 
@@ -179,6 +185,7 @@ class InstaPy:
 
     def login(self):
         """Used to login the user either with the username and password"""
+
         if not login_user(self.browser,
                           self.username,
                           self.password,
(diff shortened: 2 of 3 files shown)


snakelab:
we talked about this "interactions" recently and found a good way would be to integrate a state-machine which reacts on "like limit reached" and so on. I have evaluated a few python state-machines and found a good one to integrate: https://github.com/jtushman/state_machine
Maybe?

converge:
hey @snakelab, that's nice! Do you have some code ?

snakelab:
just a (simple) poc about the states itself, not an in instapy integrated one. The cool part is that you add it via annotation "@acts_as_state_machine":

```python
 #!/usr/bin/python

#-- https://github.com/jtushman/state_machine
from state_machine import *

@acts_as_state_machine
class Liking():
    name = 'Liking'

    limit_init = State(initial=True)
    limit_reached = State()

    start = Event(from_states=limit_reached, to_state=limit_init)
    stop = Event(from_states=limit_init, to_state=limit_reached)

    @before('stop')
    def do_one_thing(self):
        print "BEFORE {} stop".format(self.name)

    @after('stop')
    def snore(self):
        print "AFTER stop"

    @after('start')
    def big_snore(self):
        print "AFTER start"

likes = Liking()
print likes.current_state == Liking.limit_init
print likes.is_limit_init
print likes.is_limit_reached
likes.stop()
print likes.is_limit_init
likes.start()
print likes.is_limit_init
```

Next would be to try this in instapy-code...

converge:
@snakelab To be honest I still prefer the simple things. I have to say I'm a bit old school, One of those guys that still likes assembly language. I prefer to save things in a csv file and load/filter it using a dictionary than adding complexibility.

I see InstaPy as a engine, and we can extend it like InstaPy GUI and the dashboard idea.

It's just a personal point of view, probably most people disagree with me and they're probably right. Your approach is modern, and sounds pretty cool.

snakelab:
In fact state-machines turning complicated things in simple states which are available all the time. Ones used you will find it easy to work with. I my opinion Its the answer to the liking, commenting, etc. thing, because you want to react on different states in different situations.
csv => sqlite
follow/unfollow/etc. => state-machine

Your approach is the pragmatic one ;-)

sionking:
already did one: works great for me with multiple running simultaniasly:
https://github.com/timgrossmann/InstaPy/issues/797
I used pickle, enjoy.

ugurozturk:
@converge what about delaying each like / follow ? 
import time

next_like = time.now + (24 * 60 * 60 / daily_limit_like)
next_follow = time.now + (24 * 60 * 60 / daily_limit_follow)

while (true){
 if(next_like < time.now) 
    doLike()
 if(next_follow < time.now)
    doFollow()
}

converge:
@ugurozturk sounds good for run it once a day, but I think most people run it several times a day, don't you think ? In this way we have to worry about "like speed", we can have a situation where the like function is liking too fast.

My approach limit it by day whatever happens (like a crash). But you're right about the speed, we need to improve it, sometimes is too slow. Thank you for your idea, let's keep it touch. If you need something, I'm at our Slack.

ugurozturk:
@converge Well i am a person who does run it once every week :) Even a person does run multiple times for a day delaying also is great for him. Because if a person does want 300likes for a day, whenever he does stop / run the script he will get 300likes for a day. Because of the delay.

converge:
@ugurozturk this is what this PR proposes, take a look
```python
+session.set_interaction_limits(likes=150, comments=25, follows=60, unfollows=30, server_calls=1000)
```

How would you like to change it ?

If the idea is change the speed, we could focus on sleep_delay argument

ugurozturk:
Might people wants it delayed or non-delayed so
`+session.set_interaction_limits(delay=true,likes=150, comments=25, follows=60, unfollows=30, server_calls=1000)`

Couldn't merge your PR to my fork. Auto-merge failed and couldn't make it work. So i writing here an example

```python
def follow_by_list(self, followlist, times=1):
        """Allows to follow by any scrapped list"""
        # checks set_interaction_limits before start
         if (check_activity_limits(
                 self.likes_limit,
                 self.comments_limit,
                 self.follows_limit,
                 self.unfollows_limit,
                 self.server_calls_limit
                 )):
                     # force exit
                     self.end(True)
        if(self.delay):
             sleep(24 * 60 * 60 / self.follows_limit)

        self.follow_times = times or 0
        if self.aborting:
            return self

        followed = 0

        for acc_to_follow in followlist:
            if acc_to_follow in self.dont_include:
                continue

            if self.follow_restrict.get(acc_to_follow, 0) < self.follow_times:
                followed += follow_given_user(self.browser,
                                              acc_to_follow,
                                              self.follow_restrict,
                                              self.blacklist)
                self.followed += followed
                self.logFile.write('Followed: {}\n'.format(str(followed)))
                followed = 0
            else:
                print('---> {} has already been followed more than {} times'
                      .format(acc_to_follow, str(self.follow_times)))
                sleep(1)

        return self
```

check the delay

Edit : Sorry, it will not work. Because the program would do (only) likes, or follow for a day.. I have another idea but i dont know how to make it on code. Like giving date time for next like / follow

converge:
@ugurozturk about the delay=True, I think it can be confusing, because we're defining a delay, but doesn't specify where, is it for like, comment, follow or unfollow ? or for all those actions ? How 
```python
session.follow_user_followers(['friend1', 'friend2', 'friend3'], amount=10, random=False, sleep_delay=60)
```
should behave after this:
```python
session.set_interaction_limits(delay=true,likes=150, comments=25, follows=60, unfollows=30, server_calls=1000)
```
definition ? Do you understand my point about the delay ?

I think it would be better to add delay option for more actions, for example ```like_by_tags``` doesn't have the sleep_delay option, we could add this option to this action and it would be (example):
```python
session.like_by_tags(['natgeo', 'world'], amount=10, sleep_delay=30)
```

converge:
if we aprove #913 , I'll rewrite it to use sqlite.

sionking:
Will it work with 2 or more process running simultaneously ?
e.g. I like to run interact by user and like by tags and unfollow all toghter.

sionking:
I don't think aborting is the right aproch, Won't it be better to just delaying the task.
Aborting will force you to rerun the process, why not just wait for the limit to exprie (wait for next day for example).

converge:
hi @sionking , today InstaPy is synchronous/linear/single thread, it isn't possible to run 2 process simultaneously.

Aborting when limit is reached, is the natural behavior, it releases python/InstaPy resources, if not in this way, the computer/server will hold unused resources like memory and cpu. The natural behavior is, the next day, the scheduler(cron for example) will run it again.

What I do think it's a good idea, is to warn the user about the stop. It will be good through Telegram/email/sms/etc...

What do you think ?

ugurozturk:
@converge my idea had wrong algorithm anyway. And yes your option is better. 
`session.like_by_tags(['natgeo', 'world'], amount=10, sleep_delay=30)`

sionking:
Reading user followers, reading tag post links... thay all be erased and need to reread next time you start the process. running in idle "sleep" not consuming to much cpu.

I am running more than one process (3 actually) for more than 4 months, it's works great (sure there should be a mutex on some file)!

converge:
@sionking how do you run 3 process simultaneously ? Are you using threads ?

sionking:
@converge 
No, I am using windows process  :)
Threads can be a great idea since we can have a shared resources and Mutex.
I really don't think it is a good idea to do the same actions all the time i.e. likes from same tag for 2 hours, unfollow 200 users for 2 hours, like my feed for 1 hour.... Instagram will notice.

Better to mix them all together up to a daily maximum (I reaches 2000 likes a day).

BTW important**
in multi-threaded or process approach will not work since you set the max per process
+session.set_interaction_limits(likes=150, comments=25, follows=60, unfollows=30, server_calls=1000)
I use a local file to read the values, and pickle file to store the current state.

CharlesCCC:
@sionking  I kind did the same thing as you do, but a littel bit differently. I created to schedule task one for following/like another one for unfollow. But I didn't schedule it to run on the same time tho, I separately it about a hour apart. (I don't know you can actually have both run together) Very impressive.

converge:
I'm closing it since the csv process approach doesn't work on Windows and we couldn't extend it in the future.

Issue #724, opened 2017-10-06
'Load button not found' after script doesn't scroll on a tag


drogon34:
The script is not scrolling for me when it opens a tag to interact. Also, I get the 'Load button not found' message. I checked the class for load button in like_util.py and it has not changed. I had pulled and have the latest changes since yesterday.

converge:
does it happen in like_by_tags() function ? what parameters are you using ?

drogon34:
My quickstart.py file:
`from instapy import InstaPy
import os
session = InstaPy(username='xyz', password='xyz')
session.login()
session.set_do_comment(enabled=True, percentage=25)
session.set_comments(['Beautiful!', 'Great' ])
session.like_by_tags(['beach', 'sunset'], amount=40, media='Photo')

session.end()`

Script was working fine till couple of days back, started seeing this issue since yesterday.

drogon34:
It started working again now. Seems like it was a Chrome issue as I run from my local. Closing this issue.

vosh:
I'm still having this issue with GUI

drogon34:
Here is what I observed. Chrome window scrolls and the script works normally on one of my lesser used accounts. However, for the account I frequently use and run the script on, the window has stopped scrolling. If I manually keep scrolling, it picks up the required number of posts and starts interacting with them. Not sure why this is happening. Again, I have not seen this behavior before and it started since few days back.
I ran the script from both command line and GUI and same issue. Also, I don't think system setting have anything to do with it as the script works normally on my secondary account.

filipstrapinadotcz:
Its not working for me neither. But, what helped, when "hashtag" page is loaded, just scroll manually down and load as much images as you can/want and its working. Its not a solution for autoload issue, it is just letting you use the app

converge:
Im trying to reproduce the problem, it's working fine here without errors using 

```session.like_by_tags(['roadbike'], amount=5)```

can someone help to reproduce the error ?

drogon34:
I have a theory. Seems like it's an account specific issue. I don't know if it's possible or not, but looks like Instagram has disabled `browser.execute_script("window.scrollTo(0, document.body.scrollHeight);")` piece of code for my account when I login from the browser. Again, I'm not sure how they can control the DOM for a user account, if at all it can be done. But that seems like the only explanation to me for now.

I am currently just manually scrolling and using the project. Not ideal, but a quick fix nonetheless.

converge:
[...]

PR #862, opened 2017-11-04
extending PR_860 sunnypatel165 solution and small change to not break…

converge:
hey @sunnypatel165 , great work! I just had to do a little change because update_activity wasn't being called, and extended your solution to other possibilities trying to find the Load Button.

### DIFF
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -73,15 +73,17 @@ def get_links_for_location(browser,
 
     abort = True
     try:
-        load_button = body_elem.find_element_by_xpath(
-            '//a[contains(@class, "_1cr2e _epyes")]')
+        # scroll down to load posts
+        for i in range(int(ceil(amount/12))):
+            browser.execute_script(
+                "window.scrollTo(0, document.body.scrollHeight);")
+            sleep(2)
     except:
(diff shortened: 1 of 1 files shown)


converge:
Hey, I don't know if we should keep this PR or not. Do we still need the load button ? The changes I made ignore the load button to load content.

timgrossmann:
@converge Can we assume that every load button has been removed?

converge:
@timgrossmann I think so, if someone complain about it, we claim a bounty. hehe kidding, as there is no information about the existence of the load button, we can merge it and see what happens.

converge:
someone reported that his account uses the button.

timgrossmann:
@converge But that's not on the issue, right?
Did he write you a message?

converge:
@timgrossmann they reported it at #724 mpaolor and drogon34.

sounds like we need to treat more exceptions

drogon34:
@converge @timgrossmann I just checked on my new account and the Load button is no longer present. Seeing infinite scrolling as the consistent UI across both my accounts now.

converge:
just proposed another PR fixing this in a better way #888

PR #875, opened 2017-11-06
Removes Record activity functionality for windows users

converge:
There is a problem on windows when we try to delete a temporary file. To fix this, the solution is ugly, needs extra win32 libs. I think it's a good idea for now to not provide this functionality to windows users.

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -38,6 +38,7 @@ from .unfollow_util import dump_follow_restriction
 from .unfollow_util import set_automated_followed_pool
 import random
 import csv
+import os
 
 
 class InstaPy:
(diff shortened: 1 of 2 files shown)


timgrossmann:
@converge I unfortunately don't have a windows machine staying around... 

However, I think this is an ok fix for the moment. Even though this might lead to some more confusion than an actual error message. Silently doing nothing is pretty bad.

We might however simply add a log message that tells the user that the feature currently is not supported for windows. Therefore they would be informed and that hopefully should prevent a lot of issue creations.

What do you think? 
I don't think I'd approve merging this with only a return

Thank you for all the hours you put into this!

converge:
@timgrossmann I do understand your view. I found it a quick solution to keep the feature live for those who is already using it, and avoiding crash for windows users. It´s not a functionality that the user can enabled/disable I don´t see much problem in this approach.

It will add the informin message, so users can understand whats happening. Your welcome, it´s a pleasure "work"(it´s fun) in InstaPy.

Issue #878, opened 2017-11-06
Unable to get username on like_by_feed - empty username


jackpot001:
## Expected Behavior
The like_by_feed feature with Interact=True should be able to navigate through the user's profile and like a given amount of their pictures, then go back to the feed and keep liking.
## Current Behavior
I've been having an issue where the bot was getting an empty ('') username so it wasn't able to retrieve images from the user:
```
--> Image Liked!
--> User followed: ['']
Username [1/1]
--> b''
Getting   image list...
Interaction begin...
```
## Working Solution
I was able to debug the code and see the problem was in the "find_element_by_xpath()" function that retrieves the profile name in the feed. Looking at the HTML code of the page it seems the "a" tag with the title attribute is one level deeper, therefore you need to add another "div" to the path, and change the code as follows in line 1295, instapy.py:
FROM
`username = (self.browser.
                                 find_element_by_xpath(
                                                              "//main//div//div//article"
                                                              "//header//div//a"))`
TO
`username = (self.browser.find_element_by_xpath(
                                                    "//main//div//div//article"
                                                    "//header//div//div//a"))`

Hope this helps other people as well.

converge:
hey @jackpot001 , thanks for your help!

I just made a PR with your informations, can you test it ? #881 :octocat:

jackpot001:
Thanks!
I'll test it as soon as I can and leave a report here

jackpot001:
PR #881 Works like a charm!

I'm wondering if the problem was maybe caused due to a change in the HTML code by Instagram. Is it possibile that we might need to change these "find_element_by_xpath" every once in a while because Instagram is changing the source code?

Anyway, thanks as well for the scrolling on the user profile to bypass the "Load button not found" error, much appreciated.

converge:
@jackpot001 great to hear that!

I don't know if we need to worry about xpath elements, it's quick to fix it. Another approach would be find elements by it's names, but in the end, it could be the same work to fix it (locate xpath new button text content, maybe the button disappear and we need to locate it anyway)

PR #881, opened 2017-11-07
#878 bug fix (like_by_feed)

converge:
changes xpaths to locate elements, removes unnecessary sleeps, fixes load button problem for like_by_feed


converge:
hey @timgrossmann , I and @jackpot001 tested it, the fix is working well ( #878 ).
As like_by_feed is broke at the moment, I'm quickly merging it.

PR #887, opened 2017-11-07
Fix random crash issue for some users

CharlesCCC:
Credit goes to the couple guys from the issue discussion.


converge:
really ?

converge (commented):
I know the record activity feature is breaking for Windows, but we need common sense to think collective. Our friend @CharlesCCC clearly don't know what he is doing and it's ok, let's help him.

For now, we should aprove #875 and wait a better approach to have the functionality.

Issue #914, opened 2017-11-13
List Index Out of Range Error when trying to Unfollow


arjunmurali1993:
## Current Behavior
Unfollow_util was always failing when it was trying to access the 'Following' button on our Instagram profile page and was throwing the error "following_link error, list index out of range"

## Possible Solution (optional)

Changing line 131 of Unfollow_util.py to the following got it working for me:
`'//article//ul//li[3]')`

converge:
thanks @arjunmurali1993 , can anyone else test it ? It's good have more people testing it, because sometimes the layout change isn't global(countries/regions).

paranoiasystem:
it works already for me

fabriciogava:
I'm getting the "list index out of range" on unfollow, but not sure if its related to this thread as I'm using InstaPy in a server with the nogui option.
I'm having this problem for a few days already.

I just get:

> Logged in successfully!
> set_automated_followed_pool error
>  list index out of range
> --> Total people unfollowed : 0
> 
> Session ended

My user__followedPool.csv file has 1259 lines. If I cut this file in half (~632 lines) it just works.

converge:
@rediv can you help us testing #915 ?

fabriciogava:
@converge I tested and had no change here.
With a 600 lines csv file: unfollow works sucessfully with or without #915 
With a 1200 lines csv file: unfollow does not work with or without #915 

My instagram account is set to english.

converge:
@rediv with 1200 lines in csv do you get this exactly error ?

> Logged in successfully!
> set_automated_followed_pool error
>  list index out of range
> --> Total people unfollowed : 0
> 
> Session ended

or is there something different ?

fabriciogava:
@converge  I get that exactly same error.

converge:
@rediv can you send me your csv file ? So I can reproduce the error here and debug ? If yes, joaopaulo.mailfs@gmail.com

fabriciogava:
@converge I was just about to send the file to you when I decided to double check it.
It turns out that the problem was caused by a blank line in the csv file! I simply removed it and everything works fine now.
And I can confirm that is working with or without #915

arisanguinetti:
I have the exact same error as @arjunmurali1993:

Logged in successfully!
Number of people followed automatically remaining: 9182
**('following_link error \n', 'list index out of range')**

It's fixed for me with #915

flo1t:
works for me. thanks!

converge:
[...]

Issue #898, opened 2017-11-10
Follow and unfollow don't work when using nogui=True


rvarseveld:
Hi,

I downloaded InstaPy last weekend and got it running smoothly since then. It works great although there's one big issue since a couple of days: unfollow doesn't work.

The followedPool.csv is being filled with followed account names without any problems. When a new run starts with unfollowing accounts from this pool the output shows accounts are being unfollowed but in reality only a few out of 100 accounts are actually being unfollowed.

This is my unfollow setting:

```python
session.unfollow_users(amount=100, onlyInstapyFollowed = True, onlyInstapyMethod = 'FIFO', sleep_delay=60 )
```

Should I change the sleep_delay to 600? I read somewhere that you have to wait 10 minutes after unfollowing 10 accounts. But the default sleep_delay was set to 60 when I first checked out InstaPy.

Sillanmaki:
I got some issues with unfollowing. It unfollows everybody from pool but still there was users to unfollow but they were not in the pool. So I made unfollow to its own session after following and it's now working fine after couple tests.

Like this:

```
session = InstaPy(username=insta_username, password=insta_password)
session.login()
session.set_do_comment(True, percentage=10)
session.like_by_tags(['natgeo', 'world'], amount=100)
session.end()

session = InstaPy(username=insta_username, password=insta_password)
session.login()
session.unfollow_users(amount=10, onlyInstapyFollowed = True, onlyInstapyMethod = 'FIFO', sleep_delay=60)
session.end()
```

**EDIT: Didn't solve my problem.**

pleum:
I got same issues. i try to fix it with new xpath and it worked on my pc can you try this?
https://github.com/pleum/InstaPy/commit/92240e20b6d4d8416646a64f8cb8370c5cc30d50

converge:
@pleum @Sillanmaki @rvarseveld can you guys help testing #915 ? after that, we can merge.

Sillanmaki:
My problem is that bot starts follow users which are not mentioned in console and that's why they are not in followedPool (I don't know why, when or how). I try to figure it out.

rvarseveld:
I still have the same problem and follow now also has issues. 

I set the percentage follows to 35% and in the output it says quite some users are being followed but in reality only a very few users are actually being followed (for example, only 2 users in about 1 hour).

rvarseveld:
[...]

PR #915, opened 2017-11-13
layout change for unfollow button (arjunmurali1993 fix)

converge:
thanks @arjunmurali1993, #914 #898 bug fix


converge:
hi @timgrossmann , I think we can merge it, we were talking here #914 , @paranoiasystem and @rediv helped testing it.

PR #936, opened 2017-11-16
validate user before interact in interact_user_followers

converge:
bug fix for #818 

Validation added before user interaction, so we don't need to interact if the user doesnt meets requirements

1. not in blacklist
2. not in dont_include
3. respecting upper followers count limit
4. respecting lower followers count limit
5. not interacting with private accounts


timgrossmann:
@converge Testing this right now!

timgrossmann:
@converge Awesome! 
Works like charm, thank you very much!

@converge Does this already work with the new logging?

mauricioA0 (commented):
Looks good

PR #946, opened 2017-11-18
fixed small logic issue with not setting upper follower count

timgrossmann:
- When not setting the upper follower count, the it was defaulted to 0, which is completely wrong.
- Moved around the imports and removed duplicated ones

(found while testing #936 )

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -5,7 +5,7 @@ import logging
 from math import ceil
 import os
 from datetime import datetime
-from sys import maxint
+from sys import maxsize
 import random
 
 from pyvirtualdisplay import Display
@@ -437,7 +437,7 @@ class InstaPy:
 
     def set_upper_follower_count(self, limit=None):
         """Used to chose if a post is liked by the number of likes"""
-        self.like_by_followers_upper_limit = limit or maxint
+        self.like_by_followers_upper_limit = limit or maxsize
         return self
 
     def set_lower_follower_count(self, limit=None):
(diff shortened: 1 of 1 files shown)


converge:
hey @timgrossmann , how did you realized that ? for me it's like try to find the Waldo hehe, this bug was so hidden !

take a look on this:

```bash
MacBook-Pro-de-Joao:logs converge$ python3.5
Python 3.5.2 (default, Oct 11 2016, 04:59:56)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.38)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from sys import maxint
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'maxint'
>>> from sys import maxsize
>>> maxsize
9223372036854775807
>>>
MacBook-Pro-de-Joao:logs converge$ python2.7
Python 2.7.10 (default, Feb  7 2017, 00:08:15)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from sys import maxint
>>> maxint
9223372036854775807
```

timgrossmann:
@converge Will fix this!

Haha, I just had an empty quickstart file and just put the `set_user_interact` and `interact_user_followers` there and then it skipped every single one of the users saying that it exceeds the max follower count 😉 

@converge could you check again on your system? 

Thank you

PR #955, opened 2017-11-20
random argument/parameter changes

converge:
- we're using random as argument and parameter, changed every argument/parameter from random to randomize


Naramsim:
All good!

timgrossmann:
@converge besides breaking the scripts for most users everything works fine :+1: 

This will probably lead to a ton of new issues :smile:

converge:
@timgrossmann our mothers will gain some new adjetives... lol

Issue #963, opened 2017-11-21
Missed parameter in interact_by_users()


harrypython:
Missed parameter "logger" at  interact_by_users() -> get_links_for_username()

https://github.com/timgrossmann/InstaPy/blob/786397d631d4f66d6e6f752a6e95bf62f78df7cb/instapy/instapy.py#L908-L912

converge:
merged, thanks @harrypython !

PR #964, opened 2017-11-21
Missed parameter in interact_by_users() #963

harrypython:


converge:
thank you @harrypython ! we appreciate that!

PR #969, opened 2017-11-21
Readme fix for set_unfollow_active_users

victorsc:
In order to prevent unfollowing active users, enabled=True has to be used


converge:
thank you a lot @victorsc !

PR #919, opened 2017-11-14
Smart hashtags

converge:
```python
# Generate smart hashtags based on https://displaypurposes.com ranking,
# banned and spammy tags are filtered out.
# (limit) defines amount limit of generated hashtags by hashtag
# (sort) sort generated hashtag list 'top' and 'random' are available
# (log_tags) shows generated hashtags before use it
# (use_smart_hashtags) activates like_by_tag to use smart hashtags

session.set_smart_hashtags(['cycling', 'roadbike'], limit=3, sort='top', log_tags=True)
session.like_by_tags(amount=10, use_smart_hashtags=True)
```

PR #970, opened 2017-11-22
#919 Smart hashtags Sort parameter

harrypython:
In addition to sorting random and Top (equivalent to popularity or rank in the json) we can order by relevance (same name in the json) of the hashtag.


timgrossmann:
@harrypython This is a great idea!

Thank you very much!

timgrossmann:
@harrypython Any updates on this? Would really love to merge this!

harrypython:
@timgrossmann 
To be sure about the between popularity and results I sent a direct message to faymontage, the creator of the Display Purposes, he replied my "Hello" but didn't return contact when I asked about the difference between popularity and results :/

timgrossmann:
@harrypython Interesting... If we can't explain it, I think we should not have that option there.
It probably will only confuse people

Issue #976, opened 2017-11-22
Search Locations


harrypython:
Hey everybody!

I test a way to search locations based in the Facebook API. This can be useful if you need to target an audience that frequents a particular type of location / event.

What do you think? Is adding the Facebook API worth it for this feature?

timgrossmann:
@harrypython Do you think this is also possible without the API?
If not, then build a Repo for it and we can integrate it #880

harrypython:
I'll do some research and tests. 
But I don't think this is possible without accessing the API.
It's worth remembering that the venues on Instagram are actually created on Facebook.

PR #981, opened 2017-11-22
#919 Smart hashtags Sort parameter

harrypython:
In addition to sorting random and Top (equivalent to popularity or rank in the json) we can order by relevance (same name in the json) of the hashtag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,11 +365,11 @@ session.like_by_tags(['soccer', 'cr7', 'neymar'], amount=100, media='Photo')
 # Generate smart hashtags based on https://displaypurposes.com ranking,
 # banned and spammy tags are filtered out.
 # (limit) defines amount limit of generated hashtags by hashtag
-# (sort) sort generated hashtag list 'top' and 'random' are available
+# (sort) sort generated hashtag list 'rank', 'relevance' or 'random' are available
 # (log_tags) shows generated hashtags before use it
 # (use_smart_hashtags) activates like_by_tag to use smart hashtags
 
-session.set_smart_hashtags(['cycling', 'roadbike'], limit=3, sort='top', log_tags=True)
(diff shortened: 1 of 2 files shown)


converge:
Thank you for the contribution, new ideas are always welcome. I just don't know if it will be good to the users. I like the "less ir more", that's why I added just rank and random option initially.

It's just my(personal) point of view, if you guys would like to merge it, I would love to help in everything you need. @harrypython let's wait a bit more and have more one opinion about it ?

@timgrossmann what do you think ?

harrypython:
@converge I agree with you in "less is more", but I really like more order by category of relevance than by ranking. 
When we talk about  social media the relevance is more important than number of occurrences/rank.
Anyway the conversation about is valid :)

timgrossmann:
@harrypython @converge If we should be going for the new sort params, then we definitely have to write good documentation about what each of the parameters does.

What's the difference between `rank` and `relevance`?

timgrossmann:
@harrypython Any updates on this? Would really love to merge this!

harrypython:
I'm so sorry. Did I duplicate something?

https://github.com/timgrossmann/InstaPy/pull/970#issuecomment-348289771

timgrossmann:
@harrypython No it looks like this one does not include the Facebook thing, am I right?
So these are two different PRs

harrypython:
@timgrossmann the Facebook thing is about a way to search locations and we talk about here https://github.com/timgrossmann/InstaPy/issues/976

timgrossmann:
@harrypython Do you know if the facebook api also give indications about age, gender, nationality and location?

If yes, we might really need to integrate it even though it's a separate api...

timgrossmann:
@harrypython Just saw that it's mandatory to have an instagram business account in order to use the api... this is not the right approach for instapy...

harrypython:
@timgrossmann I understand

Issue #974, opened 2017-11-22
When feature onlyNotFollowMe=True, excluded users are unfollowed


ghost:
## Expected Behavior
users added to ```session.set_dont_include(['friend1', 'friend2', 'friend3'])``` should not be unfollowed when running ```session.unfollow_users(amount=10, **onlyNotFollowMe=True**, sleep_delay=60)```
## Current Behavior
running the unfollow if not following me code  ```session.unfollow_users(amount=10, onlyNotFollowMe=True, sleep_delay=60)```. I notice that it is unfollowing users that are added to ```session.set_dont_include(['friend1', 'friend2', 'friend3'])```

wlrd:
I'll work on this one.

converge:
#998 (fixed)

PR #998, opened 2017-11-26
fix issue #974

wlrd:
Making sure to keep following whitelisted users.


converge:
trying to test it, but sounds like the following url/call is not retrieving data.

converge:
sorry for the late return, amazing fix, thank you very much ! I'll just make it PEP8 style after merge.

PR #1004, opened 2017-11-28
Update instapy.py

hojabbr:
To resolve the encoding problems


timgrossmann:
@hojjabr Does this change anything in the way the encoding works right now?

Also, could you please check the travis-CI error and change the `reload` part to be supported by python3?

Thank you, would love to merge this!

Issue #1007, opened 2017-11-28
Unfollow not working


Namaslayer:
After the new update, I am running this script

<img width="512" alt="screen shot 2017-11-28 at 1 12 08 pm" src="https://user-images.githubusercontent.com/31394074/33344878-424df15e-d43f-11e7-9589-1c19c5268536.png">

when I run it on my server, I get this error.

<img width="1277" alt="screen shot 2017-11-28 at 1 16 42 pm" src="https://user-images.githubusercontent.com/31394074/33344898-52bfcbac-d43f-11e7-8f1f-402075390dac.png">

if I run "onlyInstapyFollowed = True" it works, but I just want to unfollow everyone.

Who knows what's causing the error and how to fix it?

Thank you for your help!!!

timgrossmann:
@Namaslayer When was the last time you updated the chromedriver and InstaPy?

Namaslayer:
@timgrossmann did a git-pull about 10 minutes before posting this, using chromedriver 2.33 which I think is the most up to date.

ceribbo:
same problem here

Namaslayer:
@ceribbo Are you trying to use onlyInstapyFollowed = False too?

ceribbo:
@Namaslayer yep, also using the flag false, it still crash with the same error.
Setting it to True doesn't crash but isn't unfollowing.

rodriguesprobr:
Same error here, with chromedriver and InstaPy up-to-date.

ceribbo:
it can be simply fixed by modifying row 266 in unfollow_util.py as described in this commit: https://github.com/timgrossmann/InstaPy/pull/1009/files

rodriguesprobr:
Great!

ghost:
@ceribbo that worked, thanks!

Namaslayer:
@ceribbo you're a savior. Thank you

timgrossmann:
#1009  Thank you @ceribbo !

PR #1009, opened 2017-11-29
fix issue #1007

SlashGordon:
This change should fix the dialog not found issue during Unfollow.


rodriguesprobr (approved):

rodriguesprobr (approved):
Works fine for me.

PR #1019, opened 2017-11-30
Update for dialog selection in unfollow_util.py

SlashGordon:
I realized that after #1009 the scroll bottom method was not working correctly anymore. The dialog selection is now correct and hopefully more robust than the previous one.


converge:
fantastic ! thank you very much @alfonsodiecko

timgrossmann:
@cdieck88 Nice! This looks more like it could survive some DOM tree changes!

Thank you

PR #991, opened 2017-11-25
avoid setuid sandbox type

converge:
I tested it for 3/4 hours without **element no visible**.

(Tested on MacOS and Linux Debian)

When testing it, please remember to update InstaPy and chrome driver to latest version.

PR #1020, opened 2017-11-30
Implemented option to execute script with Headless Chrome in the InstaPy constructor

chrismaltais:
Use `headless_browser=True` in the constructor to enable Headless
Chrome - executing your scripts without a Chrome window or GUI opening.
It’s all done in the CLI! Excellent for servers or running in the
background while consuming less CPU.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Table of Contents
 * [Emoji Support](#emoji-support)
 * [Clarifai ImageAPI](#clarifai-imageapi)
 * [Running on a Server](#running-on-a-server)
-* [Running on a headless browser](#running-on-a-headless-browser)
+* [Running on a Headless Browser](#running-on-a-headless-browser)
 * [Running with Docker microservices manual](#running-with-docker-microservices-manual)
 * [Running all-in-one with Docker (obsolete)](#running-all-in-one-with-docker-obsolete)
 * [Automate with cron](#automate-with-cron)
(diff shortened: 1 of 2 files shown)


chrismaltais:
In action: http://g.recordit.co/BhEgXANLhJ.gif

Also, explanation of the lines (166-67):
```python
user_agent = "Chrome" # Replaces browser User Agent from "HeadlessChrome".
chrome_options.add_argument(f'user-agent={user_agent}')
```

a) Due to how new Headless Chrome is, some websites don't support the browser! You can get around this by changing the "ID" (aka User Agent) of the browser from HeadlessChrome back to Chrome.

b) Some sites may specifically track Headless browsers and block them for bot scripting. This is an added layer of security we can now bypass (albeit, not the most robust method haha)

timgrossmann:
@chrismaltais This is awesome!
At one point we should also compare how much better headless does compared to the initial apppraoch :wink: 

There are two more thinks I have to ask you for before we merge this...
 1. Please make sure to not use any python3 specific features like the formatted strings. We have a lot of python2 users and want to make sure to keep compatibility . Travis checks that, when your build fails, then there might be some problem

2. Please add a small section in the documentation (Readme) about that feature, what it does and how to use it.

This is great, thank you very much!

chrismaltais:
@timgrossmann I think the benchmarking is an excellent idea.

I'm actually backpacking Vietnam for a month starting next week and I'll be away from the computer, so if anyone else would like to do it I don't mind. If not, I'll look into it and submit a PR when I'm back. Cheers! 👍

timgrossmann:
@chrismaltais You're the best, thanks for giving InstaPy such a great feature!

rvoets:
I just tested this new feature but after 10 likes the script crashed with the "Element not visible" exception.

timgrossmann:
@rvoets This is really weird... This should be covered in #991 
Sorry to hear that error is still here

PR #1034, opened 2017-12-03
Daily limit based on converge/limit_daily_interactions

ugurozturk:
Please review carefully. I am not good on python. I dont want to damage this amazing InstaPy repo.

Note : it didn't auto-merge so i did manuelly, anyhow i might did something wrong in the merging. Saying again, please review carefully

### DIFF
--- a/README.md
+++ b/README.md
@@ -373,6 +373,15 @@ session.like_by_tags(['soccer', 'cr7', 'neymar'], amount=100, media='Photo')
 session.set_smart_hashtags(['cycling', 'roadbike'], limit=3, sort='top', log_tags=True)
 session.like_by_tags(amount=10, use_smart_hashtags=True)
 ```
+### Limit Daily Interactions
+
+```python
+# You can limit InstaPy actions by day. 
+# likes = 100 means it will stop InstaPy when 100 posts are liked 
+# When some limitation (likes, comments, follows, unfollows or server calls) is 
+# reached, it will stop InstaPy for this day
+
+session.set_interaction_limits(likes=150, comments=25, follows=60, unfollows=30, server_calls=1000)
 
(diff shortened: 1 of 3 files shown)


timgrossmann:
@ugurozturk Thank you very much!
We will review this soon.

converge:
Hey mate, it will be great, just need some little improvements.

- we're using a new logging system, and we show log informations in this way (ex.):
```self.logger.info('Session started')```

- we can avoid repeating verification for every method and use only one verification at update_statistics method, to keep InstaPy working after it, we could have something like an action control (able to like, but comment is not available because it already reached it's daily limit)

ugurozturk:
is self.logFile.write still avaible or no-needed ? 

https://github.com/timgrossmann/InstaPy/pull/1034/files#diff-de2ecb8866219bb53ba97e8377a2f26dR425

converge:
hey @ugurozturk, it's not needed, neither print, just logger is enough to do the magic.

ugurozturk:
- [x] logger system
- [ ] avoid repeating verification
-- couldn't find any solution for it. Couldn't use limit parameters inside update_activity so couldn't send it to check_activity_limits. Tried to create an table and go with it then got problems.
-- Using sqlite: What will happen if a person did use the limitions then will not use anymore. The table would be there so the inside check_activity_limits how to know if the person is using or not
- [ ] able to like, but comment is not available because it already reached it's daily limit.

beranPro:
+1

luigifaticoso:
This would be awesome!

timgrossmann:
@ugurozturk Any updates on this?

ugurozturk:
@timgrossmann nope and closing this

PR #1039, opened 2017-12-04
Extension Feature

converge:
The idea is host the new extensions feature inside extensions folder, to be able to import extensions into quickstart.py we need extensions folder as a package folder

tree would be:
> ./extensions
> ./extensions/instapy_telegram_extension

and we will import it in quickstart.py as:
> from extensions.instapy_telegram_extension import InstaPyTelegramExtension

I prepared a very basic extension:

https://github.com/converge/instapy_telegram_extension

I don't like the way it need to be downloaded/installed, is there a better approach ?


timgrossmann:
@converge This is a really good idea! 

You're totally right, the way we will need to download and put the extensions into the folder is really bad... But as of now, this might be a good start to make people develop new extensions. 
As we go, we can think about how to improve it.

Best would be to have two ways of doing it. 
- Having a little `npm`-like cli where more code prune people could easily do `ipm install telegram-extension` (ipm = InstaPy package manager 😄)

- Later on once there is a packaged solution, we most likely want to have a tab in the GUI in which we simply can browse extensions and install them from there (for the less technical people)

What do you guys think?

I think to start it, it would be perfectly fine to have people download the tool from github and put it in the extensions folder (maybe go for the cli tool from the begnning? It could be "easily" installed with `pip install`)

All comments, critics and improvements are very welcome

eracle:
Hello there!
Sorry, maybe my comment is inappropriate, but, I think reimplement another package manager seems too heavy, and error prone, why not using the "django way" to manage extension.
I mean, there could be the concept of app (or extension, or plugin) and they could all be managed by using the configuration file.
Regards

timgrossmann:
@converge How far are we with this?
Can we add a section in the ReadMe on how to 'install' your telegram extension into InstaPy?
I'd really love to see this getting used by people

converge:
@timgrossmann , I found out a new way to have it in a professional way following @eracle idea.

./extensions/**manager.py**

python **manager.py** help
- manager.py search package_name
- manager.py install package_name
- manager.py uninstall package_name

**extensions.json**:
"telegram_extension" [
{
  "package_name": "telegram_extension",
  "version": 0.1,
  "author": "João Vanzuita",
  "repository": "https://github.com/converge/instapy_telegram_extension"
}
]

converge:
I would love some help to develop this feature, is there anyone available ?

eracle:
Hello there!
Sorry, I think my proposal was misunderstood.
The django way uses classical python package manager pip, by prefixing every django "app" with the "django-" prefix.
For instance, if REST functionalities want to be added to a django project, there is the django-rest-framework python library that can be downloaded and added to the installed_apps configuration variable in the settings.py file.

For instance (django's example): 

`````python
INSTALLED_APPS = (
    'django.contrib.auth',
    'django.contrib.contenttypes',
    'django.contrib.sessions',
    'django.contrib.sites',
    'django.contrib.messages',
    'django.contrib.staticfiles',
    'django-rest-framework',
}
`````

So no need to have our own instapy package manager, but using instapy-* prefix for the plugins + standard pip command.
To tackle the aforementioned case: instapy-telegram could be a good naming for the library that offers the telegram plugin. 
So:
`pip install instapy-telegram`
plus: including 'instapy-telegram' in a installed_app (or installed_plugins) variable in the main instapy's configuration file.
I'm not too aware of the instapy's anatomy to propose more deeper implementation details, but I can suggest to have a procedure in the instapy's entry point (or main, or boot phase) which reads this installed_apps variable and reads the content of all the issued apps. In the django's case it basically adds models, migrations, urls and other stuff to the project by following a convention over configuration approach.
Regards

converge:
It will require much time and unfortunately I don't have it at the moment, would be nice if someone else could work on it.

douwevandermeij:
I like this approach, especially with `pip install ...`, I can help on this part. First step may be as well my PR #1663 to make InstaPy itself pip installable.

douwevandermeij:
See https://github.com/converge/instapy_telegram_extension/pull/5

Issue #1062, opened 2017-12-08
CRITICAL - Wrong login data!


assafelovic:
## Expected Behavior
I've been running InstaPy for quite a while now, thank you very much for this amazing community! 
In the past few days, I've been running into issues logging into already verified accounts. I usually expect "Logged in successfully!" alert after starting the script from a Linux server, however, logging in fails. 

## Current Behavior
The login attempts fail again and again.  I also can't input a verification code. 
I also click on "It was me" from the mobile app, and when I try to login again it still fails.
I've took an image of the browser and this is what it shows:
![screenie](https://user-images.githubusercontent.com/13554167/33768885-8b3ed056-dc30-11e7-861b-44d36bc314ce.png)

## InstaPy configuration
I've updated chromedriver to the latest (2.33) and am running the latest InstaPy version. I'm using a AWS linux server.

Thank you!

TGIFelix:
Having the same issue running at digital ocean.

timgrossmann:
@assafelovic Simply login with your account from the browser once and verify it. 
Once you did, everything should work fine

TGIFelix:
What if I'm running headless?

nsgolosoff:
The same problem :( Instagram requires verification each time, even after logining in and verifying from browser. There were no problems three days ago

thomasuno:
no @timgrossmann it's not working. I think we should add cookie management to the software.

I'm testing your software with some modifications to keep session cookies, and it's working. :-)

No need to log-in every time.

1) login first time with VNC and do authentication with code verification
2) save cookies
3) do instapy work and then quit 

4) start again instapy? no login! load cookies and go

nsgolosoff:
@thomasuno can you share your modifications, please?

thomasuno:
yes @nsgolosoff  I'll do here because last time I tried to commit something, someone hasn't appreciated 

*** file instapy.py ***

1) after line self.logger.info('Logged in successfully!') add this line (Remember indentation)

pickle.dump(self.browser.get_cookies() , open("instacookies.pkl","wb"))

3) at the beginning add import pickle

4) touch file instacookies.pkl (create a blank file named that way)

***

4) NOW start a session of instapy and do verification trought vnc

5) then edit your python script and comment #session.login()
add after these lines

session.browser.get("https://www.instagram.com")
for cookie in pickle.load(open("instacookies.pkl", "rb")):
   session.browser.add_cookie(cookie)

and at beginning this 

 import pickle

[...]

PR #1064, opened 2017-12-09
Fix Login https://github.com/timgrossmann/InstaPy/issues/1062

sionking:
Changes only in login file.
if Cookie file exist go with the cookies.
If not exist than wait for successful login and dump cookie to file for next login.

Thanks to @thomasuno

SC.


instapy/login_util.py
```diff
@@ -2,55 +2,63 @@
 from .time_util import sleep
 from selenium.webdriver.common.action_chains import ActionChains
 from .util import update_activity
-
+import pickle
 
 def login_user(browser, username, password, switch_language=True):
     """Logins the user with the given username and password"""
-    browser.get('https://www.instagram.com')
+    browser.get('https://www.google.com')
```
Naramsim: Why google?

sionking: it is suspicious to log to Instagram with no login cookie and only after to set the cookie.
We need to act human

sionking: But I see I missed the Instagram get url ... I pused a fix

timgrossmann:
@sionking This is great, I'll make sure it works as expected later today and we can merge it once some people said so as well.

timgrossmann:
Sorry... accidentally clicked `Close and Comment`...

Naramsim:
Hi guys, I think that would be another great idea to add to the Instapy constructor a cookie field so that if we manage to get by ourselves the cookie (for those headless), we can pass it to Instapy

`session = InstaPy(cookieFile='./whatever')`

Also, it would be good to know how the cookieFile has to be structured. I see that in the PR `pickle` was used, I don't really know how it serializes data. Maybe a better serializer could be used like the native `json` [module](https://docs.python.org/3/library/json.html). So that we can have the structure and fill it with our cookies data.

timgrossmann:
@Naramsim Great idea!

CharlesCCC:
I tested it today, working perfectly fine.

JulienDev:
Hi,
Thanks for the PR! It doesn't work on Firefox : 
"selenium.common.exceptions.WebDriverException: Message: You may only set cookies for the current domain"
Also, I think it would be better to have one cookie file per user

CharlesCCC:
@JulienDev that will be good idea, but How do you mange have multiple user within the same folder without create multiple InstaPy  folders ?

JulienDev:
We could just have a different name for each file eg : myAccount-Cookies.pkl
I think we need it because if we try to login with cookies from another account we could have problems with the accounts

ceribbo:
Cookies: great idea!
It would be perfect if after the first login instapy saves the cookie for each account and the keep reusing it for the next sessions until the expiration date of the cookie.

sionking:
IMPORTANT: updated the chromedriver and chrome and I had issues with liking pages redirected me to error page. I needed to recapture the cookie file.

converge:
#1074 #1125

PR #1079, opened 2017-12-11
#1074 bugfix (wrong login data/suspicious attempt)

converge:
* when suspicious account is detected, send security code to users email, and ask it in console.

PR #1097, opened 2017-12-14
Fix Suspicious Login Attempt - Wrong login data

nilkesede:
I put together #1079 #1064 in order to solve the `suspicious attempt` problem.
I'm not a python programmer so feel free to correct me if I made some mistake.

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -132,10 +132,6 @@ class InstaPy:
                          'We\'re working to fix this soon!')
             self.logger.critical(error_msg)
 
-    def get_dont_like(self):
-        for i in self.dont_like:
-            print(i)
-
     def set_selenium_local_session(self):
         """Starts local session for a selenium server.
         Default case scenario."""
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -54,8 +54,7 @@ class InstaPy:
                  use_firefox=False,
                  page_delay=25,
                  show_logs=True,
-                 headless_browser=False,
-                 bypass_suspicious_attempt=False):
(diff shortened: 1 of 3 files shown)


Naramsim:
Could you please change `pickle` with `json`? Reasons:

* Having cookies saved as `json` is much more readable (human-readable). 

* With `json` we can dump/update by ourselves the cookie file if we are running headless.

* `Json` and pickle have basically the same methods (`dumps`, `loads`).

* We don't really need to save cookies as a binary file.

* `Json` is much [faster](https://konstantin.blog/2010/pickle-vs-json-which-is-faster/) and [safer](https://stackoverflow.com/a/2259351/3482533)

instapy/instapy.py
```diff
@@ -132,6 +132,10 @@ def __init__(self,
                          'We\'re working to fix this soon!')
             self.logger.critical(error_msg)
 
+    def get_dont_like(self):
```
converge: do we need it ?

converge (changes requested):
hey mate! nice ideas!

Unfortunatelly it doesn't login for me on my VPS, I receive the wrong login data message.

another point is, we should avoid try to bypass suspicious attempt page every time we login, it increases the delay to login.

instapy/instapy.py
nilkesede: we don't really need it

nilkesede:
thanks, man!
it doesn't try to bypass every time, just when we have the suspicious attempt page. but i got you, it really increases the delay to login.

converge:
#1074 #1125

PR #1124, opened 2017-12-22
fix #1110

### LINKED ISSUE #1110: Where are the unfollowed users stored?
I'm new to instapy and I'm trying to find out where instapy stores the users, which have been followed and unfollowed, to make sure they are not followed a second time.

My understanding is, that the log/usernameFollowedPool.csv keeps the actual followed users, but not the ones who have been followed previously.

Thanks for your help!

nilkesede:
`follow_user_following` and `follow_user_followers` wasn't checking for `follow_restriction` and `follow_times`


beranPro:
@timgrossmann 

+1

Issue #1096, opened 2017-12-14
How can I follow by tag?


alessiomtt:
I know that maybe it's a very stupid question, but how can I follow a profile that have posted a photo whith a specific tag without liking the image?

mvaldes14:
you mean following people with certain tags in their posts?
or you want to see all posts tied to a tag sort of like twitter?

camilaavilarinho:
I did a modification of the function like_by_tags and was possible to follow the user without liking the image. It worked for me, but I don't know if is good enough to do a PR...

timgrossmann:
@camilaavilarinho Please feel free to share your changes, I'm pretty sure that all of us can work on it together if that is a feature people want to have.

Thank you very much and Happy Holidays!

PR #1135, opened 2017-12-25
Add follow_by_tags

camilaavilarinho:
Modification of the function like_by_tags to make possible follow a user without liking the image.

#1096


timgrossmann:
@camilaavilarinho Thank you very much for contributing this!

Even though I think in the longer run, the methods should all be switched to `.interact_by_xxx` where we can set all the different things like `.set_do_like`, `.set_do_comment` and  `.set_do_follow` to make sure that there is only one method to use, this most likely will help a lot of people.

Ivan-Klabucar (approved):

timgrossmann:
@camilaavilarinho You definitely have to update the Readme to make sure your feature is documented, please :wink:

PR #1137, opened 2017-12-26
Memory leak fix

converge:
Just found out that our memory leak issue is caused when:

1. start InstaPy
2. InstaPy generates some error/exception
3. self.end() method is not called in the process
4. chrome and xvfb keeps open, consuming resources (cpu / memory)

### DIFF
--- a/quickstart.py
+++ b/quickstart.py
@@ -1,7 +1,7 @@
 from instapy import InstaPy
 
-insta_username = ''
-insta_password = ''
+insta_username = 'cycling_apparel'
+insta_password = 'lx7qt5'
 
 # set headless_browser=True if you want to run InstaPy on a server
 try:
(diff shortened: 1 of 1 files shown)


timgrossmann:
@converge Yes, that's the same problem I tried to fix with the other memory leak PR.

converge:
@timgrossmann cool! I saw your PR about it, it was exclusively calling end() when ctrl+c is hit, right ?

This approach calls always call end() whenever an error/exception is raised or the script reach its end.

What do you think ?

Naramsim:
@converge There is your account password in the PR!!!!!!!!!!!!!!!!!!!!!!!!

converge:
@Naramsim thanks! that was funny.

Fortunately is a testing account.

Naramsim:
Yeah, consider that we are using Git, so everything remains, you should remove that commit.

https://github.com/timgrossmann/InstaPy/pull/1137/commits/44cfac52faa78005f1fcfead0f3fe8bfe5414968

converge:
@Naramsim no problem, I don't use that passwd for other stuffs.

PR #1151, opened 2017-12-28
Adding the capability to define a custom `user-agent` on chromedriver

ArturGrigio:
I saw that there was no way to change the `user-agent` other than setting the `headless_browser=True`.

It might look weird to Instagram if the user is logging in with 2 different devices.

I also updated the documentation to reflect the new functionality.

*Note:* I had to add an additional redirect inside `login_util.py` to go around `#reactivated` prompt when logging in using a phone `user-agent`


main.py
```diff
@@ -0,0 +1,71 @@
+from instapy import InstaPy
+import schedule
+import time
+from time import sleep
+import sys
+
+# insta_username = 'futiconaki'
+# insta_password = 'al7217'
```
converge: your passwd

Naramsim:
Be careful that in some cases changing UA will also change the rendered DOM, causing Instapy to not properly work. Add a note on the readme.md

timgrossmann:
@ArturGrigio Thank you for contributing! 
We'll review your PR shortly and get back to you with any changes.

What is the `fan.util` for?

converge (changes requested):

Naramsim:
This guy...

sionking:
I don't understand how user agent == Phone while the bot run a PC web.
It will not make sense to Instagram .... Only if you log from phone and choose desktop view.

Issue #451, opened 2017-07-30
run 3 account at the same time


spikerang:
i have 3 accounts. if i run 3 programe at the same time with same computer, same ip, will i get banned?

bardascat:
I'am also interested about this question...

balakedev:
@bardascat , @spikerang 

I currently run 2 at a time, but I have 3 unique accounts that I run instapy on. I have done extensive research on this and this is what I have discovered:

1. Running multiple accounts from the same IP address is dangerous. Instagram looks at activity from IP addresses and if there is too much coming from the same IP address it will assume that there is suspicious activity at that locaton; ie a bot farm - this can lead to a shadowban

^^^ that being said, I've run 2 accounts, daily and the 3rd once every three days or so for almost 40 days now on the same computer and I haven't had a problem. Each account at a maximum likes 279 posts in total and normally passes around 300 as I have A LOT of banned words to keep the bot as accurate as possible to specific industries. 

2. The location of the IP address is important. if you have instapy running on an account that is regularly accessed in Melbourne, Australia, and the bot is being highly active from another geographical location (like Los Angeles) within an unreasonable time frame - this can also be seen as suspicious and can lead to a shadowban. 

There is evidence to suggest that instagram now allows a single account to be accessed by multiple devices, as Instagram is aware that is a driving force behind marketing for companies all over the world. 

What I have mentioned above is just what I've read and experienced. Could be totally inaccurate tho haha.

All the best

bardascat:
@blakemmw - thanks for tbe reply.
Do you know if the "multiple accounts" problem can be avoided using some proxy servers ?
Or instagram would know that you are using proxy and ban your account. ? 
Any info on this one ?

mathmorrison:
I want run raspberries 8 simultaneity together to ride the power shouldn't I get banned?

RomeoJuliett:
Thanks for clearing that up Blakemmw. I was actually thinking of running multiple Ubuntu instances in order to make this happen properly. Good to know we can run a small amount from the same IP.

wizarduk:
You would need more ram so if you wanted lots running it would cost you more mine is running 61% ram on 1GB ram so if i run two im maxed out...

wizarduk:
Yes running two accounts is fine but all depends some ig bots you can run 10 on 1 account depending what its doing///

bardascat:
Why would you want to run multiple instances of bot for one instagram
accont in the same time ?

[...]

PR #1150, opened 2017-12-28
Parameter to save log files by username #1149

dvlprlife:
Fixes #1149 - Add a parameter to determine if the log files should be stored in a separate log file subfolder by username use the existing one log folder structure

PR #1175, opened 2018-01-01
Added multi-account example #451

rzfzr:
I had no problems running from the same IP.
The code is still crude but works fine, I'll be working on it and accepting suggestions.

### DIFF
--- a/examples/multi-account.py
+++ b/examples/multi-account.py
@@ -1,0 +1,53 @@
+import random 
+from instapy import InstaPy
+import os
+import multiprocessing
+import datetime
+import time
+
+instaUser = ['username1','username2'] 
+instaPass = ['pass1','pass2']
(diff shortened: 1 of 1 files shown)


CharlesCCC:
@rzfzr this will be a good combination with the [PR-1150](https://github.com/timgrossmann/InstaPy/pull/1150)

rzfzr:
@CharlesCCC Sure, I guess it's a recurrent theme to have multiple users, I edited the source files to print the username whenever it prints an "INFO" or "-->", I plan to make it more elegant by overriding logger.info to print it, then I'll share it. Maybe then look into the logs.

dimahadgi:
So multi-account is working? Can I start using it?

timgrossmann:
@dimahadgi This is mostly an example on how you could do it. 
It's not the final thing but a start to make people aware that we are working on it and how they can use it right now (Note: This does not take separate log files into account yet)

PR #1192, opened 2018-01-04
Fix unfollow=False issue in "Like by Feed" def

uluQulu:
Either way, unfollow=False will not stop unfollowing inappropriate media owners. Actually, stopping unfollow in certain inappropriate medias is **vital**. Such as, i dont want to like #food pics but it should not unfollow everybody with that tag.
 Or maybe, **much better**, there could be some improvement to unfollow tool, like if it gets certain amount of inappropriate tags in a media, it should do unfollow (e.g. very inappropriate).

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1449,7 +1449,7 @@ class InstaPy:
                                 self.logger.info(
                                     '--> Image not liked: {}'.format(reason))
                                 inap_img += 1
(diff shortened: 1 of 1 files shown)


timgrossmann:
@uluQulu How could we miss that, thank you very much!

uluQulu:
@timgrossmann Glad I helped :)

PR #1224, opened 2018-01-12
telegram_util.py

LoopForYou:
Notifies you (or someone else) that the bot has started, finished or is paused. This is better than using SMS because Telegram is free and easy to use.

I am not that good at Python and i'm still learning, but this code works when I put it in quickstart.py and I would like your help to make the code better and have it as its own file so we can just call the functions from quickstart.py

You can find more details for integrating this util in the file comments

### DIFF
--- a/instapy/telegram_util.py
+++ b/instapy/telegram_util.py
@@ -5,23 +5,24 @@ import telegram
 from telegram.error import NetworkError, Unauthorized
 
 update_id = None
+insta_username = ""
 
 
-def start():
+def start_telegram():
     """Run the bot."""
     global update_id
     # Telegram Bot Authorization Token
     bot = telegram.Bot('<token>')
     echo_start(bot)
     
-def finish():
+def finish_telegram():
     """Run the bot."""
     global update_id
     # Telegram Bot Authorization Token
     bot = telegram.Bot('<token>')
     echo_finish(bot)
     
-def pause():
+def pause_telegram():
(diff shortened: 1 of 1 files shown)


timgrossmann:
@LoopForYou This is awesome! Thank you for contributing.

Could you please add some documentation to the README and maybe an example in the examples folder?

Also, make sure to fix the travis-ci errors.

Thank you very much

LoopForYou:
Fixed the travis-ci error. Going to add documentation to README later tonight and an example in the examples folder.

thx

timgrossmann:
@LoopForYou Thank you very much for this!
I'm not sure if we should put this into the instapy core module itself .
@converge aslo has a telegram extension that can be downloaded and put into the extensions folder

sionking:
Maybe it is good idea to send telegram msg when there is no activity (maybe crash)

timgrossmann:
@LoopForYou @sionking We should definitely merge this with @converge [telegram extension](https://github.com/converge/instapy_telegram_extension) and merge it into InstaPy

timgrossmann:
@LoopForYou I think we'd use #1039

sionking:
I already uses telegram when I have no action(like/follow) for last hour (external tool check the statistics for each user every hour).
I also use it when follow/like block detected.


Will PR soon

timgrossmann:
Awesome!

PR #1228, opened 2018-01-13
Update README.md with telegram_util.py instructions

LoopForYou:
Added Telegram Notifications section


timgrossmann:
@LoopForYou I think we'd use #1039

PR #1229, opened 2018-01-13
Create telegramexample.py

LoopForYou:
Example for showing how to use start_telegram(insta_username), pause_telegram(insta_username), finish_telegram(insta_username)


timgrossmann:
@LoopForYou I think we'd go for #1039

Issue #1165, opened 2017-12-30
Personalize comment with each user name?


bsoum:
Hello,

It's possible to comment a post with the user name account? Exemple: 'Great @user! 😃🎉👏'

It's just for personalize each comments.

Thank you.

juan0822031:
#1165

juan0822031:
@caroflores___

timgrossmann:
@bsoum Not possible at the moment... However, this should be pretty straight forward to implement.
If you put a bounty on this issue, I'm sure there soon will be someone having a working solution  for you ready.
https://www.bountysource.com/issues/53225694-personalize-comment-with-each-user-name?show_profile_modal=1

Thank you, all the best and
Happy Holiday

aclowkey:
#1240
It could also be possible to add other users than OP to the comment. 
For example: I want to comment on memes, tagging other random pages on instagram.

What do you think?

PR #1240, opened 2018-01-16
Personlizing comments with username

aclowkey:
#1165 

I considered adding a flag whether to format a comment or not, as it is possible to have many comments without format. However, comments are fairly short and I think it wouldn't affect performance too much.


davidrocket:
Great feature :)

Issue #894, opened 2017-11-09
Like multiple photos per profile


jksyrek:
Is it possible to like multiple photos per account? In the case the bot is looking for a hashtag, can I set the option to like not only one photo but also a few other of the same author?

converge:
hello @jksyrek , it's possible, check the docs for better understanding: https://github.com/timgrossmann/InstaPy#instapy-available-features

jksyrek:
As I've read that it is possible for the photos from thee feed using
```session.like_by_feed(amount=100, randomize=True, unfollow=True, interact=True)```
where `interact` visits the author's profile page of a certain post and likes a given number of his pictures, then returns to feed.

But I'm looking for such a option for the pictures on the hashtag search page. So basically `interact` flag for the `session.like_by_tags(['natgeo', 'world'], amount=10)` command.

Or basically something like  `session.interact_user_following(['natgeo'], amount=10, random=True)` but not for an user but for a hashtag. Like `interact_by_tag` function or something like that.

**Comment by converge** [comment]
hey @timgrossmann , should
```python
session.set_user_interact(amount=3, random=True, percentage=100, media='Photo')
session.like_by_tags(['cyclingshoes'], amount=5)
```
do it ?

timgrossmann:
@converge Yes, that should work out

converge:
@jksyrek please, reopen it if you still need some help.

PR #1241, opened 2018-01-16
added interact option to like_by_tags #894

Onatcer:
I had the same issue mentioned in #894 and did not get it to work, so I basically copied the interact parts from like_by_feed and added them to like_by_tags. 

But the like_by_user called in like_by_tags didn't seem to work without a proper like_by_followers_upper_limit (because of util.py:31 with like_by_followers_upper_limit set to 0). So i figured the most simple way is to change the default value of like_by_followers_upper_limit to 90000. I don't really see a reason why it should make any problems, but i'm not 100% sure that's why i mentioned it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,6 +346,14 @@ Example:
 session.like_by_tags(['natgeo', 'world'], amount=10)
 ```
 
+### Like by Tags and interact with user
+
+```python
+# Like posts based on hashtags and like 3 posts of its poster
+session.set_user_interact(amount=3, randomize=True, percentage=100, media='Photo')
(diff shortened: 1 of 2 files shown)


luigifaticoso:
This would be awesome!!

timgrossmann:
@Onatcer This is awesome!
Thank you for contributing.

Just make sure that you fix all the errors that travis-ci is throwing here :wink:

Onatcer:
@timgrossmann I'm sorry if that's a stupid question, but I do not have a lot of experience with python and these sort of build testings, but I just tried to build my repo with the changes here (https://travis-ci.org/Onatcer/InstaPy) and it worked, I don't really get why this is the case? What am I missing?

PR #1255, opened 2018-01-20
Update logging instantiation

Marto32:
I don't believe there was an issue for this, but I encountered an error where the log messages were duplicating themselves multiple times. The issue had to do with instantiating the `logging.getLogger(__name__)` singleton multiple times.

I was using the [retrying library](https://github.com/rholder/retrying) in my quickstart file as follows:

```python
@retry(stop_max_attempt_number=20)
def main():
    session = InstaPy(username=insta_username, password=insta_password)
    ...
```
However, because I would re-instantiate the `InstaPy` object upon each retry, it would add additional loggers to the existing logging instance. This resulted in duplicate output.

I have updated the logging setup in the instapy file to avoid this issue (by retrieving an existing logger from a cache if one has already been created).

This appears to be working great for me.

Open to any feedback as needed. Hopefully this solves similar issues that others have seen.

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -43,6 +43,10 @@ from .unfollow_util import dump_follow_restriction
 from .unfollow_util import set_automated_followed_pool
 
 
+# Set a logger cache outside the InstaPy object to avoid re-instantiation issues
+loggers = {}
+
+
 class InstaPy:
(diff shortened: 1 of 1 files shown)


timgrossmann:
@Marto32 Works like charm, thank you!

PR #1277, opened 2018-01-24
missing self.logfolder argument to follow_user

gregsadetsky:
`follow_by_tags` calls were failing since `follow_user` was called with 6, instead of 7 arguments, as `self.logfolder` was missing.


arypurnomoz:
Can we merge this?

timgrossmann:
@gregsadetsky Great Catch!

Thank you very much for your contribution.

PR #1280, opened 2018-01-25
fix the login issue with the new login page layout

CharlesCCC:
commented out the line for now until further fix created.

### DIFF
--- a/instapy/login_util.py
+++ b/instapy/login_util.py
@@ -110,13 +110,14 @@ def login_user(browser,
     # Changes instagram language to english, to ensure no errors ensue from
     # having the site on a different language
     # Might cause problems if the OS language is english
-    if switch_language:
+    '''
+	if switch_language:
         browser.find_element_by_xpath(
             "//footer[@class='_s5vm9']/div[@class='_g7lf5 _9z659']/nav["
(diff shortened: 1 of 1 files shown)


ganeshkrishnan1:
We don't have to comment out that line. Just set the switch_language=False in quickstart.py

deronsizemore:
switch_language=False doesn't seem to be working for me. Can someone confirm that it's the fix for this?

deronsizemore:
I figured it out after looking through some of the other similar issues. self.switch_language = True needs to be set to False inside of instapy.py.

Setting it to Flase in quickstart.py didn't seem to work for me.

uni1783:
In quickstart.py you need to set it as session.switch_language=False

spoonyto:
@uni1783 Thanks! Actually works!

fegomess:
I still can't make it work. It just don't create the cookie file. By the way, Always when instagram changes my password, i had to delete the cookie file from log folder, then run the script. Now it just dont login... it trys to create a cookie file and crashes. Help, pls

edit:
i tryed everything u guys posted here, but nothing worked for me. Here is what worked: i changed it to Firefox browser, so it could create the freaking file, then i switched back to headless chrome, and TADAA, its working now.

farschmessivo:
@fegomess hi, can you please share the steps how to use FF instead of Chrome? 
thnx

PR #1287, opened 2018-01-25
BUG: set link as inappropriate if contains string in ignore_if_contains

lsternlicht:


instapy/like_util.py
```diff
@@ -505,7 +505,7 @@ def check_link(browser,
         return True, user_name, is_video, 'Username'
 
     if any((word in image_text for word in ignore_if_contains)):
-        return False, user_name, is_video, 'None'
+        return True, user_name, is_video, 'None'
```
lsternlicht, line 508: If the link contains a word in ignore list, then we want to mark it as inappropriate (i.e. return True)

sionking:
ignore_if_contains is to ignore the set_dont_like if a word exist in line for example:
line is : "sexy girl do laundry"
set_dont_like = sex
ignore_if_contains = laundry

than this line is OK

sionking:
I do agree name shall be changed.

timgrossmann:
@lsternlicht Thank you very much!
Good thing you spotted this mistake.

Issue #1242, opened 2018-01-16
Error 'onlyNotFollowMe'


guilhermeherzer:
Hello,

When I try to execute onlyNotFollowMe gives an error.

ERROR - set_automated_followed_pool error [Errno 2] No such file or directory: './logs/guilherme.sublop_followedPool.csv'
unable to get followers and following information
 'after'

Help me

aleducode:
same problem...

gregsadetsky:
Here's the fix: change https://github.com/timgrossmann/InstaPy/blob/master/instapy/unfollow_util.py#L163

from

```
del variables['after']
```

to

```
if 'after' in variables:
    del variables['after']
```

making sure to keep the indentation correct (i.e. the `if` should be at the same level than where the `del  ...` was. the `del` statement should now be one level more indented).

I'll open a PR later for this..!

luigifaticoso:
If the problem is solved, you can close the issue

PR #1295, opened 2018-01-27
fixes an issue when running unfollow_users with onlyNotFollowMe

gregsadetsky:
a chrome "element not found" error happens -- the specific Python exception mentions a KeyError on the 'after' key. the commit in this PR fixes this issue.

this relates to issue #1242


timgrossmann:
@gregsadetsky Awesome eye for catching bugs.

Thank you very much!

PR #1296, opened 2018-01-27
add support for unicode smart_hashtags

gregsadetsky:
this fixes exceptions when attempting to use unicode (i.e., non-ascii) smart hashtags


timgrossmann:
@gregsadetsky Another great catch!

Thank you for splitting them up into different PR's.
Makes it really easy to review

PR #1308, opened 2018-01-29
Change from 10 in the langunage bar to 11

dar2015:
Instagram added another option that make script fail when start.


timgrossmann:
@dar2015 Great you caught that!

Thanks for your contributions!

Issue #1311, opened 2018-01-30
NameError: name 'session' is not defined


sergiodfg:
Hey all, went through the setup process as described. Only added my username and password. Ran quickstart.py and expected to all work well, but I got this error: 
<img width="730" alt="screen shot 2018-01-29 at 9 22 35 pm" src="https://user-images.githubusercontent.com/31285842/35548222-37ce17ea-053b-11e8-9068-6443bc05210c.png">

tried running it like this, still nothing: 
<img width="335" alt="screen shot 2018-01-29 at 9 21 37 pm" src="https://user-images.githubusercontent.com/31285842/35548295-bb8318ec-053b-11e8-8d60-9ac1e2fa8b5c.png">

running on newest osx release, fresh python install with all dependencies. i've tried troubleshooting but not sure why it's not running, since I'm just learning python right now.

davegallant:
d1f5cfddf08c55173d333cb6e551c1ae4d30cd99 broke it.

This commit needs to be fixed/reverted.

Temporary workaround:
```
git revert d1f5cfddf08c55173d333cb6e551c1ae4d30cd99
```

A suite of pytests would be nice on PRs. ;)

self.logger is being referenced before it's created.

sergiodfg:
thank you, works great now! I was trying everything but who would've known it was as simple as switching to an old commit

uluQulu:
@astroserg 
line 165 in instapy.py
### Replace both self.logger to logger
[read](https://github.com/timgrossmann/InstaPy/issues/1310#issuecomment-361485160)

aledebessa:
@davegallant 
Hello, i'm iniciating on python right now and i got the same error just now, so i don't have much knowledge on coding. How/where should i fix it? 

I put this on the prompt: 

`git revert d1f5cfddf08c55173d333cb6e551c1ae4d30cd99`

as you said, but i got this:

error: Reverting is not possible because you have unmerged files.

Don't know what to do...


Best regards.

davegallant:
@aledebessa, try to the pull the latest?

If that doesn't work:
```
git stash
git revert d1f5cfddf08c55173d333cb6e551c1ae4d30cd99
git stash apply
```

And take a peak at [git documentation](https://git-scm.com/docs/git-stash). :)

aledebessa:
@davegallant Have just pull the latest. It has not the error, but some others. I will search and if i don't find anything i'll create a topic.

Thank you!

PR #1309, opened 2018-01-29
Added username to log

dar2015:
Ability to track users action in log if there is multi users using same script.

PR #1313, opened 2018-01-30
self.logger referenced before it's created.

davegallant:
trivial fix

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -162,7 +162,7 @@ class InstaPy:
                 console_handler.setFormatter(logger_formatter)
                 logger.addHandler(console_handler)
             
-            self.logger = logging.LoggerAdapter(self.logger, extra)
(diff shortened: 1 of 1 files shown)


instapy/instapy.py
```diff
@@ -162,7 +162,7 @@ def get_instapy_logger(self, show_logs):
                 console_handler.setFormatter(logger_formatter)
                 logger.addHandler(console_handler)
             
-            self.logger = logging.LoggerAdapter(self.logger, extra)
+            self.logger = logging.LoggerAdapter(logger, extra)
```
CharlesCCC, line 165: I believe this should be `logger = logging.LoggerAdapter(logger, extra)`

davegallant: @CharlesCCC You're right. I can fix and test this tonight. I was unsure about what the the logging was doing (#1309) and tried to quickly put out this fire: #1311

PR #1325, opened 2018-02-01
Showing the matched Inappropriate word or the word contained it

uluQulu:
In many places we use `check_link` to find out if it contains any inappropriate word included in 
```python
session.set_dont_like(tags='badword1', 'badword2')
```
If a post contains any of these tags, it will not take actions. And will output,
### Before:
`Image not liked: Inappropriate`

But for me, having a huge list of blacklisted words (about 700 unique words) I do get that message so often, that's why it is interested to me that which word is Inappropriate. So I modified it,
### After:
`Image not liked: Inappropriate! ~ contains 'badword1'`   #if matched word is same as blacklisted word
`Image not liked: Inappropriate! ~ contains 'badword2' in 'very_badword2'`

> After seeing the output, I did remove some of the not important blacklisted words :-)

For any user, this will improve awareness, rather than mystic plain `Inappropriate`...

_Also, added the fix for the latest commit where both of the instances of `self.logger` should be `logger`_

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -161,9 +161,9 @@ class InstaPy:
                 console_handler.setLevel(logging.DEBUG)
                 console_handler.setFormatter(logger_formatter)
                 logger.addHandler(console_handler)
-            
-            self.logger = logging.LoggerAdapter(logger, extra)
-            
+
+            logger = logging.LoggerAdapter(logger, extra)
+
             loggers[__name__] = logger
             return logger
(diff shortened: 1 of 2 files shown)


sionking:
thanks

timgrossmann:
@sionking Can you approve this PR?
Is everything working?

Thank you very much @uluQulu for this contribution, I'm pretty sure this will help a ton of people!

Issue #992, opened 2017-11-25
Delete line from file - > error/PC down can cause the file to be malformed. (should use temp file)


sionking:
under util.py delete_line_from_file is not safe to use.
Please change it to the following:

````
def delete_line_from_file(filepath, lineToDelete, logger):
    try:
        filepathOld = filepath+".old"
        filepathTemp = filepath+".temp"

        f = open(filepath, "r")
        lines = f.readlines()
        f.close()

        f = open(filepathTemp, "w")
        for line in lines:
            if line != lineToDelete:
                f.write(line)
            else:
                print("123")
        f.close()

        # File leftovers that should not exist, but if so remove it
        try:
            os.remove(filepathOld)
        except:
            pass
        # rename original file to _old
        os.rename(filepath, filepathOld)
        # rename new temp file to filepath
        os.rename(filepathTemp, filepath)
       # remove old and temp file
        os.remove(filepathOld)

    except BaseException as e:
        logger.error("delete_line_from_file error {}".format(str(e)))
````
*** Something improtant ***
Related to the file "userX_followedPool.csv", it contain many usernames without ',' at the end of the line before the \n.
Maybe an old syntex leftovers? it results in not removing the username from list after unfollowing it !!!!

sionking:
rewriting the file can be stopped in the middle - > results in a file with missing users.

PR #1329, opened 2018-02-01
Validate action; Fix unsafe delete_line_from_file ;

sionking:
Like click validation is done to check not been blocked, if been blocked we sleep for 2 minute but we can also handle it differently in the future.

Unfollow validation is done to not remove a username from list since been unfollowed but Instagram blocked us from unfollwing so the user is still followed but removed from the list.

Fix unsafe delete_line_from_file #992


sionking:
also a missing logfolder:
-def like_image(browser, username, blacklist, logger):
 +def like_image(browser, username, blacklist, logger, logfolder):

CharlesCCC:
If the like is blocked, is that two minutes later, the block will still exist?  Sleep for two minute won't help to remove the block, right ?

sionking:
It is better than liking 100 times an hour like it happens now, and get no like this seems much stranger.
of course we need to handle it somehow.

sionking:
Why not merged ? It's working very good.

Naramsim:
Right, this PR broke the script

jeremycjang:
Hi @sionking , could you please clarify how this new code should function? I have the most recent pull of the master, but currently the program doesn't remove any users from my `FollowedPool.csv` at all, nor do I see any old or temp files created in the logs folder. The new `delete_line_from_file` is never triggered when the `unfollow_users` is called upon and  the `removed from csv` message is never displayed.

Issue #1108, opened 2017-12-19
 TypeError: not all arguments converted during string formatting


TR0X:
I've chcked the output of InstaPy and found an error message in there: 
```
WARNING - Post Randomly Skipped...

Traceback (most recent call last):
  File "/usr/lib/python2.7/logging/__init__.py", line 859, in emit
    msg = self.format(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 732, in format
    return fmt.format(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 471, in format
    record.message = record.getMessage()
  File "/usr/lib/python2.7/logging/__init__.py", line 335, in getMessage
    msg = msg % self.args
TypeError: not all arguments converted during string formatting
Logged from file instapy.py, line 1314
Traceback (most recent call last):
  File "/usr/lib/python2.7/logging/__init__.py", line 859, in emit
    msg = self.format(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 732, in format
    return fmt.format(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 471, in format
    record.message = record.getMessage()
  File "/usr/lib/python2.7/logging/__init__.py", line 335, in getMessage
    msg = msg % self.args
TypeError: not all arguments converted during string formatting
Logged from file instapy.py, line 1314
```

converge:
hey @TR0X , what's your config (quickstart.py) ?

TR0X:
Attached:

```
insta_username = '**************'
insta_password = '**************'
session = InstaPy(username=insta_username, password=insta_password, use_firefox=True, nogui=True)
session.login()
session.set_do_comment(True, percentage = 25)
session.set_comments(['Beautiful ! :)', 'Pretty! :D', 'Great photo'])
session.like_by_locations(['262309028/austria', '63134088/vienna-austria', '213933962/zermatt-switzerland'], amount=50)
session.follow_user_followers(['sonyalpha', 'kai.boet', 'menschfotograf'], amount=10, randomize=True)
session.unfollow_users(amount=10, onlyInstapyFollowed = True, onlyInstapyMethod = 'FIFO')
session.set_user_interact(amount=3, randomize=True, percentage=50, media='Photo')
session.set_do_like(enabled=True, percentage=75)
session.set_comments(["Great!", "Nice one!"])
session.set_do_comment(enabled=True, percentage=10)
[...]

PR #1343, opened 2018-02-02
Fix string formatting, issue #1108

christian0101:
TypeError: not all arguments converted during string formatting


converge:
hi @christian0101 , we do use PEP8 code style, your PR will break it.

christian0101:
Fixed in the new PR.

https://github.com/timgrossmann/InstaPy/pull/1346

PR #1355, opened 2018-02-04
Added global variables

lucanello:
I added global variables to instapy.py and a return statement to print_log_writer.py so that the number of followers, liked images, followed users, unfollowed users, already liked images, inappropriate images and commented images can be stored in the InstaPy-object and read from quickstart.py.

This also refers to: timgrossmann#1342

### DIFF
--- a/instapy/__init__.py
+++ b/instapy/__init__.py
@@ -1,1 +1,2 @@
+from .settings import Settings
 from .instapy import InstaPy
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -84,7 +84,7 @@ class InstaPy:
             os.makedirs(self.logfolder)
 
         self.page_delay = page_delay
-        self.switch_language = False
+        self.switch_language = True
         self.use_firefox = use_firefox
         self.firefox_profile_path = None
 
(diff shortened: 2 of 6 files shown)


lucanello:
I've added some further changes in order to give the possibility to locate your quickstart.py outside the "InstaPy" directory. Depending on your Python version, there is a directory in `/usr/lib/pythonX.X/` where you may store the "InstaPy" folder. You'll then be able to import InstaPy into a Python file located anywhere on the system. In this case you should set the
```
Settings.browser_location
```
and the
```
Settings.database_location
```
in your quickstart.py to match the new paths. Usually like this:
```
Settings.browser_location = "/usr/lib/pythonX.X/InstaPy/assets/chromedriver"
Settings.database_location = "/usr/lib/pythonX.X/InstaPy/db/instapy.db"
```
If you didn't move the InstaPy library to this folder, you should leave the two lines in quickstart.py commented out, though. Then the default configuration (database in `./db/instapy.db` and browser in `./assets/chromedriver`) will be applied.

Using my changes you'll be able to provide a pip installer package.

timgrossmann:
@lucanello This is awesome! 
Thank you very much for your contribution.

Considering all the up votes I can assume that this works for different platforms as well?

uluQulu:
I haven't thought it would be this easy, now having **specified** the database's location by the usernames in `quickstart.py`, it did solve the behaviour of multi-user capability of my Quota Supervisor, thanks @lucanello

PR #1364, opened 2018-02-06
update instapy to disable image load in chrome webdriver

ganeshkrishnan1:
Disable image loading on chrome. This speeds up the process and we don't really need images for automation

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -216,7 +216,8 @@ class InstaPy:
                                             .format(user_agent=user_agent))
 
             chrome_prefs = {
-                'intl.accept_languages': 'en-US'
+                'intl.accept_languages': 'en-US',
+                "profile.managed_default_content_settings.images":2
             }
             chrome_options.add_experimental_option('prefs', chrome_prefs)
             self.browser = webdriver.Chrome(chromedriver_location,
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -209,14 +209,14 @@ class InstaPy:
             # GUI-less browser. chromedriver 2.9 and above required
             if self.headless_browser:
                 chrome_options.add_argument('--headless')
+                chrome_options.add_argument('--blink-settings=imagesEnabled=false')
                 # Replaces browser User Agent from "HeadlessChrome".
                 user_agent = "Chrome"
                 chrome_options.add_argument('user-agent={user_agent}'
                                             .format(user_agent=user_agent))
 
             chrome_prefs = {
-                'intl.accept_languages': 'en-US',
-                "profile.managed_default_content_settings.images":2
+                'intl.accept_languages': 'en-US'
             }
             chrome_options.add_experimental_option('prefs', chrome_prefs)
             self.browser = webdriver.Chrome(chromedriver_location,
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -215,7 +215,8 @@ class InstaPy:
                                             .format(user_agent=user_agent))
 
             chrome_prefs = {
-                'intl.accept_languages': 'en-US'
+                'intl.accept_languages': 'en-US',
+                "profile.managed_default_content_settings.images":2
             }
             chrome_options.add_experimental_option('prefs', chrome_prefs)
             self.browser = webdriver.Chrome(chromedriver_location,


uluQulu:
hi @ganeshkrishnan1 
thanks for bringing this amazing idea into attention!

it would be really cool to speed up the process by skipping the phase of **downloading** pictures
especially in `headless` mode where we don't really need to see pics visually imho...

but I'm afraid **preferences** are not currently supported in **headless** Chrome ([it's a chromedriver issue](https://bugs.chromium.org/p/chromedriver/issues/detail?id=1925))
instead, to disable images we can still use the `blink-settings` **command line flag** for it,
```python
if self.headless_browser:
    chrome_options.add_argument('--headless')
```
becomes
```python
if self.headless_browser:
    chrome_options.add_argument('--headless')
    chrome_options.add_argument('--blink-settings=imagesEnabled=false')
```
else than headless mode, i don't think it's a good idea..

CharlesCCC:
@uluQulu this is great idea. One thing I would like to point out is, after get the `DONT-Download-Image` flag set, this might have negative affect for those folks who is using the Clarifai image API, if the API call was send through browser image.   If this get merged in, make sure you add the corresponding `READ.me`

uluQulu:
hi @CharlesCCC 
thanks for clarifying, but we really don't need to download pictures for `clarifa`ing :-)
Clarifai API _only_ needs the **image url** which is sent _without downloading the picture data to your local/remote temporary storage..._
in clarifai_util.py,
```python
img_link = get_imagelink(browser)   #simply finds img's url with `find_element_by_xpath` in `get_imagelink()` function
# Uses Clarifai's v2 API
model = clarifai_api.models.get('general-v1.3')   
image = ClImage(url=img_link)   #sends img url for clarifying
result = model.predict([image])   #returns us clarified results
```
it's fully local, by the way I am already using this functionality in my headless browser mode and it is amazing (clarifai also enabled)

CharlesCCC:
@uluQulu thank you for the confirmation. 
you suggest change worked like a charm. I would suggest the originator @ganeshkrishnan1  of this pull request re-open the PR based off @uluQulu suggestion.

ganeshkrishnan1:
I have added the flag to the headless mode and removed the no image mode for GUI. I guess we can leave the images for GUI mode as a user might want to see what's going on; not exciting but helpful to see.

sionking:
What about Instagram <?> they will see it automation since no Image is downloaded....

uluQulu:
good point + @sionking 
I think "imagesEnabled" functionality is not generic and simply prevents the images from **displaying** but yet Google Chrome keeps downloading  (you would check it with **fiddler**) ...
but ganesh's method by preferences is really **not** downloading images
And in headless, I think, loading **images** for nothing really is non-**sense** whilst increases reload speed when disabled

CharlesCCC:
@uluQulu so, you are saying this ` "profile.managed_default_content_settings.images":2` will actually preventing the image from downloading but not for the latest change ? uhh... that is interesting. 

I guess the Chrome flag to don't display image is used for headless to further reduce memory/cpu usage then.

uluQulu:
@CharlesCCC 
I will try to verify it with `fiddler`, but @sionking is right😏, if it skips downloading and Instagram would notice that could not be safe but is the only option for headless.. :/
in any case, loading non-graphical pages will speed up the process a lot and save some bits of CPU 👍

CharlesCCC:
@sionking @uluQulu how could Instagram detect if you download image or not ?  
I can think of, the browser sends the request but detached the response for the image part ? The Instagram server can't tell if you download or not, right ?

uluQulu:
@CharlesCCC I can't be sure, really, I've a very basic knowledge of programming and esp web&html...you should know better about it, but to say, my concern is that, if we generate activity without communicating with their media **download**/upload servers, the host connection will look spammy...

ganeshkrishnan1:
The images are downloaded from a CDN. For example here is an image : https://scontent-syd2-1.cdninstagram.com/vp/6a3308d2dd7fbe2cff4a70ca338027c2/5B23A12E/t51.2885-15/s1080x1080/e35/27579878_193258524742430_6043553257928785920_n.jpg

It would be crazy for instagram to check if a user is downloading the images are not (possible but highly improbable and defeats the purpose of CDN). I seriously doubt if a CDN is even capable of knowing that a particular user didn't download the image. 

It's safe to assume that we can skip downloading images and there won't be consequences. I have this running on my rpi 3.

Also per my original pull its much better to skip downloading than disable. I reduced my run time from 4-6 hours to less than 2 hours (I am in Australia with very shitty internet)

CharlesCCC:
@ganeshkrishnan1 Awesome, thank you for verify. 
Based off the research @uluQulu did, it sounds like the `headless` and `don't-download image` can't be enabled at the same time due to whatever the bug chromedrive have. (not sure if it been fixed on the newer version or not).   So, you setup is `don't download iamge` with `headless` mode off ?

ganeshkrishnan1:
Yes I have both headless and image 'don't download' set. 

```
                      password=insta_password,
                      headless_browser=True,
                      multi_logs=True)
```

```
           chrome_prefs = {
                'intl.accept_languages': 'en-US',
                "profile.managed_default_content_settings.images":2
            }
```

Have the same code running on Ubuntu and rpi3

CharlesCCC:
@ganeshkrishnan1 surprising and awesome. 
 I thought it will not work together based off @uluQulu's research. 

If this was the case, I would suggest you still provide both flag in your pull request.
so,

have this 
```
if self.headless_browser:
    chrome_options.add_argument('--headless')
    chrome_options.add_argument('--blink-settings=imagesEnabled=false')
```

and 

```
  chrome_prefs = {
                'intl.accept_languages': 'en-US',
                "profile.managed_default_content_settings.images":2
            }
```

If people are using headless mode, they would careless of not download the picture.

ganeshkrishnan1:
done. Both headless and gui wont have images.

CharlesCCC (approved):
test this locally with two of my account. 
Windows 10 with Chrome 
Works perfect fine.

timgrossmann:
@ganeshkrishnan1 @CharlesCCC I think we had something similar to this before and I made some problems, so we need to either make sure that this does not lead to any kind of problems or make it easy for the user to deactivate (e.g. parameter)

riccardorighettiphoto:
Hello,
Really great feature. I needed something to reduce bandwidth and this is the trick.
I have just tested both code modifications: `headless_browser=True`along with those suggested by you guys in chrome prefs.
It looks like working as a silk.

Thank you.

CharlesCCC:
@timgrossmann  good point,
maybe add another flag to 
```
                      password=insta_password,
                      headless_browser=True,
                      multi_logs=True,
                     dontDownloadImage=True)
```

uluQulu:
@CharlesCCC 
Actually, it is a feature rather than  a **bug**
From chromium developer: "_The content settings are a chrome/ layer feature. Getting this to work in headless mode would be a large project, which isn't currently planned.  Sorry._" (Jan 28)

just like the ganesh's **preference**, any other preference in Chrome browser is not available for headless mode and **including** it would be a moderately large task, I do not think it will be available any soon  . . .

 When I saw ganesh's post I was amazed, but when I realized that instagram may notice seamless activity I prefered the disabling method for favor of some bits of speed, if some says it really **optimally** impossible for Instagram to detect us not-downloading, I would _enjoy_ non-download method **as a parameter**  (sometimes, I use non-headless and really don't need to see visuals) for non-headless mode..

Also, Charles, I'm sorry😟, but I did not say sth like that, actually, if you provide **both**, preferences and flag, headless mode will **skip** **any** given preference and use available flags (if some)...
_I added it after headless cos I did not see the point of disabling images in non-headless mode,,_

uluQulu:
**Finally**, I did test both options through **fiddler** web debugger:
**Preference method**: no images downloaded, except website icons (**1** Google and **3** Instagram **.ico** files, not a big deal..)
**Command-line argument**: the same as previous...

> Sorry guys, as a matter of fact **_I thought_** `imagesEnabled` flag is like _old_ `--disable-images` flag which was downloading but not displaying (deprecated now...)

So, **to say**, _both_ of the options are **not downloading** content **type of** image whilst **command**-line argument is working **for** both headless and non-headless sessions.

_For now, I will not use it until somebody says it **really** is safe (**btw** don't try to disable images with old `--disable-images` flag, it really is deprecated, doesn't work anymore )_

@CharlesCCC @ganeshkrishnan1 I should have tested it before :-)

uluQulu:
Also, I have to say,
```python
firefox_profile = webdriver.FirefoxProfile()
# 1 - Allow all images
# 2 - Block all images
# 3 - Block 3rd party images 
firefox_profile.set_preference("permissions.default.image", 2)
```
[stops](http://kb.mozillazine.org/Permissions.default.image) images from being downloaded on webpage in Firefox and **I see** that we **already** use this #2 option, already 😅 in `set_selenium_local_session()` method and it did not yet brought up any issue for firefox users (or it did?)

To mention, they say,
> HTTP(S) sends request for A PAGE, which is just HTML text; it's up to the browser to decide what to do with image URLS

above is true, web buddies?

CharlesCCC:
@uluQulu thank you for look that up. Based off the limit web knowledge I know, yes, just like the clarifia API that take image URL, i believe the browser will take the URL from the webserver reponse and then decide what to do with it, if the user says don't download it. then the web browser will simply disregard the URL and don't download.  

Which in terms of Instagram.com, they have no idea what did you do with the html reponse on your browser side. it all depend on the browser and the user. So to say, it is very safe to enable this feature.

keusta:
IG "can" certainly record anything with the great help of javascript, something like "is all images loaded".. but do they ? maybe for stats

sionking:
I work in a data link level of 7 layers (wifi eng).
There is nothing easier than to know if user download an image or not.
Also, Instagram can know, I believe, if you loaded the images or not.
I reach 2000 likes a day easily, what speed/BW do you want to save ?

If you handle real users, keep it real!

CharlesCCC:
@sionking not everybody have 100Mpbs internet connection across the global. some people might still stuck with 512Kbps. so any opiton like this, i believe it should be merged in. If you talk about real, everybody should just click 2000 likes each day, instead of running this script.   🥇

timgrossmann:
Maybe we should do a little quick poll here... 
@sionking I don't know the implications when it comes to Instagram checking for this kind of stuff and blocking users earlier... However I'm also the same opinion @CharlesCCC has about people having slower internet connections and therefore loading speed...

Let's just do it this way, React with 
:+1: if you think we should merge this
:-1: if you think we shouldn't

This is a huge community by now, so I'm not going to make decisions without people giving their opinions before that.

uluQulu:
It is so real, I run through 1 Mbps (_shared to three ppl_) and have not had speed problems, yet and I also think **over-speed**ing is not so good, too, which will generate inordinately ostentatious activity

I said above, we are already using it for firefox (since last couple of days, I have been using it in headless Chrome, too, without problems)
my opinion is to offer it as a parameter in quickstart and **inform** user about the effect & possible consequences :-)
```python
session = InstaPy (DL_images= ['regular', 'headless'])
```
> Yes, Instagram can find a way to record it, but I think they don't do it
reason: having served to **hunders** of millions of ppl, they will try to use the minimal resources and detect&record&inpect only what is important and since (I guess) **very less** ppl are not downloading pictures, they do not care it (yeah, if they are aware of it...))

sionking:
What r u talking about ? Instagram fighting bots like hell.
They buy and use very sophisticated cyber companies, what is resources for those multi billion dollars companies (i.e Facebook, Instagram)???

Most of my real accounts never got blocked (not even likes block) since I keep it real, I run each account from it owner home (ip and geo are real), the wifi is also used for the account owner mobile so it seems real.

I know companies that use different applications for bots and get blocked and even removed after couple of months. 

BTW, CAN YOU MAYBE TRY DL IMAGES IN **LOW QUALITY** INSTEAD ?

**Please if you decide to add no-image load / no-image DL please make it optional**

uluQulu:
@sionking 
There are lots of primitive commercial bots (~7$ per month) and some professional automation services (~100$/m) and none of them have a feature to turn off downloading images I heard of, minority doesn't count up
Keep in mind that, these companies are not againist bots (well they are, in papers), they are againist **excessive** automation
It will be a parameter if dont wanna use, keep it `None` or dont include at all
Also, average pic in IG is below 100 KB, low quality?

sionking:
@uluQulu Instagram pictures are "HD" mean 1080X1080, I think you can ask for images 640X640 instead, maybe in settings.

"Keep in mind that, these companies are not against bots (well they are, in papers), they are againist excessive automation"
NOT TOTALLY TRUE, they are really against **fake bots** "spammers" and not real account uses automation, since those real accounts generate good traffic for those companies "active users".

sionking:
Anyway I don't want to see one of my 5 years old costumers accounts getting account blocked, I take safety at top. I even run over all links to check for "bed links" ("Page Not Found" in browser.title or "Content Unavailable" in browser.title) and other stuff before I serve it to the account bot.

uluQulu:
@sionking 
 There are tons of users using automation, if they were to hard ban all of them once (or twice or more than twice) detected, they would lose tons of users and couldn't grow, all they want is no problems, if you exceed limits continiously and not stop you will most prob end up being banned, even though you do not use automation.
 Yes, IG compresses and stores pics in max 1.91x1 (1080x566 px) landscape and 1x1.25 (1080x1350 px) portrait ratios according to your image but higher or lower DPI requests in smaller displays doesn't really affect quality
 Since you have high-speed bandwith, you should not use it, for extra safety
But I will use and I think, as of today, it is definitely safe!

timgrossmann:
Even though it adds even more complexity, could we add this as a parameter? 
I'd really like for people to be able to set this themselves easily...

Maybe we should think about moving all the `InstaPy(...)` parameters to some `.set_....` method

CharlesCCC:
FYI, I have manually made the change from the date this PR create. 
I haven't had issue with the Instagram at all. the speed is increased and the internet usage reduced quite a lots (not that matter too much).    wonder if @ganeshkrishnan1 still have the time to make the suggest change ?

sionking:
JFYTK,
14 hours of run is 1.4GB

CharlesCCC:
@sionking is the with or w/o the image disabled=on ?

Syslog777:
Um if the image load is disabled Instagram may detect that it is a bot and block the user

Syslog777:
@CharlesCCC

CharlesCCC:
@Syslog777 well, we had the discussion before. I have used this feature for probably more than 3-4 months now. without issue (two accounts).  Instagram.com haven't done anything yet, but I would say, Instagram.com probably knows everything, it just matter if they decide to do anything with it or not.

sionking:
@CharlesCCC  I think **with** downloading images I get 1 GB a day / account.

fgisslen:
Hi,
i use nogui=True, headless_browser=False because i use proxy authentication. 
How do I setup instapy not to download the images? 

Edit: I added chrome_options.add_argument('--blink-settings=imagesEnabled=false') outside headless_browser so it will work with nogui.

I hope that works?

uluQulu:
Hey @fgisslen 
While using InstaPy, all of the interactions (_literally_) are being happened through a web browser and it is being controlled by its web driver.
The point is, you should not worry bout any other software rather than web browsers, whether it is a _no-GUI_ or a _HEADless_ mode.  

For Firefox, InstaPy will not download pictures by default.  
But if you are using a Chrome browser, then that code
```erlang
chrome_options.add_argument('--blink-settings=imagesEnabled=False')
```
will do it.

>If you have read the discussion in this thread you will find many helpful matters on your interest.  

------

Cheers 😁

sionking:
Do any one have issues with blocking and image dl ? or is it safe ?

sionking:
Isn't there is just "partial" image dl lets say 10% ?

uluQulu:
@sionking 
Well, I use it for a long time and I don't know if there were any problems 😂 but I had excellent results during that time 😋

IDK about low quality images to be offered or IF it is available for use in Chrome, e.g. Opera has it for a long time.
And it depends on the server too, IG must offer low quality images and earlier I thought about it is just DPI but after some use I see that it also serves pics in different quality [PARTIALLY or FULLY].
#### Page size also changes the pic quality [AT LEAST FOR SOME SORT OF PICTURES e.g. profile pics that I tested]

-----

Cheers 😁
------

timgrossmann:
@sionking @uluQulu I still think it should be optional.
Partially loading an image is probably nothing you can decide via selenium... 

I think the contributor is out of this discussion... We probably have to apply his changes again to the updated version of InstaPy

timgrossmann:
Updated version #2938

CharlesCCC:
@sionking I have been keeping using it for more than 6 months now. it been fine and safe, no issue noticed.

fgisslen:
I have also used this for a couple of months, no problems

PR #1390, opened 2018-02-09
Add utf-8 to reason not liked;Avoid divide by zero;Add missing functions: Capture all followed and unfollowed ever

sionking:
Good when we handle real users and facing any bug so we can't unfollow anymore since the pool is empty.

### DIFF
--- a/instapy/print_log_writer.py
+++ b/instapy/print_log_writer.py
@@ -36,23 +36,23 @@ def log_uncertain_unfollowed_pool(login, unfollowed, logger, logfolder):
         with open('{0}{1}_uncertain_unfollowedPool.csv'.format(logfolder, login), 'a+') as followPool:
             followPool.write(unfollowed + ",\n")
     except BaseException as e:
-        logger.error("log_followed_pool error {}".format(str(e)))
+        logger.error("log_uncertain_unfollowed_pool error {}".format(str(e)))
 
 
 def log_record_all_unfollowed(login, unfollowed, logger, logfolder):
-    """Prints and logs the uncertain unfollowed to
+    """logs all unfollowed ever to
     a seperate file"""
     try:
         with open('{0}{1}_record_all_unfollowed.csv'.format(logfolder, login), 'a+') as followPool:
             followPool.write(unfollowed + ",\n")
     except BaseException as e:
-        logger.error("log_followed_pool error {}".format(str(e)))
+        logger.error("log_record_all_unfollowed_pool error {}".format(str(e)))
 
 
 def log_record_all_followed(login, followed, logger, logfolder):
-    """We save all followed to a pool that will never be erase"""
+    """logs all followed ever to a pool that will never be erase"""
     try:
         with open('{0}{1}_record_all_followed.csv'.format(logfolder, login), 'a+') as followPool:
             followPool.write(followed + ",\n")
     except BaseException as e:
-        logger.error("log_followed_pool error {}".format(str(e)))
+        logger.error("log__record_all_followed_pool error {}".format(str(e)))
--- a/instapy/print_log_writer.py
+++ b/instapy/print_log_writer.py
@@ -24,3 +24,35 @@ def log_followed_pool(login, followed, logger, logfolder):
             followPool.write(followed + ",\n")
     except BaseException as e:
         logger.error("log_followed_pool error {}".format(str(e)))
+
+    # We save all followed to a pool that will never be erase
+    log_record_all_followed(login, followed, logger, logfolder)
+
+
+def log_uncertain_unfollowed_pool(login, unfollowed, logger, logfolder):
+    """Prints and logs the uncertain unfollowed to
+    a seperate file"""
+    try:
+        with open('{0}{1}_uncertain_unfollowedPool.csv'.format(logfolder, login), 'a+') as followPool:
+            followPool.write(unfollowed + ",\n")
+    except BaseException as e:
+        logger.error("log_followed_pool error {}".format(str(e)))
+
+
+def log_record_all_unfollowed(login, unfollowed, logger, logfolder):
+    """Prints and logs the uncertain unfollowed to
+    a seperate file"""
+    try:
+        with open('{0}{1}_record_all_unfollowed.csv'.format(logfolder, login), 'a+') as followPool:
+            followPool.write(unfollowed + ",\n")
+    except BaseException as e:
+        logger.error("log_followed_pool error {}".format(str(e)))
+
+
+def log_record_all_followed(login, followed, logger, logfolder):
+    """We save all followed to a pool that will never be erase"""
+    try:
+        with open('{0}{1}_record_all_followed.csv'.format(logfolder, login), 'a+') as followPool:
+            followPool.write(followed + ",\n")
+    except BaseException as e:
+        logger.error("log_followed_pool error {}".format(str(e)))
--- a/instapy/print_log_writer.py
+++ b/instapy/print_log_writer.py
@@ -21,7 +21,7 @@ def log_followed_pool(login, followed, logger, logfolder):
     a seperate file"""
     try:
         with open('{0}{1}_followedPool.csv'.format(logfolder, login), 'a+') as followPool:
-            followPool.write(followed + ",\n")
+            followPool.write('{},\n'.format(followed))
     except BaseException as e:
         logger.error("log_followed_pool error {}".format(str(e)))
 
@@ -34,7 +34,7 @@ def log_uncertain_unfollowed_pool(login, unfollowed, logger, logfolder):
     a seperate file"""
     try:
         with open('{0}{1}_uncertain_unfollowedPool.csv'.format(logfolder, login), 'a+') as followPool:
-            followPool.write(unfollowed + ",\n")
+            followPool.write('{},\n'.format(unfollowed))
     except BaseException as e:
         logger.error("log_uncertain_unfollowed_pool error {}".format(str(e)))
 
@@ -44,7 +44,7 @@ def log_record_all_unfollowed(login, unfollowed, logger, logfolder):
     a seperate file"""
     try:
         with open('{0}{1}_record_all_unfollowed.csv'.format(logfolder, login), 'a+') as followPool:
-            followPool.write(unfollowed + ",\n")
+            followPool.write('{},\n'.format(unfollowed))
     except BaseException as e:
         logger.error("log_record_all_unfollowed_pool error {}".format(str(e)))
 
@@ -53,6 +53,6 @@ def log_record_all_followed(login, followed, logger, logfolder):
     """logs all followed ever to a pool that will never be erase"""
     try:
         with open('{0}{1}_record_all_followed.csv'.format(logfolder, login), 'a+') as followPool:
-            followPool.write(followed + ",\n")
+            followPool.write('{},\n'.format(followed))
     except BaseException as e:
         logger.error("log__record_all_followed_pool error {}".format(str(e)))
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -618,7 +618,7 @@ class InstaPy:
                             already_liked += 1
                     else:
                         self.logger.info(
-                            '--> Image not liked: {}'.format(reason))
+                            '--> Image not liked: {}'.format(reason.encode('utf-8')))
                         inap_img += 1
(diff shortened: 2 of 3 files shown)


instapy/print_log_writer.py
```diff
@@ -24,3 +24,35 @@ def log_followed_pool(login, followed, logger, logfolder):
...
+    # We save all followed to a pool that will never be erase
+    log_record_all_followed(login, followed, logger, logfolder)
+
+
+def log_uncertain_unfollowed_pool(login, unfollowed, logger, logfolder):
+    """Prints and logs the uncertain unfollowed to
+    a seperate file"""
+    try:
+        with open('{0}{1}_uncertain_unfollowedPool.csv'.format(logfolder, login), 'a+') as followPool:
+            followPool.write(unfollowed + ",\n")
```
timgrossmann: Could we use .format() here aswell, please?

instapy/print_log_writer.py
```diff
@@ -24,3 +24,35 @@ def log_followed_pool(login, followed, logger, logfolder):
...
+    except BaseException as e:
+        logger.error("log_followed_pool error {}".format(str(e)))
+
+
+def log_record_all_unfollowed(login, unfollowed, logger, logfolder):
+    """Prints and logs the uncertain unfollowed to
+    a seperate file"""
+    try:
+        with open('{0}{1}_record_all_unfollowed.csv'.format(logfolder, login), 'a+') as followPool:
+            followPool.write(unfollowed + ",\n")
```
timgrossmann: Could we use .format() here aswell, please?

instapy/print_log_writer.py
```diff
@@ -24,3 +24,35 @@ def log_followed_pool(login, followed, logger, logfolder):
...
+    except BaseException as e:
+        logger.error("log_followed_pool error {}".format(str(e)))
+
+
+def log_record_all_followed(login, followed, logger, logfolder):
+    """We save all followed to a pool that will never be erase"""
+    try:
+        with open('{0}{1}_record_all_followed.csv'.format(logfolder, login), 'a+') as followPool:
+            followPool.write(followed + ",\n")
```
timgrossmann: Could we use .format() here aswell, please?

timgrossmann (approved):

instapy/print_log_writer.py
sionking: there is a format().. I dont get it...

CharlesCCC:
I'm getting the following error with the change you implemented (avoid divide by zero) : 

```
  File "quickstart-bypass.py", line 80, in <module>
    main(sys.argv[1:])
  File "quickstart-bypass.py", line 73, in main
    session.like_by_tags(amount=int(likeamount), use_smart_hashtags=True)
  File "C:\Users\Charles\OneDrive\InstaPyProd\instapy\instapy.py", line 678, in like_by_tags
    for i, link in enumerate(links):
TypeError: 'bool' object is not iterable
```

sionking:
@CharlesCCC 
Hi I don't get how you get from like_by_tags  to get_links_for_username
And we always check:
            if links is False:
                continue
So please tell me what you see...

CharlesCCC:
@sionking I'm not sure buddy, I first changed my local file to the one you made in `like_util.py`
and I'm getting the error I posted, then later on I have to change the code like the following, it seems worked so far:(temp. solution)
```
    link_elems = main_elem.find_elements_by_tag_name('a')
    total_links = len(link_elems)
    if total_links == 0:
        total_links = 1;
    links = []
```

sionking:
I am not sure I hate this github merging please check you merged right and not just switched files.

There are only 2 places with this call and both check for 'False':
```
            try:
                links = get_links_for_username(
                    self.browser,
                    username,
                    amount,
                    self.logger,
                    randomize,
                    media)
                self.logger.error('Element not found, skipping this username')
                continue

            if (self.do_follow and
                username not in self.dont_include and
                following and
                    self.follow_restrict.get(username, 0) < self.follow_times):
                followed += follow_user(self.browser,
                                        self.follow_restrict,
                                        self.username,
                                        username,
                                        self.blacklist,
                                        self.logger,
                                        self.logfolder)
            else:
                self.logger.info('--> Not following')
                sleep(1)

            if links is False:
                continue
```

```
            try:
                links = get_links_for_username(self.browser,
                                               username,
                                               amount,
                                               self.logger,
                                               randomize,
                                               media)
                self.logger.error('Element not found, skipping this username')
                continue

            if links is False:
                continue
```

...
Can someone try?

sionking:
Now I get it I have this check but latest version don't I will PR

uluQulu:
In `like_image()` function of `like_util.py`
lines from 561 to 569, after latest commit of @sionking,
should be replaced with
```python
        elif len(liked_elem) == 1:
            logger.info('--> Image already liked! ')
            return False
        
        else:
            logger.info('--> Invalid Like Element!')
            return False
```
i noticed while merging, that, without this 👆 it will not notify correctly once encountered an **invalid like element**...

sionking:
@uluQulu hi
```
def like_image(browser, username, blacklist, logger, logfolder):
    """Likes the browser opened image"""
    like_elem = browser.find_elements_by_xpath(
        "//a[@role='button']/span[text()='Like']/..")
    if len(like_elem) == 1:
        # sleep real quick right before clicking the element
        sleep(2)
        click_element(browser, like_elem[0])
        # check now we have unlike instead of like
        liked_elem = browser.find_elements_by_xpath(
            "//a[@role='button']/span[text()='Unlike']")
        if len(liked_elem) == 1:
            logger.info('--> Image Liked!')
            update_activity('likes')
            if blacklist['enabled'] is True:
                action = 'liked'
                add_user_to_blacklist(
                    browser, username, blacklist['campaign'], action, logger, logfolder
                )
            sleep(2)
            return True
        else:
            # if like not seceded wait for 2 min
            logger.info('--> Image was not able to get Liked! maybe blocked ?')
            sleep(120)
    else:
        liked_elem = browser.find_elements_by_xpath(
            "//a[@role='button']/span[text()='Unlike']")
        if len(liked_elem) == 1:
            logger.info('--> Image already liked! ')
            return False

    logger.info('--> Invalid Like Element!')
    return False
```

When invalid it returns False, So I don't see your point ...

uluQulu:
@sionking 
Yea you're right, checked again, that's a crazy design xD

sionking:
Hi, so is it working for you guys ?

vladislav-vasilenko:
from .print_log_writer import log_uncertain_unfollowed_pool
ImportError: cannot import name 'log_uncertain_unfollowed_pool'

How to solve this?

sionking:
@applehawk go and merge this PR
OR 
comment few lines

timgrossmann:
@sionking Could you please change the mentioned lines to use `.format(...)`?

sionking:
@timgrossmann Sorry I don't get it... It all uses .format()

timgrossmann:
@sionking I commented the lines in your code:

```bash
with open('{0}{1}_uncertain_unfollowedPool.csv'.format(logfolder, login), 'a+') as followPool:
            followPool.write(unfollowed + ",\n")
```

sionking:
@timgrossmann
I see the line you mentined but I don't understand what you want me to change ?
I use the .format in all lines.

timgrossmann:
@sionking No you don't 

you use
```
followPool.write(unfollowed + ",\n")
```

to write to the file instead of something like

```
followPool.write('{},\n'.format(unfollowed))
```

sionking:
Hi, 
@timgrossmann I now understand I will change it now.
But JFYTK it was like that in function log_followed_pool:
```
def log_followed_pool(login, followed, logger, logfolder):
    """Prints and logs the followed to
    a seperate file"""
    try:
        with open('{0}{1}_followedPool.csv'.format(logfolder, login), 'a+') as followPool:
            followPool.write(followed + ",\n")
    except BaseException as e:
        logger.error("log_followed_pool error {}".format(str(e)))
```

So I left it in the same format

sionking:
Done

timgrossmann:
@sionking Awesome!

Could we please get 2 or at least 1 more approval here?

uluQulu:
encoding `reason` to UTF-8 will solve possible exceptions once inappropriate word contains a non-unicode char (e.g. contains `ö`), I did this by just encoding `inapp_unit` in `check_link` but this is a more general solution for future releases.
Everything works **fine**, but I wonder one thing, @sionking 
what are those newly written three files for?
`_record_allfollowed`, `_record_allunfollowed`, `_uncertain_unfollowedPool`

do we **really** need those?

sionking:
@uluQulu hi
look at:
https://github.com/timgrossmann/InstaPy/issues/1420

This just saved me.
I could have take the list from blackList or logs but it is easier that way.
We can add a flag for it...

Why more files is an issue ? size?

uluQulu:
Right, stability is much important
for me, I  just started to use `multi_logs=True` parameter to have a clean **logs** folder :-)

CharlesCCC:
@uluQulu agree with you. 
We should keep the flow simple and clean to achieve whatever function we would like to achieve. "more not always be good"   👍

uluQulu (approved):

sionking:
After 3 different real accounts found with incorrect following pool I found the only solution is to log all activity, so I can manually clean those account.

If you prepare I can remove all logs but debug will be much harder/ or we can save all pools to a pool folder ....

sionking:
@timgrossmann can you decide ?

instapy/print_log_writer.py
estebancortero: I suppose, he refers to inside the `write`

timgrossmann:
@sionking Let's merge it for now... I hope at one point we can start a big refactor that makes things more clean.
We then should keep discussions like this open for longer.

What I'm wondering about at the moment however is if that is correct:

```python
with open('{0}{1}_record_all_followed.csv'.format(logfolder, login), 'a+') as followPool:
            followPool.write('{},\n'.format(followed))
```

Do you actually want the name of the csv file to be "<logfolder><login>"?
shouldn't the log folder be the path to a log folder?

Like this?
```python
with open('{0}/{1}_record_all_followed.csv'.format(logfolder, login), 'a+') as followPool:
            followPool.write('{},\n'.format(followed))
```

uluQulu:
self.logfolder handles it while `multi_logs==True`
```python
self.logfolder = './logs/'
if multi_logs is True:
    self.logfolder = './logs/{}/'.format(self.username)
```
I think it is better for user once logging is dealt by `multi_logs` arg to provide flexibility

sionking:
Sorry I don't use multi_logs, is mine correct ?

sionking:
BTW, the change of like image will save alot of time since looking for xpath that not exist is a 20 sec issue per image ! so Now it should be much faster for you all.

uluQulu:
if you don't use multi_logs your `logfolder` will be `self.logfolder = './logs/'`, so that every time you start a new username from the same instapy folder, will save logs into that `./logs/` folder instead of `./logs/myusername1/`
it is fine to not enable `multi_logs` but vice verse it is much more cleaner...

sionking:
@uluQulu I don't run on a server I have real account running from each one own laptop/PC so it uses home IP address (safer) plus no expenses for server/proxy for me.
All files are sync using file sharing cloud.

timgrossmann:
@sionking Alright, looks like everything is right then.
Let's merge this and make people happy!

Thank you @sionking for this contribution.

uluQulu:
@sionking 
i see, i also run locally except multiple accaunts from same instapy folder, that's why having multiple log folders worths using

jeremycjang:
Pulled these commits and it seems to have broken the unfollow program. I created an issue for it #1436 and will be reverting to the previous commit without these additional lists in the meantime.

timgrossmann:
@sionking just found a bug when using upper follower count:

```
--> Image not liked: b'None'
```

Edit: looks like it's not only when using upper follower count but a lot of others aswell

uluQulu:
@timgrossmann 
if im not mistaken, you're pointing out `b''` bytestring notation which appears in Python 3 consoles
it is very annoying for me, and i thing it is annoying for those who do data analysis from logs, too, cos logs also include that stamp **b'foobar'**
it's happening after encoding the raw output string to utf-8 bytes, since we can't quit utf-8 for reliability, we can remove it's `b''` stamp from output just by slicing encoded bytestring before sending the message to logger
i've posted a solution to this in #1413 (works for both Python 2 & 3)

Issue #1394, opened 2018-02-09
Self.logger, logfolder and unfollow log issues


Luisetty:
getting this error with unfollow and crash

## Current Behavior
ERROR [user]  unfollow loop error name 'log_uncertain_unfollowed_pool' is not defined
INFO [user]  --> Total people unfollowed : 0 

## Current Behavior
log:
raceback (most recent call last):
  File "quickstart.py", line 35, in <module>
    session.like_by_tags(['happy','sad'], amount=70, media='Photo')
  File "/Users/InstaPy/instapy/instapy.py", line 850, in like_by_tags
    self.logger)
TypeError: like_image() missing 1 required positional argument: 'logfolder'

any idea what i'm doing wrong or something is missing?

what is this thing wrong with self.logger? or how i can do a temporal fix? also with the unfollow error of loop error name

estebancortero:
Same problem here running on Raspberry Pi 3.

aellispierce:
I experienced the same issue and opened a pr above to fix ☝️. Until that gets merged you can fix it locally by adding `self.logfolder` after line 559 in `instapy/instapy.py`

It should look like: 
```ruby
                         liked = like_image(self.browser,
                                            user_name,
                                            self.blacklist,
                                            self.logger,
                                            self.logfolder)
```

Luisetty:
@aellispierce my instapy.py already has that and i'm getting the same issue.

looks like this my instapy.py file

            `if not inappropriate:
                        liked = like_image(self.browser,
                                           user_name,
                                           self.blacklist,
                                           self.logger,
                                           self.logfolder)`

I'm missing something?

felipewove:
also found a `self.logfolder` missing around line 845 in the same file

```python
                        liked = like_image(self.browser,
                                           user_name,
                                           self.blacklist,
                                           self.logger,
                                           self.logfolder)
```

@aellispierce could you update your PR, please? :)

felipewove:
Just founded out that this PR https://github.com/timgrossmann/InstaPy/pull/1329 added new arg to `like_image()` and broke everything...

To test, i checkout to commit before this merge `e68ac8d30ee254e5dd0f7df5e333d7bc879b4548` and it's working

sionking:
@felipewove the new arg should be there ...

sebmer-com:
to add those lines worked. @timgrossmann @sionking

sionking:
But you should also add the lines to blacklist function ... that is why it added in first place

PR #1395, opened 2018-02-09
Add required argument logfolder to like_image

aellispierce:
Fixes https://github.com/timgrossmann/InstaPy/issues/1394

This passes the new required logfolder argument when `like_image` is called


sionking:
thanks

sionking:
Hi, you added it only to one place ... It should be in any call.

aellispierce:
Updated to pass the argument in all instances when like_image is called

timgrossmann:
@aellispierce Thank you for contributing!

PR #1412, opened 2018-02-11
fix unfollow users that are following you

luigifaticoso:
screen shot 2018-02-11 at 14 51 23
Quick fix when something like this happens.
Only to some accounts, like mine.
When it happens it skips to the next step.


uluQulu:
Amazing @luigifaticoso 
update from future!

timgrossmann:
@luigifaticoso Awesome, thank you very much for this nice fix!

PR #1442, opened 2018-02-16
Update How_to_Raspberry.md: Minor simplification

estebancortero:
No need to add any extra repository to install chromium: https://raspberrypi.stackexchange.com/a/523/78381
Besides the referenced command gave me an error as it the repo wasn't signed correctly.


estebancortero:
Actually, installing chromium doesn't work.

PR #1454, opened 2018-02-18
Update amount of follow

dar2015:
Amount to follow for giving users follow split between each account.
For example if you have 10 different accounts you follow for followers intend of set for 20 each account you can set 200 for all and it will split it automatically.

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -4,7 +4,6 @@ import json
 import logging
 from math import ceil
 import os
-from platform import python_version
 from datetime import datetime
 from sys import maxsize
 import random
@@ -18,7 +17,6 @@ import requests
 
 if os.name != 'nt':
     from .clarifai_util import check_image
-from .settings import Settings
 from .comment_util import comment_image
 from .like_util import check_link
 from .like_util import get_links_for_tag
@@ -45,6 +43,7 @@ from .unfollow_util import load_follow_restriction
 from .unfollow_util import dump_follow_restriction
 from .unfollow_util import set_automated_followed_pool
 
+
 # Set a logger cache outside the InstaPy object to avoid re-instantiation issues
 loggers = {}
 
@@ -62,7 +61,6 @@ class InstaPy:
                  show_logs=True,
                  headless_browser=False,
                  proxy_address=None,
-                 proxy_chrome_extension=None,
                  proxy_port=0,
                  bypass_suspicious_attempt=False,
                  multi_logs=False):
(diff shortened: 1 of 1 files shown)


timgrossmann:
@dar2015 Thank you for contributing!

However, I don't really understand what this feature does, I'm sorry.
Could you explain it a little bit more?

dar2015:
@timgrossmann today, when you set amount to follow it's will replicate the amount for each users you want to follow from his list. for example you set amount 100 and you have 5 different users you follow from the bot will follow 500 users. 
My update will split 100 following between all the 5 users and each list will follow over 20 users.

dar2015:
Today
amount=100 
users=5
total follow = 500
5*100 = 500

My update
amount=100 
users=5
total follow = 100
100/5 = 20 * 5

timgrossmann:
@dar2015 Ok now I get it.
But this most likely will only make sense if we can make it optional.
Could you think about making this a parameter or setting, please?

dar2015:
I can do it, how you suggest to implement it?

sionking:
I think this is great!
I always trim my list to 5 tags (out of 20 let say) to not excceed with total tags amount.
So total is preferred .

GitNees:
Or put 20 with 5 users = 100 :D #quickmath

instapy/instapy.py
```diff
@@ -1221,12 +1221,15 @@ def follow_user_followers(self,
         userFollowed = []
         if not isinstance(usernames, list):
             usernames = [usernames]
-        for user in usernames:
 
+        amount_per_user = int(ceil(amount / len(usernames)))
```
Tjorriemorrie: please handle zerodivisionerror. Add return above if empty
Does not seem necessary to cast it to int (scroll_bottom casts it)

Tjorriemorrie (changes requested):

Tjorriemorrie:
I'm not sure I understand this change. Just call the method with 20.

dar2015:
Fixed for your request added as param

GitNees:
Still a very confusing option. Its creating a solution for a problem that doesn't exist.

sionking:
The problem:
The number of tags vary all the time, so if you have now 5 tags and amount 20 you will get **100**.
Tmrw you will add a new tag (6 tags) amount is the same and you will have the **120** instead.

Anyway I thought about it and still prefer to control the amount of tags I am entering.
A better feature will be amount per tag (use tuples). and total of 5 tags shuffled every run.

Tjorriemorrie:
That's still an implementation detail.

dar2015:
Idea of amount is to control how many follow do you want. I use external list of accounts that change for time to time and don't really know how many accounts on the list. the amount of follow changing for me now. this idea solve it.

timgrossmann:
I'm extremely unsure about this...
We either put this as a parameter (if we don't overdue it) or keep it off...

dar2015:
Now i able it as a parameter you can choose to by toggle it.

Tjorriemorrie:
But you also pass in the usernames: so you have all the info for the calculation which you can do just before you call the function; and pass in the calculated amount.

PR #1457, opened 2018-02-18
Added date and time to log

dar2015:
Added date and time to log

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -160,7 +160,7 @@ class InstaPy:
             file_handler = logging.FileHandler( '{}general.log'.format(self.logfolder))
             file_handler.setLevel(logging.DEBUG)
             extra = {"username": self.username}
-            logger_formatter = logging.Formatter('%(levelname)s [%(username)s]  %(message)s', datefmt='%Y-%m-%d %H:%M:%S')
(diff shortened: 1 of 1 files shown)


sionking:
I already added to mine

sionking:
We can than debug better

dar2015:
I didn't saw yours can you give a link to the commit?

uluQulu:
thanks @dar2015 
it is amazing now, champion commit!

althought, it would be significantly delightful for eyes if you could put datetime into round brackets instead of square brackets
```python
INFO (2018-02-18 18:42:34) [foobar] 
```
looks **cooler** (_and more readable_) **than**
```python
INFO [2018-02-18 18:42:34] [foobar] 
```

dar2015:
@uluQulu sure, but I would like the community to say what is the best wrapping for date.
What do you think round or square?

sionking:
I really dont care lets go with 
INFO [foobar] [2018-02-18 18:42:34] 

I prefer the data to be afeter the user...
I mean start with the things that less changed ... and end with the changing one

uluQulu:
@sionking 
when looked to log, `date` should have higher priority off `username`
_you can look for a certain date and then find data according to the username easily
but when the username has higher priority, it become an aggravating job_

> I meant in the multi-user perspective

@timgrossmann although will get a longer output with a _full datetime stamp_, this is literally required piece for the substantially effective **log**ging activity and I've been using it for the last days, works just smoothly 👌

PR #1459, opened 2018-02-18
Remove Requested follow

dar2015:
Remove Requested follow from users that didn't approve follow request. 
* Instagram limit 7,500 of following users. include following requests. means that if you have many unapproved requests it's can block you from follow new accounts. there is no way to go back and delete the requests. Instagram don't provide feature to see all outgoing requests therefore this feature is highly important.

### DIFF
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -96,10 +96,28 @@ def unfollow(browser,
                 if person not in dont_include:
                     browser.get('https://www.instagram.com/' + person)
                     sleep(2)
-                    follow_button = browser.find_element_by_xpath(
-                        "//*[contains(text(), 'Follow')]")
 
-                    if follow_button.text == 'Following':
+                    following = False
+                    try:
+                        follow_button = browser.find_element_by_xpath(
+                            "//*[contains(text(), 'Follow')]")
+                        if (follow_button.text == 'Following'):
+                            following = "Following"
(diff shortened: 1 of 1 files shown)


sionking:
Only if you try to follow private users ... ?

dar2015:
@sionking yes, absolutely.

CharlesCCC:
wow, this is very essential. good catch @dar2015

sionking:
Why not just saving a list (csv) of all requested and once in a while run over it and clean it ?

dar2015:
@sionking , actully it's can be great idea to add to csv of following request what is the following status (requested, followed) and what is the action or referral. for example from X followers.
It's give must information of following tracking.

timgrossmann:
@sionking @dar2015 Let's please not go back to having csv files for everything when we now start using a db.
If we persist it, then please let's save it in a separate table in the database.

@dar2015 Awesome catch and work btw. Thank you

CharlesCCC:
second on @timgrossmann 's point, with SQLite, we should have a more clear structured data to monitoring/analysis for the future.  It should also easier to transfer if we need to. since it all in one central place.

dar2015:
@timgrossmann,
Agree. let's merge it.
I have so many features in the pipeline and I don't want to get block by simple merge like that.

Next idea, option to use remote db or local db!

PR #1490, opened 2018-02-23
New Feature - Follows and Likes based on location

JoshMustafa:
Hi Guys,

I was wondering if I can get any help on implementing a new mode/feature where we can get the bot to follow users and like their post that is based in a specific location; like a town or country. This is essential for building an account with a specific user base.

Peace,


CharlesCCC:
Is this already a feature [Like by location](https://github.com/timgrossmann/InstaPy#like-by-locations)
(And you can also set do_follow users).

sionking:
When location is important better user popular profile from this location.

PR #1511, opened 2018-02-28
Fixed "bad" cookie

craigricker:
If the cookie associated with a login becomes "bad" for any reason, InstaPy previously did not have the ability to catch this. The most common reason for a cookie failing to work, is the user changing their password. 

login_util.py will check if a cookie is present, and if so, load it. With these changes, InstaPy will ensure the user is successfully logged in by checking if when loading www.instagram.com the user still sees a "Log in" button. If so, the cookie is assumed to be "bad" and re-created.

There are three states, and three associated test cases. These are listed below, and all three have passed successfully
1. Naive user  - no cookie has been created for this account
2. "Good" cookie - user has been successfully logged in on the active machine in the past, and loading the associated cookie successfully logs in the user.
2. "Bad" cookie - there is a cookie associated with the user, and it is successfully loaded. However, the successful loading of the cookie is not associated with the user successfully logging in. (For quick test case, change password of a "good" cookie account).

### DIFF
--- a/instapy/login_util.py
+++ b/instapy/login_util.py
@@ -94,6 +94,7 @@ def login_user(browser,
     browser.get('https://www.instagram.com')
     # update server calls
     update_activity()
+    cookie_loaded = False
 
     # try to load cookie from username
     try:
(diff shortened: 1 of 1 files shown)


timgrossmann:
Awesome @craigricker ! Thank you very much for this.

Let's wait until someone else from the community can approve everything is working to, then we'll merge this and make your changes available to everyone!

Thank you

CharlesCCC:
Good idea, But I was wondering how would `cookie` know it is changed the password ? When you use `cookie` to login in, there is no password required.  I personally have never experienced the bad `cookie` scenario before and wondering how could that ever happen.   Is instagram.com will expired your cookie in certain number of days ? hours ? to re-check the password ???

craigricker:
@CharlesCCC  `cookie` itself will not know if the password has been changed. Once the cookie has been loaded, if the user is logged in, it continues as normal. However, if loading the cookie is not successfully logging the user in, then there is something "bad" about it.

If the user changes their password for any reason (security, or a forced reset by Instagram) with the current code, they will not be able to log in. As you said, if a cookie exists, the password is never utilized. However, because the password has been changed the cookie is no longer valid, and the user will not be logged in. With the changes in this pull request, InstaPy will detect that loading the cookie is not successfully logging the user in, and manually log the user in, and create a new cookie.

sionking:
Great since my users changes password and it removes the followed pool ! since all in the pool was not followed since no login.
Also one changed the profile name 🥇

PR #1571, opened 2018-03-09
Refactored formatNumber and fixed decimal bug

Tjorriemorrie:
There is a bug that when the count does not end with a decimal e.g. 123k instead of 12.3k, the conversion is incorrectly done with dropping a zero.

Refactored the name to pep8.
Optimized imports.
Added test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -99,3 +99,6 @@ ENV/
 # Editors
 .idea
 .vscode
+
+# pytest
+.pytest_cache/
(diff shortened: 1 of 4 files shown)


CharlesCCC:
Like++++++++++++++
on
`Added test`

timgrossmann:
@Tjorriemorrie Awesome, thank you very much for your contribution!

PR #1578, opened 2018-03-10
Added variable to prevent undefined error

### LINKED ISSUE #1576: self.liked_img not initialized
I just had a crash because self.liked_img in instapy.py line 1115 was not initialized. Looking at the code, if usernames in like_by_users(self, usernames, amount=10, randomize=False, media=None) is empty, it is possible that the for loop will exit without having initialized the variable.

## Expected Behavior
variable is initialized

## Current Behavior
variable is not initialized.

## Possible Solution (optional)
initialize variable after line 964

Tjorriemorrie:
Fixes #1576

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,4 @@ ENV/
 
 # pytest
 .pytest_cache/
+tests/logs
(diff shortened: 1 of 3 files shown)


timgrossmann:
@Tjorriemorrie Perfect, thank you very much!

PR #1596, opened 2018-03-12
Removed unnecessary import condition

### LINKED ISSUE #1579: ERROR [#]  Image check error: name 'check_image' is not defined
Hey im getting an error relating to:

`ERROR [#]  Image check error: name 'check_image' is not defined`

I commented out: `from .clarifai_util import check_image`  but the bot crashes on launch with it commented out. 

The bot is working, im just not getting my clarifai comments - which is sort of my bread and butter because i have so many different arguments / comments based on detection. 

Has anyone patched this up yet? I noticed when i updated to the current build, the likeCompetitors seems to be crashing a lot so i reverted to an old build. Im getting a index out of range for trying to follow another accounts followers. 

Also, if i update to the current version of instapy - im still getting the same bug.
ERROR [#] Image check error: name 'check_image' is not defined

Cheers guys

Tjorriemorrie:
Removed the import condition for clarify which causes an exception. Fixes #1579 

Moved the session creation outside try-except to prevent error for `session is undefined` and added printing of the exception to get better stacktraces in issues.

Made the paths in settings relative to settings path, otherwise exception if ran from different directory. Changed browser_location as it points to the chromedriver specifically. Ref #1627 

Asserting username and passwords to prevent `None type errors`

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -16,9 +16,7 @@ from selenium.webdriver.chrome.options import Options
 from selenium.webdriver import DesiredCapabilities
 import requests
 
-if os.name != 'nt':
-    from .clarifai_util import check_image
-from .settings import Settings
+from .clarifai_util import check_image
 from .comment_util import comment_image
 from .like_util import check_link
 from .like_util import get_links_for_tag
@@ -29,6 +27,7 @@ from .like_util import like_image
 from .like_util import get_links_for_username
 from .login_util import login_user
 from .print_log_writer import log_follower_num
+from .settings import Settings
 from .time_util import sleep
 from .time_util import set_sleep_percentage
 from .util import get_active_users
@@ -48,6 +47,10 @@ from .unfollow_util import set_automated_followed_pool
 
 # Set a logger cache outside the InstaPy object to avoid re-instantiation issues
 loggers = {}
+
+
+class InstaPyError(Exception):
+    """General error for InstaPy exceptions"""
 
 
 class InstaPy:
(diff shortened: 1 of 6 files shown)


instapy/instapy.py
```diff
@@ -16,9 +16,7 @@
 from selenium.webdriver import DesiredCapabilities
 import requests
 
-if os.name != 'nt':
-    from .clarifai_util import check_image
-from .settings import Settings
+from .clarifai_util import check_image
```
timgrossmann, line 19: I think we need the os check here since we came across some problems with clarifai and windows nt in the past..

timgrossmann (commented):
Thank you very much for your contributions!

instapy/instapy.py
Tjorriemorrie: I thought about that. But without the import the code raises. I was thinking of making this a proxy, but easier would be to change `set_use_clarifai` and raise an exception right there if NT. What do you think? because where `check_image` is used, is within that condition of `use_clarifiy`

Tjorriemorrie:
@timgrossmann I've moved the check to set_use_clarifai.
Note that for the test to run I had to abspath the logs location.
I've just read the condition, isn't it the other way around? Can we just double check that? I haven't used clarifai, so is it supported on windows or not supported? Because the condition should be == then?

timgrossmann:
@Tjorriemorrie Awesome!
Let's wait for at least one more approval and then merge this!

Thank you very much!

Tjorriemorrie:
@timgrossmann do you mind tagging a second reviewer?

timgrossmann:
@Tjorriemorrie Sorry, good idea!

@sionking @CharlesCCC Could one of you guys please double check this?

CharlesCCC:
@timgrossmann I will take a look and test it out this weekend. still busy at the daily work/job.

timgrossmann:
@CharlesCCC Sure, thank you very much

Tjorriemorrie:
bump

CharlesCCC:
all the change looks good to me.

timgrossmann:
@Tjorriemorrie Awesome, thank you very much for your work!

PR #1656, opened 2018-03-16
Respect amount on interact_user_followers

texmex44:
At final of interact_user_followers restart if followed < amount. (if param respectAmount is True)

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1316,7 +1316,7 @@ class InstaPy:
 
         return self
 
-    def interact_user_followers(self, usernames, amount=10, randomize=False):
+    def interact_user_followers(self, usernames, amount=10, randomize=False, respectAmount=False):
 
         userToInteract = []
         if not isinstance(usernames, list):
@@ -1353,6 +1353,12 @@ class InstaPy:
                            self.user_interact_amount,
                            self.user_interact_random,
                            self.user_interact_media)
+
+        if self.followed < amount and respectAmount:
+            self.logger.info('!!! Followed : {} < Amount : {}'.format(self.followed, amount))
+            newAmount = amount - self.followed
+            self.logger.info('Restarting with New amount : {}'.format(newAmount))
+            return self.interact_user_followers(usernames, newAmount, randomize, respectAmount)
 
         return self
 
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -105,6 +105,7 @@ class InstaPy:
         self.commented = 0
         self.followed_by = 0
         self.unfollowNumber = 0
+        self.valid_users = 0
 
         self.follow_restrict = load_follow_restriction(self.logfolder)
         self.follow_times = 1
@@ -974,6 +975,7 @@ class InstaPy:
         inap_img = 0
         commented = 0
         followed = 0
+        valid_users = 0
         usernames = usernames or []
 
         for index, username in enumerate(usernames):
@@ -991,6 +993,8 @@ class InstaPy:
             if valid_user is not True:
                 self.logger.info(valid_user)
                 continue
+            else:
+                valid_users += 1
 
             try:
                 links = get_links_for_username(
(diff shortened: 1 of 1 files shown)


timgrossmann:
@texmex44 Thank you for the contribution!
This is a good catch, however I'm not sure if we should use recursion here.
Would you mind changing that to e.g. a while loop?

sionking:
Recursion with no stop condition == memory overflow
We need to add another stop conditions that is unconditioned in IG response (e,g a counter).

texmex44:
@timgrossmann While loop is here 👍 Can you check ? I'm just not sure if i need to put a "break" in the except error or if the return statement do the job ...

timgrossmann:
@texmex44 Seems to work great!

@sionking Can you please double check that?

instapy/instapy.py
```diff
@@ -1316,43 +1324,60 @@ def like_from_image(self, url, amount=50, media=None):
...
-                self.logger.warning(
-                    u'Warning: {} , stopping follow_users'.format(err))
-                return self
-            else:
-                self.logger.error('Sorry, an error occured: {}'.format(err))
-                self.aborting = True
-                return self
+        restart = 0
+        while self.valid_users == 0 or (amount - self.valid_users > 0 and respectAmount):
+            userToInteract = []
```
Tjorriemorrie: use snake case for python, e.g. user_to_interact

instapy/instapy.py
```diff
@@ -1316,43 +1324,60 @@ def like_from_image(self, url, amount=50, media=None):
...
-            if isinstance(err, RuntimeWarning):
-                self.logger.warning(
-                    u'Warning: {} , stopping follow_users'.format(err))
-                return self
-            else:
-                self.logger.error('Sorry, an error occured: {}'.format(err))
-                self.aborting = True
-                return self
+        restart = 0
+        while self.valid_users == 0 or (amount - self.valid_users > 0 and respectAmount):
```
Tjorriemorrie: replace `== 0` with `not self.valid_users`

instapy/instapy.py
```diff
@@ -1316,43 +1324,60 @@ def like_from_image(self, url, amount=50, media=None):
 
         return self
 
-    def interact_user_followers(self, usernames, amount=10, randomize=False):
+    def interact_user_followers(self, usernames, amount=10, randomize=False, respectAmount=False):
```
Tjorriemorrie: respect_amount

instapy/instapy.py
```diff
@@ -1316,43 +1324,60 @@ def like_from_image(self, url, amount=50, media=None):
...
-        self.like_by_users(userToInteract,
-                           self.user_interact_amount,
-                           self.user_interact_random,
-                           self.user_interact_media)
+            self.like_by_users(userToInteract,
+                               self.user_interact_amount,
+                               self.user_interact_random,
+                               self.user_interact_media)
+
+            if respectAmount and self.valid_users > 0 and self.valid_users < amount:
```
Tjorriemorrie: if you make `restart` start with -1 then you don't need the second condition.
If you don't want to change that then you still need to change it to `0 < self.valid_users < amount`. it's also different to the while condition, shouldn't it be the same?

Tjorriemorrie (changes requested):

instapy/instapy.py
texmex44: i can change but in the master code `userToInteract` is used.

instapy/instapy.py
texmex44: ok

instapy/instapy.py
texmex44: ok

instapy/instapy.py
texmex44: ok i prefer to keep restart = 0 but i implement your condition `0 < self.valid_users < amount`

texmex44:
@Tjorriemorrie @timgrossmann is it ok ?

timgrossmann:
@texmex44 Sorry for the long delay... 
Could you give a quick explanation in the README what "respect_amount" does? 
We really need to make sure that the documentation stays descriptive and explains all possible features

texmex44:
@timgrossmann sure i will took the time to do this

PR #1662, opened 2018-03-16
Add cookies support

douwevandermeij:
With cookies support you don't have to login all the time and prevent Instagram from blocking login attempts.


CharlesCCC:
@douwevandermeij  Isn't this feature already created months ago ?

douwevandermeij:
Hmm, I think you're right... I created this code months ago in my own fork in decided to make a PR. Didn't realise someone already added it. So never mind.
On the other hand, I think my code is a bit cleaner.

CharlesCCC:
@douwevandermeij haha, it always worth to check out the latest version first. but good job for the clean code tho. keep up the good work :)

PR #1663, opened 2018-03-16
Install improvements

douwevandermeij:
Add ability to install package as library (instead of main project) and install assets and db as well.


timgrossmann:
@douwevandermeij Awesome, thank you very much!

Let's still wait for one more approval before merging.

douwevandermeij:
I updated the readme to be able to use this new functionality to its full extent.

Tjorriemorrie:
I'm just worried that the readme update will confuse a lot of people that are new to python, which seems to be a lot of users from looking at the issues. Can we not rather keep the basic installation as it is, and rather add a 'preferred' section underneath that, or something like that.

douwevandermeij:
Yes agree, let's keep basic "basic" and add a preferred section.

PR #1664, opened 2018-03-16
faster like button search for image liking

douwevandermeij:
First check if there is a like or unlike button via existing spans, next find the like button if it exists, which is much faster.

### DIFF
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -543,8 +543,14 @@ def check_link(browser,
 
 def like_image(browser, username, blacklist, logger, logfolder):
     """Likes the browser opened image"""
-    like_elem = browser.find_elements_by_xpath(
-        "//a[@role='button']/span[text()='Like']/..")
+    like_elem = []
+
+    # fetch spans fast
+    spans = [x.text.lower() for x in browser.find_elements_by_xpath("//article//a[@role='button']/span")]
+
+    if 'like' in spans:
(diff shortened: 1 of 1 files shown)


timgrossmann:
@douwevandermeij Looks good, thank you very much for the contribution!

Let's just wait for one more approval before we merge this.

instapy/like_util.py
```diff
@@ -543,8 +543,14 @@ def check_link(browser,
...
-        "//a[@role='button']/span[text()='Like']/..")
+    like_elem = []
+
+    # fetch spans fast
+    spans = [x.text.lower() for x in browser.find_elements_by_xpath("//article//a[@role='button']/span")]
+
+    if 'like' in spans:
+        like_elem = browser.find_elements_by_xpath(
+            "//a[@role='button']/span[text()='Like']/..")
     if len(like_elem) == 1:
```
Tjorriemorrie: can remove this line then. and therefore also just remove `like_elem = []`

Tjorriemorrie (changes requested):

instapy/like_util.py
douwevandermeij: agree

timgrossmann:
@douwevandermeij Awesome neat and concise upgrades for InstaPy, I love them!
Thank you very much for your contributions!

douwevandermeij:
You're welcome!

PR #1670, opened 2018-03-17
like by feed generator

douwevandermeij:
Refactor original `like_by_feed` to generator to yield on every liked page to chain individual liked pages. The old `like_by_feed` now calls the generator as a function so it doesn't alter the original functionality.

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1541,7 +1541,12 @@ class InstaPy:
 
         return self
 
-    def like_by_feed(self,
+    def like_by_feed(self, **kwargs):
+        """Like the users feed"""
+        self.like_by_feed_generator(**kwargs)
+        return self
+
+    def like_by_feed_generator(self,
                      amount=50,
                      randomize=False,
                      unfollow=False,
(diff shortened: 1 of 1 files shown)


david-vendel:
I am curious to see that working, as I was working on the same feature 2 days ago, but wasn't working everytime.
Thanks!

douwevandermeij:
I'll add some example code.

MichalGallovic:
Hey @douwevandermeij, just pulled new master, and `like_by_feed` doesn't work for me. I tried to use `bot.like_by_feed(amount=100, randomize=True, unfollow=True, interact=True)` and it seems to be skipped. Could that be becuase of the refactor to generators ? thx

douwevandermeij:
Hi @MichalGallovic, can you be a little bit more specific about what exactly is being skipped?
Try following the `print` statements to see what happened and perhaps create a few more `print` statements to increase verbosity.

MNovak12:
@douwevandermeij I can confirm that this breaks backwards compatibility for me. I used the setting `bot.like_by_feed(amount=200)` and added some "here" print statements but it doesn't seem to even enter the function. Others seem to be having this issue: https://github.com/timgrossmann/InstaPy/issues/1748

PR #1685, opened 2018-03-19
Use chrome proxy without extention

dar2015:
Support headless and UI proxy for chrome based on selenium proxy class (no extension needed)


timgrossmann:
@dar2015 What is the difference between this and the already existent proxy config?

dar2015:
It's required browser extension.
I coulnd't find it working headless.

timgrossmann:
@dar2015  Do we need additional documentation for this?

dar2015:
Absolutely no. It's should work seamless when setup proxy with chrome.

timgrossmann:
@dar2015 Awesome! Thank you very much for this!

PR #1695, opened 2018-03-20
Loggin User Followers (not the own ones) to sqlite for botdetection

timmoh:
For a little bit of statistic analyses we should log the follower if the users we are following, like pics... to the database.
that would be a nice feature to see if the followers increase unnormaly like it does when a bot controls the account :-)
screenshot 2018-03-20 23 29 07

### DIFF
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -6,11 +6,12 @@ from math import ceil
 from re import findall
 from selenium.webdriver.common.keys import Keys
 
+from .settings import Settings
 from .time_util import sleep
 from .util import update_activity
 from .util import add_user_to_blacklist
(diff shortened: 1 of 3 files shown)


Tjorriemorrie:
I think that's already logged to `followerNum.txt`

sionking:
I don't get the purpose, most of us do unfollow a day after, no much statistics from that can be learn.
Another thing, how whould you fill the database with username ? from the CSV ?

timmoh:
for some of us it is interesting no to follow bot user. with the db you are able to  see if there is an unnormal user growth.
This works like: show the following user x days ago an see if there is a huge increase of follower....

Issue #1692, opened 2018-03-20
[SOLVED] [MacOS]: Message: unknown error: Cannot read property '0' of undefined


akogay:
INFO [2018-03-20 22:12:47] [##@my.com]  Session started - 2018-03-20 22:12:47
INFO [2018-03-20 22:13:18] [##@my.com]  **Logged in successfully!**
**Message: unknown error: Cannot read property '0' of undefined**
  (Session info: chrome=65.0.3325.162)
  (Driver info: chromedriver=2.37.544337 (8c0344a12e552148c185f7d5117db1f28d6c9e85),platform=Mac OS X 10.12.6 x86_64)

INFO [2018-03-20 22:13:20] [##@my.com]  Session ended - 2018-03-20 22:13:20
INFO [2018-03-20 22:13:20] [##@my.com]  --------------------

akogay:
quickstart.py open browser, log in my instagram acc, wait a bit and close.
What step I have to see to repair that?

CharlesCCC:
Please post your configure as well.

akogay:
from instapy import InstaPy

insta_username = '##@my.com'
insta_password = '###'

session = InstaPy(username=insta_username,
                  password=insta_password,
                  headless_browser=False,
                  multi_logs=True)

try:
    session.login()
    session.like_by_tags(['dance'], amount=10)

except Exception as e:
    print(e)

finally:
    # end the bot session
    session.end()

akogay:
macOS Sierra
**iMac:InstaPy imac$ python --version
Python 3.6.4 :: Anaconda, Inc.**

Tjorriemorrie:
@akogay can you pull the latest from master and run it again, and provide the full stacktrace please?

akogay:
Thank you gyus so much. It work now!

roliag:
Hi, exactly the same thing happens to me, how did you solve it?

roliag:
I solved it, my problem was that I put the email in the user and not the user

PR #1698, opened 2018-03-21
Provide full traceback on exception in quickstart

Tjorriemorrie:
Relates  #1692 

The single line of the error is not helpful, rather print the whole stacktrace.


timgrossmann:
@Tjorriemorrie Thank you, very good idea!

PR #1699, opened 2018-03-21
Fixes incorrect Logfolder

Tjorriemorrie:
I've noticed that the logfolder are created in the root incorrectly. This fixes that.

Apologies!

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -87,7 +87,8 @@ class InstaPy:
         self.nogui = nogui
         self.logfolder = Settings.log_location + os.path.sep
         if multi_logs is True:
-            self.logfolder = '{}/{}/'.format(Settings.log_location, self.username)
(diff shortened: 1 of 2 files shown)


timgrossmann:
@Tjorriemorrie Could we please still stick to the `.format()` way? Just to keep it consistent?

Tjorriemorrie:
@timgrossmann done.

I personally think using the platform separator is more safe though.

timgrossmann:
@Tjorriemorrie You're right...
Can't we combine that?

Tjorriemorrie:
I don't know what you mean by that _combine_?

timgrossmann:
@Tjorriemorrie Use the format method but still use the separator

timgrossmann:
@Tjorriemorrie You're awesome!
Thank you very much

PR #1704, opened 2018-03-21
Adds the ability to output logging to Slack via webhook

jrg89:
I'm far from an expert so there may be issues with python formatting or code standards, but this is working for me and I find it useless for a setup on a server.

Details of how to integrate are added to the README


jrg89:
Messed up the PR, closing and will re-submit

PR #1705, opened 2018-03-21
Adds ability to output logging to Slack

jrg89:
Adds support for Slack notifications through the incoming webhook feature.
This just outputs the logging statistics.

I'm far from a developer, so there may be some issues with the way I've made the changes but it works for me and I find it very useful to get updates when running on a server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ Table of Contents
 * [Clarifai ImageAPI](#clarifai-imageapi)
 * [Running on a Server](#running-on-a-server)
 * [Running on a Headless Browser](#running-on-a-headless-browser)
+* [Slack Notifications](#slack-notifications)
 * [Running Multiple Accounts](#running-multiple-accounts)
 * [Running with Docker microservices manual](#running-with-docker-microservices-manual)
 * [Running all-in-one with Docker (obsolete)](#running-all-in-one-with-docker-obsolete)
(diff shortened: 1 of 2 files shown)


timgrossmann:
@jrg89 This is a really good idea!
But judging on the repetition in the code, this looks like something that should be extracted into a separate file...

jrg89:
@timgrossmann I assumed it wouldn't be the most elegant solution! Extracting it would be beyond my expertise so I'm happy for someone else to pick this up if they think it'd be useful

timgrossmann:
@jrg89 Alright, thank you very much for throwing this into the bucket!
This is a really nice feature I'd love for people to be available.

blitzh0t:
it's a good idea... I've been running the script on a server (Google Cloud)  along with a job scheduler to run every x hours or so.  I don't always have access to a terminal to check how the script is running and it'll be a very handy feature..

imjustin:
This is tangential to what was created/discussed here, but I implemented basic slack logging by simply including it on my quickstart.py file, and triggering it to send me a message every time the script runs, and informing me each step of the way. it doesn't work on a lower level (e.g., it'll tell me once i'm done following users, but won't report out each individual user followed).

i bring it up just to share another way i'd thought of to implement slack reporting; it doesn't have to be super complex to give the necessary amount of info?

converge:
It would be great as a module.

timgrossmann:
@jrg89 As @converge mentioned, would be perfect if you could create a module for this as @converge did for telegram: https://github.com/converge/instapy_telegram_extension

HCWcoder:
```
C:\Users\xxx\Documents\GitHub\InstaPy>python xxx.py
  File "xxx.py", line 12, in <module>
    from instapy import InstaPy
  File "C:\Users\JayPixx\Documents\GitHub\InstaPy\instapy\__init__.py", line 3, in <module>
    from .instapy import InstaPy
  File "C:\Users\xxxx\Documents\GitHub\InstaPy\instapy\instapy.py", line 39, in <module>
    from .unfollow_util import follow_given_user_followers
ImportError: cannot import name 'follow_given_user_followers' from 'instapy.unfollow_util' (C:\Users\xxx\Documents\GitHub\InstaPy\instapy\unfollow_util.py)
```
```
slack_url = 'https://hooks.slack.com/services/xxxx/xxx/xxxx'
                  password=insta_password,
                  slack=slack_url,
                  headless_browser=False,
                  disable_image_load=True,
                  multi_logs=True,
                  bypass_suspicious_attempt=False)
```

timgrossmann:
@HCWcoder Thank you for reporting the error.
This PR is quite old and we just closed it since this shouldn't be part of InstaPy core but rather an extension like the telegram extension mentioned in the comment above.

Thank you
Best
Tim

PR #1743, opened 2018-03-25
Logs in separate files for every start

denisemenov:
Change "general.log" to "general_2018_03_02_16_34_48.log".
This helps to find the correct log if necessary.  
For me it was much more useful than a single large file.
Perhaps it will be needed by someone else.

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -167,7 +167,7 @@ class InstaPy:
             # initialize and setup logging system for the InstaPy object
             logger = logging.getLogger(__name__)
             logger.setLevel(logging.DEBUG)
-            file_handler = logging.FileHandler('{}general.log'.format(self.logfolder))
+            file_handler = logging.FileHandler('{}'.format(self.logfolder) + 'general' + datetime.now().strftime('_%Y_%m_%d_%H_%M_%S') + '.log')
             file_handler.setLevel(logging.DEBUG)
             extra = {"username": self.username}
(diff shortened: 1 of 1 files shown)


sionking:
please make it optional

instapy/instapy.py
```diff
@@ -167,7 +167,7 @@ def get_instapy_logger(self, show_logs):
             # initialize and setup logging system for the InstaPy object
             logger = logging.getLogger(__name__)
             logger.setLevel(logging.DEBUG)
-            file_handler = logging.FileHandler('{}general.log'.format(self.logfolder))
+            file_handler = logging.FileHandler('{}'.format(self.logfolder) + 'general' + datetime.now().strftime('_%Y_%m_%d_%H_%M_%S') + '.log')
```
Tjorriemorrie, line 170: Use TimedRotatingFileHandler. Probably need to set the `suffix` and `extMatch` as well.

Tjorriemorrie (changes requested):
Can also make that the default, with a setting to keep it one file.

I can pick this up and make the change if you can wait a week or two.

CharlesCCC:
@Tjorriemorrie I would agree with @sionking on this one, consider we only look at the log when we need debug and checkout. so I would say default is one file and you have the option to log it day by day(run by run) as necessary.

denisemenov:
I agree that it's right to make this an additional option, but I do not have enough knowledge to do so ... :)

HansDaigle:
It dosen't have to be in instapy, you can just add the handler yourself if you need to.

PR #1690, opened 2018-03-20
Chromedriver validations

Tjorriemorrie:
Fixes #1687 (and many others). Hopefully the error message will provide the user the information to update the chromedriver instead of raising issues.

PR #1752, opened 2018-03-26
Added setting to skip chromedriver validation

Tjorriemorrie:
Fixes #1690 by giving them the option to opt-out of the validation. Still, the default should to perform the validation - not sure if 32bit versions are lacking behind or not being developed anymore.


Tjorriemorrie:
Thinking about it now, the min version a setting for this reason, it can just be changed.

Issue #1819, opened 2018-04-08
set_dont_unfollow_active_users only gets first 20 users and from "n-1" posts


ortegafernando:
## Expected Behavior
I think that set_dont_unfollow_active_users  will have to get all likes from posts.

## Current Behavior
It only get first 20 users for each post
Also I think it also check the first "n-1" posts. If you select only 1 posts, you will get a empty list and ERROR

## Possible Solution (optional)

I am using "get_active_users" function (that is what set_dont_unfollow_active_users uses) and it only gets first 20 users. I think it needs to scroll down.

Thanks a lot for your help
(similar to https://github.com/timgrossmann/InstaPy/issues/851 ?
This is the script that I am using to check it:

```from instapy import InstaPy
from instapy.util import get_active_users

def main():
	try:
		session = InstaPy(username='USERNAME', password='PASSWORD')
		session.login()
		
		# list of users who liked our media. ALSO check with 1 instead 2 posts
		active_users = get_active_users(session.browser,session.username,2,session.logger)
		session.logger.info('Active_users: %d', len(active_users))
		print('Active_users: %d', len(active_users))
		session.end()
		quit()
	except Exception as e: 
		print('Error desconocido: {}'.format(e))
	
if __name__ == '__main__':
    main()

**Comment by uluQulu** [comment]
Hey @ortegafernando, it's a known issue, I have once looked into it for Jonas from the Slack channel and then I wrote a smart scroller which could scroll down through all dialog boxes, grab all usernames and when hits bottom it would switch to the next post, so **all of** usernames who liked your latest n posts was grabbed...

"_Also I think it also check the first "n-1" posts._"  - **you are right**, the reason **behind** that is:
```python
for count in range(1, posts):
```
**line** in `get_active_users` function of **util.py** file,
it **should** be 
```python
for count in range(0, posts):
```

**Currently** InstaPy takes only the visible usernames in the opened dialog boxes without scrolling down, **but** if we did apply scrolling deep down and taking all usernames, then it would take some **big time** for people _with huge_ accaunts who gets about **~5k likes** per posts, **but** good news is that we can add a new parameter to ask users to choose how many active users to grab per post..

>If you **really** decide to grab **entire** active usernames, I can look into backups to find that solution laying somewhere..

ortegafernando:
Thanks a lot @uluQulu 

[...]

PR #1855, opened 2018-04-15
Detect likers past 20 on posts by scrolling

d7laungani:
Fixes issue #1819 .

The issue is that it detects only the first 20 users that liked a post. This pull requests implements code by @uluQulu that contains a scroller to go through the list of likes. 

Note: Video likes do not show up on a desktop and won't be found by this function.

Boundary controls the amount of likes that should be looked at so that users can control how long they want the function to take.

if Bounday = 0 then grab only visible names as the default
if Boundary = None then grab all the usernames that liked a photo
if Boundary = xx then grab xx amount of likers starting from the top.

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1773,7 +1773,7 @@ class InstaPy:
 
         return
 
-    def set_dont_unfollow_active_users(self, enabled=False, posts=4):
+    def set_dont_unfollow_active_users(self, enabled=False, posts=4, boundary=500):
         """Prevents unfollow followers who have liked one of
         your latest X posts"""
 
@@ -1785,6 +1785,7 @@ class InstaPy:
         active_users = get_active_users(self.browser,
                                         self.username,
                                         posts,
+                                        boundary,
                                         self.logger)
 
         for user in active_users:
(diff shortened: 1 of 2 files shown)


timgrossmann:
@d7laungani Looks good, thank you very much for fixing this!

uluQulu:
I want to acknowledge all the extra effort for this commit getting merged 🎊

A few notes, using `sleep` instead of `sleep_actual` from **time_util** will result a **slightly** randomized timeout which suits the desired stochastic activity
>`sleep` itself, I think, returns in 99.73% range of the called value (_not too delayed_)

Also, there should be a few lines in README so that people undestands the usage of the boundary variable (_I've also written a resourceful words for the usage [there](https://github.com/timgrossmann/InstaPy/issues/1819#issuecomment-380495531) in #1819)_

And,
```python
sleep_actual(1.2)  # old value 5.6 
```
line, I remember while tests, scroll down sometimes was taking more time than usual and **regarding** the design of the standalone scroller here (_detect once hits bottom to save unneeded requests_) once it tries to scroll again (_after ~1.2 seconds_) _while_ previous scroll down **did not accomplish** `offsetHeight + scrollTop` will be `>=`  `scrollHeight` and it will return **false** (_unwanted behaviour, although, `try_again` **will handle** it for nothing_)..
**So**, not even using old value 5.6, use _at least_, `4.2` or `3.5` seconds, to operate effectively, it's just `1.2` seconds is so less for users with lower bandwith speeds.

PR #1868, opened 2018-04-17
Add docker alpine image

funkypenguin:
In preparation for deploying InstaPy in Docker Swarm, I created an additional Dockerfile using Alpine instead of Python base image, because:

1. The image can be auto-built using docker hub, making it quick to integrate into swarm
2. The alpine image is 329MB, whereas the python image is 807MB

I'll send another PR for a docker-compose file which will actually _use_ the alpine image, this is just a preparatory contribution.


timgrossmann:
@funkypenguin This is awesome, sounds like good improvements!
Thank you very much

PR #1869, opened 2018-04-17
Improve docker alpine image

funkypenguin:
Made a stupid copy/paste error in the first PR. It was hard to test, bit of a chicken-and-egg issue :)

Anyway, this one is confirmed working in swarm. I also added "GitPython" to requirements.txt (since it gets baked into the image), so that I can later force a "git pull" on every subsequent bot run, using something like this (untested):

````
# Before we start, let's git-pull the latest code
import git
g = git.cmd.Git('.')
g.pull()
````


timgrossmann:
@funkypenguin Awesome, thank you!

Issue #1858, opened 2018-04-15
How to limit users by the number of following 


DimIsaev:
## Expected Behavior
How to limit users by the number of **following**

## Current Behavior
Now only there is a limit on the number of **followers**

> session.set_upper_follower_count(limit = 250)
> session.set_lower_follower_count(limit = 1)

how to set a limit for **following**

DimIsaev:
Sorry my english  is  not very good.

how to create a user filter for the subscription, use the number of his subscriptions to other users?
[Screenshot](https://c2n.me/3TuiesD.png)

uluQulu:
hi @DimIsaev, so you want:
```python
set_upper_following_count(limit=1299)
set_upper_following_count(limit=199)
```
it is very easy to achieve, do you really need it?

DimIsaev:
@uluQulu  

yes yes it is very necessary

PS  This setting can filter out not very attentive / not real users
it is impossible to monitor more than 500 users
Most likely these accounts are bots

uluQulu:
Understood, @DimIsaev,
What about some sophisticated version?
```python
session.set_relationship_bounds (enabled=True,
                                  potency_ratio=1.25,
                                   delimit_by_numbers=True,
                                    max_followers=120000,
                                     max_following=2200, 
                                      min_followers=442, 
                                       min_following=132)
```
>`potency_ratio` == `followers count`/`following count` of a user

**1**. you can **use** `potency_ratio` **or not** (`potency_ratio=None`, `delimit_by_numbers=True`), - _will decide only by your pre-defined `max` & `min` values regardless of the `potency_ratio`_
**2**. also **you can** use **only** `potency_ratio` (`potency_ratio=1.5`, `delimit_by_numbers=False`) - _will decide per `potency_ratio` regardless of the counts_ 
**3**. or **use both** `potency_ratio` **and** predefined `min`&`max` values (`potency_ratio=1.5`, `delimit_by_numbers=True`) - _will decide per `potency_ratio` & your predefined `min` & `max` values_

DimIsaev:
@uluQulu  

Yes, Will be very cool 
I'm waiting for your implementation

uluQulu:
@DimIsaev, it's done #1876

DimIsaev:
@uluQulu 
tested the script with your patch

setting  
``` 
   session.set_relationship_bounds(enabled=True,
    #					 potency_ratio=1.2,
    #				  delimit_by_numbers=True,
[...]

Issue #1338, opened 2018-02-02
Regular Crashes - Unknown error: Cannot read property 'entry_data' of undefined


Tachenz:
Running latest version of chrome driver/chrome/Instapy.
Standard configuration (just likes/comments by hashtag).

Crashes every 10-15 minutes with this message:
![1](https://user-images.githubusercontent.com/36073534/35731993-b18fd59c-080f-11e8-99e2-c4fda896c441.png)

uluQulu:
Sometimes latest versions of chromedriver, GoogleChrome and selenium are incompatible..
Start by downgrading Google Chrome to previous working version _and disable automatic updates_, maybe it's the problem..
_In Windows 10, I'm still using Google Chrome v62 without problems_

ArdySan2012:
Aha, I did not know that! Will give it a try. Which chrome driver version do you use with v62?  I have this issue all the time, will try downgrading to 62

Wisdomlocker:
Greetings!  getting exactly the same error at random intervals.  Love the program but the error is very annoying because I have to restart every twenty minutes.  Just doing basic following, liking, commenting, nothing fancy

Tachenz:
Will try that tonight!

uluQulu:
@ArdySan2012 
ChromeDriver 2.34.522940
Google Chrome Version 62.0.3202.62 (Official Build) (64-bit)
Selenium Version 3.8.1

@Tachenz 
For some non-specific policital reasons installing an old version of Google Chrome is a big pain, so is disabling it's updates..
Here I included some helpful links I used last time,
**There are a few websites share old Google Chromes,**
[FilePuma](www.filepuma.com/download/google_chrome_64bit_63.0.3239.132-17914/)
[Slimjet](https://www.slimjet.com/chrome/google-chrome-old-version.php)
_(Google doesn't offer old version officially or inofficially)_

**To disable Google Chrome updates see:**
Read instructions [(Google official paper)](www.chromium.org/administrators/turning-off-auto-updates)
Download [Group Policies from Google](https://support.google.com/chrome/a/answer/6350036#Policies)
> from there, you must download **Administrative Template XML-based (ADMX) file** - it is a Policy template to be added to `gpedit.msc` in order to disable Google Updates from policies..

Be sure to check executables at [Virus Total](https://www.virustotal.com/#/home/upload) before installing

_Personally, 
[...]

PR #1876, opened 2018-04-18
Enhanced interaction verification based on relationship bounds

uluQulu:
As of @DimIsaev's request from #1858, revised and enhanced helpful feature


timgrossmann:
@uluQulu This is great, thank you!

Would you be able to add two more examples in the readme?
It might be a little bit confusing to only see the whole example with all the values set.

Maybe change it to one using the ratio, one only using upper and lower bound and one where only one of the values is set. Add the documentation you put below the big example to every example on it's on.

This will clear things up a lot. 

I really like this, thank you very much for all the hard work you put into this.

sionking:
Thanks
I will test it someday :)

uluQulu:
Welcome @timgrossmann,
I've also thought of adding more examples but in favor of saving some space in README I put only one example, but **now** I think you're right, it is setting **4 + 1** configuration which could be confusing to new users, so just **added** all of the required examples in the simplest form and rewrote text within markdown syntax.
>BTW thanks for inviting to collaborate, it is an honor! 🎖

My pleasure, @sionking, due to being requested too many times I've added three blocks of getting _followers_/_following_ count (2 different _XPATH_ & a _graphql_) 🎯

sionking:
I do like your coding style.
I acctually using graphql only for long time with no issues, adding XPATH may results in incorrect numbers (windows size change the xpath for example) so I would have start with graphql then check for xpath

uluQulu:
Thanks mate, @sionking,
I really thought about it and if remembers that problematic `entry_data` queries sometimes stucks in temporary outage (_in `Issues` panel, there are threads from last year July as of last times_) and needs reloading (_eats from time_) and it happened so often to so many guys, so, recently rewrote an XPATH variant in #1338 already and used it here, too. 
>Also, after the latest change in some endpoints so-called _stable_ graphql suffers.  

Well, we are dependant on page elements in whole program if they differ after window size change, it would crash in many places before reaching this part, and if anytime some class name changes, we can re-find them easily with plain `inspection`, take it easy, page elements are good  xD
_BTW_ I also did take their time difference, before and after, for each 3 and it takes (_with `time.time()`_)
1st _XPATH_: 0.14901328086853027
2nd _XPATH_: 0.15800833702087402
`entry_data`: 0.020002126693725586 (without reload)
_graphql approach is ~0.13 seconds faster (_but if it takes a reload after outage then it will be ~1 seconds slower_)_.
You choose, a bit of speed or easier stability..

uluQulu:
Heh, I misunderstood, @sionking, you mean, _e.g._, after a window resize _it can get_, _e.g._, _**posts count**_ in `li[2]` element instead of **followers count**, although 1st XPATH can't do that cos it directly looks for followers & following keywords `@href,'followers')]/span"` & `@href,'following')]/span"` in links, but 2nd one with `li` elements _has a drolling possibility_ to misbehave as you said.
I will replace the order in your suggestion. 🙂

timgrossmann:
@uluQulu Awesome, thank you.

One last thing we have to change might be the layout and "style" of the README part, the link from the "table of content" is not working and it looks a bit overwhelming. But that's just a small fix.

uluQulu:
Welcome, @timgrossmann,
Yeah right, I noticed Atom's markdown **previewer** shows in _different looks_ than Github's _webpage_ markdown viewer. It looks well now.
I think modifications are done.

timgrossmann:
@uluQulu Awesome, thank you very much!

PR #1912, opened 2018-04-25
update login_util.py

matousek-martin:
added time.sleep(1), my script got from time to time stuck on google.com, this solved it

i think it might be just me, i have a small laptop

### DIFF
--- a/instapy/login_util.py
+++ b/instapy/login_util.py
@@ -5,6 +5,7 @@ from selenium.common.exceptions import NoSuchElementException
 from selenium.common.exceptions import WebDriverException
 from .util import update_activity
 import pickle
+import time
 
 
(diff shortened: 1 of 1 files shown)


CharlesCCC:
looks good to me.

timgrossmann:
@MMorte Never had that problem, but I have a good connection and quite powerful computer, therefore let's add this, seems like it won't hurt 😀

matousek-martin:
@timgrossmann i guess it wont, i have cca 100mb/s connection, but only [this laptop ](https://www.amazon.com/Lenovo-Thinkpad-Business-Notebook-Black/dp/B00U7ZXN8Y) which might be the issue but i looked through older issues and nobody else reported soooooooooooooo long story short ill leave the decision up to you haha

PR #1913, opened 2018-04-25
fix password in numbers only

sionking:
Fix the exception if the password is number only


timgrossmann:
@sionking Awesome! Good catch.

PR #1916, opened 2018-04-26
Add option to use logger outside instapy moudle

sionking:
Create a pointer to this handler in settings
the use should be as follow:
```
from .settings import Settings

logger = Settings.logger
logger.info('sleep for long nap')
```


timgrossmann:
@sionking Thank you!

PR #1939, opened 2018-05-01
Additional Fix the issue that cant get followed by count

CharlesCCC:
For some user (a.k.a myself) that can't get `graphql` after login. 
Add back the old piece of code which will do a refresh to reload the data to prevent the `cannot read property 'graphql' of undefined` error.


timgrossmann:
@CharlesCCC Thank you for this, could you double check if this is still a problem after #1940 , please?

CharlesCCC:
@timgrossmann Hi Tim, this change is still necessary to merge in with the PR from #1940  (Actually it was in, but somehow/someone decided to remove it)
The reason be, if Instagram.com decide to change their data again. It won't blow up during the run, but instead, it will logged `followed` amount to 0.

SummitKaiju:
I am unable to run today also.

 /instapy/like_util.py", line 461, in check_link
    media = post_page['media']

TypeError: list indices must be integers or slices, not str

SummitKaiju:
Nevermind, I figured it out after looking over the code. The media='None' was incorrect. It needed to be set to 'Post'

PR #1969, opened 2018-05-04
Revised `like_by_users` feature to speed up the process

uluQulu:
_`like_by_users` is prob my fav feature to have fun sometimes liking bunch of pics but it has an unwanted behavior should be changed.._
**Currently**, `check_link` utility is visiting the user's profile page in every posts in order to find followers & following counts.
_And it does **the same** for the feature of `like_by_users` where, from the same user(s) it views bunch of posts and visits profile page in each post to get those values..._
#### It is simply unwanted behavior. So I **modified** a few lines:
**Now** when a user's relationship bounds **is matched** for once, it **will not be checked** in the next posts of that same user again.
**Similarly**, if a user's relationship bounds **is not matched** with the range desired, then it **will simply skip** that user _without seeing further posts_.
>This applies to all features using `like_by_users` method and affected features are:
`interact_user_followers`, 
`interact_user_following`, 
`follow_user_followers`,
`follow_user_following`,
`like_by_feed`,
`like_by_tags`,
and `like_by_users` itself :)

**The result** is lots of saved `time` and less web `requests` sent. 

**_Also_**, **moved** the process of _getting follower and following counts_ into **a new** function for the _ease of maintenance_ and _to make usable_ in other places, too, e.g., right from quickstart scripts...
_e.g._,
```python
##quickstart script
from instapy.unfollow_util import get_relationship_counts
#some code

followers_count, following_count = get_relationship_counts(session.browser, insta_username, session.logger)
```
> If there is something I missed, please notify.

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -676,7 +676,7 @@ class InstaPy:
                 self.logger.info(link)
 
                 try:
-                    inappropriate, user_name, is_video, reason = (
+                    inappropriate, user_name, is_video, reason, scope = (
                         check_link(self.browser,
                                    link,
                                    self.dont_like,
@@ -834,7 +834,7 @@ class InstaPy:
                 self.logger.info(link)
 
                 try:
-                    inappropriate, user_name, is_video, reason = (
+                    inappropriate, user_name, is_video, reason, scope = (
                         check_link(self.browser,
                                    link,
                                    self.dont_like,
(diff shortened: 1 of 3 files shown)


timgrossmann:
@uluQulu This is so awesome and as you mentioned, long overdue. 
Thank you very much @uluQulu !

PR #1989, opened 2018-05-08
Added exception 'InvalidElementStateException' because comment_input[…

dbknv:
Added exception 'InvalidElementStateException' because comment_input[0].clear() often finishes with error. The reason is comment_input[0] hangs in disabled state.


CharlesCCC:
chang looks good to me.

Issue #1975, opened 2018-05-06
Relationship ratio: unknown


redsector72:
I suppose there is a bug.
If the ratio is almost a unit, like 1 or 2  i get in the log:
followers: 1253  |  following: 1276  |  relationship ratio: 1.0
followers: 1048  |  following: 2610  |  relationship ratio: 2.0

otherwise i get ratio: unknow like:
241  |  following: 186  |  relationship ratio: unknown
1159  |  following: 778  |  relationship ratio: unknown

my setup is: session.set_relationship_bounds(enabled=True,potency_ratio=-1.05,delimit_by_numbers=True,max_followers=None,max_following=None,min_followers=30,min_following=30)

uluQulu:
Hi @redsector72,
Can you show a console output of that example?
I would like to see, cos the value of `unknown` should only be returned when either _followers count_ or _following count_ was _unknown_.

redsector72:
is this a console log?

INFO [2018-05-06 08:34:36] [uuu]  https://www.instagram.com/p/Bib3-jGA1y2/?taken-at=214548286
INFO [2018-05-06 08:34:39] [uuu]  Image from: ceci1602
INFO [2018-05-06 08:34:42] [uuu]  User: ceci1602 >> followers: 154  |  following: 150  |  relationship ratio: unknown

[8/10]
INFO [2018-05-06 08:36:08] [uuu]  https://www.instagram.com/p/Bib6gHChXhU/?taken-at=213050058
INFO [2018-05-06 08:36:10] [uuu]  Image from: tjevina
INFO [2018-05-06 08:36:15] [uuu]  User: tjevina >> followers: 394  |  following: 387  |  relationship ratio: unknown
INFO [2018-05-06 08:36:15] [uuu]  Link: https://www.instagram.com/p/Bib6gHChXhU/?taken-at=213050058
INFO [2018-05-06 08:36:15] [uuu]  Description: “birds” 🐦🦅
INFO [2018-05-06 08:36:17] [uuu]  --> Image Liked!

uluQulu:
Very weird! @redsector72,
I wonder what is your system (`OS` & python `version`)
But try this:
in **like_util.py** file, find and **replace**
```python
float("{0:.2f}".format(relationship_ratio)) if relationship_ratio else 'unknown'))
```
**with**
```python
float("{0:.2f}".format(relationship_ratio)) if relationship_ratio is not None else 'unknown'))
```
_See if it works, now_

[...]

PR #1997, opened 2018-05-11
#1975 float 

redsector72:
This fix the issue where interger are not generating a float on python 2.7 and earlier.


uluQulu:
Amazing @redsector72 ✨

timgrossmann:
@redsector72 Awesome, thank you very much!

PR #2005, opened 2018-05-12
Prevent UnicodeDecodeError when inappropriate words found

joanroig:
The bot currently stops if the image has any inappropriate word.
Fix bot error when inappropriate words found: "UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 43: ordinal not in range(128)"

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1027,7 +1027,7 @@ class InstaPy:
                             already_liked += 1
                     else:
                         self.logger.info(
-                            '--> Image not liked: {}'.format(reason.encode('utf-8')))
+                            '--> Image not liked: {}'.format(reason))
                         inap_img += 1
                 except NoSuchElementException as err:
                     self.logger.error('Invalid Page: {}'.format(err))
@@ -1209,7 +1209,7 @@ class InstaPy:
                             already_liked += 1
                     else:
                         self.logger.info(
-                            '--> Image not liked: {}'.format(reason.encode('utf-8')))
+                            '--> Image not liked: {}'.format(reason))
                         inap_img += 1
                 except NoSuchElementException as err:
                     self.logger.error('Invalid Page: {}'.format(err))
@@ -1398,7 +1398,7 @@ class InstaPy:
 
                     else:
                         self.logger.info(
-                            '--> Image not liked: {}'.format(reason.encode('utf-8')))
+                            '--> Image not liked: {}'.format(reason))
                         inap_img += 1
                 except NoSuchElementException as err:
                     self.logger.error('Invalid Page: {}'.format(err))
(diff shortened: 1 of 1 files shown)


uluQulu:
Hi @joanroig,
Why don't you just remove encoding of `quashed` at,
```python
quashed = (((quash.group(0)).split('#')[1]).split(' ')[0]).split('\n')[0].encode('utf-8')   # dismiss possible space and newlines
```
_which is causing the error while repeatedly attempting to encode._

Or if you decide not to encode them in `reason`,
you should encode `iffy` too, alongside `quashed`, so that it will not crash again.
> I am the one who implemented those `quashed` and `iffy` variables, so it is a bit of my mistake, too 🔨 

Best regards!

timgrossmann:
@joanroig Thank you for the contribution. 

However, we can not "simply" remove the encoding, this will lead to way more problems than it fixes.

Let's try about any other way to fix this issue.
It might only be a problem with the encoding of your terminal.

Looking forward to future contributions

joanroig:
@uluQulu great, I'll check your solution at the afternoon :)
@timgrossmann no problem, I've just checked it on my Mac and on a Raspberry Pi 2, the error was reproducible every day in both environments when trying to interact with an image that contains inappropriate hashtags and was solved with those changes (I use scheduled tasks). I'll try to use @uluQulu recommendations

joanroig:
@uluQulu you suggestion does not worked for me. For testing that, can you try running quickstart with this code please? :

```
    session.set_do_like(True, percentage=100)
    session.interact_by_users(['mesdevi'], amount=4, randomize=True, media='Photo')
```

This user uses some prohibited hashtags, please tell me if you got the same error:

```
  File "/Users/jotakode/Documents/GitHub/InstaPy/instapy/instapy.py", line 1619, in interact_by_users
    '--> Image not liked: {}'.format(reason.encode('utf-8')))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 43: ordinal not in range(128)
```

Thanks!

joanroig:
@uluQulu you suggestion does not worked for me. For testing that, can you try running quickstart with this code please? :
   ```
 session.set_do_like(True, percentage=100)
    session.interact_by_users(['mesdevi'], amount=4, randomize=True, media='Photo')
```
This user uses some prohibited hashtags, please tell me if you got the same error:
  ```
File "/Users/jotakode/Documents/GitHub/InstaPy/instapy/instapy.py", line 1619, in interact_by_users
    '--> Image not liked: {}'.format(reason.encode('utf-8')))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 43: ordinal not in range(128)
Thanks!
```
From my side it is solved changing all `Image not liked: {}'.format(reason.encode('utf-8')))` to `Image not liked: {}'.format(reason))` in instapy.py

uluQulu:
@joanroig,
Although it did not crash in my side, it actually did show that unreadable character which is a terminal specific issue.
And in that particular page, it encounters a non-breaking space (`\xc2\xa0` ), which you can replace with a normal space.
_That must be solved, cos there, `quashed` variable is being converted into _bytestring_ 2 times._
Can you tell which part you did modify exactly?

Issue #2009, opened 2018-05-13
Not just follow but also like and comment for follow_user_following


FranklinClinton:
## Expected Behavior
- InstaPy follows every person on the following list
- Instapy likes one picture of the person
- Instapy comments on the same picture
## Current Behavior
- InstaPy follows every person on the following list
- Instapy **does not** like one picture of the person
- Instapy **does not** comment on the same picture
## Possible Solution (optional)
Tell me :-(
## InstaPy configuration
My Script:
```
import sys
from instapy import InstaPy

insta_username = sys.argv[1]
insta_password = sys.argv[2]

session = InstaPy(username=insta_username,
                      password=insta_password,
                      nogui=True)
session.login()

session.set_relationship_bounds(enabled=True,
                         potency_ratio=None,
                          delimit_by_numbers=True,
                           max_followers=4590,
                            max_following=5555,
                             min_followers=45,
                              min_following=45)

session.set_user_interact(amount=1, randomize=True, percentage=100, media='Photo')
session.set_do_like(enabled=True, percentage=100)
session.set_comments([u':heart_eyes: :heart_eyes:', u'Nice :heart:'])
session.set_do_comment(enabled=True, percentage=100)

session.follow_user_following(['THEUSERNAME'], amount=100, randomize=False, interact=True)

session.end()
```

uluQulu:
hi @FranklinClinton,
Currently interactions happen after following the **all of** the users, and after then you can get likes & comments at those interactions..
But **soon** I am gonna share enhanced versions of a few functions, where for `follow_user_followers` feature, interactions will happen right after following a user, also, then the users can be matched against relationship bounds

FranklinClinton:
@uluQulu Thank you for all your efford. It's not the first time I see you fixing things! For now I will just use session.interact_user_following which seems to be working for me. Instapy ignores all bounds in regard of following. He follows everybody and does not respect the bounds in that matter. But that's  another issue. Thank you!

uluQulu:
Welcome, @FranklinClinton,
_In that PR, comes that feature, too!_
I have added **relationship bounds** _verification_ support for `follow_user_followers` and some others missing it. (_watch PRs_)

Issue #2003, opened 2018-05-12
Code doesn't follow relationship bounds


harrycha:
## Expected Behavior
I expect this code to sort through a given profiles followers and follow only the ones with the relationship bounds
## Current Behavior
It ends up following all of the users even though I've set potency_ratio at 1.20 meaning it wont interact with users under it
## InstaPy configuration

try:
    session.login()

    # settings
    session.set_relationship_bounds(enabled=True,
				 potency_ratio=1.20,
				  delimit_by_numbers=True,
				   max_followers=7000,
				    max_following=5555,
                                     min_followers=45,
				      min_following=77)
    session.set_do_comment(True, percentage=10)
    session.set_comments(['nice', 'nioce'])
    session.set_delimit_liking(enabled=True, max=1020, min=20)
    session.set_delimit_commenting(enabled=True, max=200, min=5)
    session.set_user_interact(amount=3, randomize=True, percentage=30, media='Photo')
    session.set_do_follow(enabled=True, percentage=100)
    session.set_delimit_liking(enabled=True, max=1020, min=20)
    session.set_delimit_commenting(enabled=True, max=200, min=0)
    session.set_do_like(enabled=True, percentage=70)
   # actions
    session.follow_user_followers(['XXXXXX'], amount=1000, randomize=True)

uluQulu:
hi @OGHARRY,
It is working inside a dialog box where you can't expect to verify user's relationship bounds.
Previously it has been issued at #1858 by @Dimlsaev and I got it implemented yesterday, soon, will share it in a PR
> Lots of design has changed! (_watch PRs_)

Issue #1548, opened 2018-03-05
Follow User Followers Issue


cormo1990:
**## Expected Behavior**

Please see the attached image.
I expect to follow 1000 users from a selected user follower list.

**## Current Behavior**

At the moment i can't follow 1000 users, the message is:
-"username" less user to follow than requested 
the number of followers of the account selected is around 20K

Thanks for the support
i hope that the problem is explain clearly.
![followuserfollowers](https://user-images.githubusercontent.com/36590321/36962200-dda4c6ea-204e-11e8-8830-ba9efaf03e90.png)

MrBeardedGuy:
I believe you have specified a `set_lower_follower_count(limit = 20000)` that's why the code is preventing those users to be followed.  

Just remove that limit

cormo1990:
Thanks for the reply @MrBeardedGuy , actually i have setted as you can see in the pic attached only that limit. can you help me? 
thanks

![limit](https://user-images.githubusercontent.com/36590321/37087358-59c03f78-21fa-11e8-84b1-831511e02833.png)

SPQRBrutus:
I have same problem.

But it is not for limits or something like that but for function "follow_through_dialog". It's start scrolling down and after few scrolls instagram will stop loading. Resault is "Less users to follow than requested."

Can somebady help ?

cormo1990:
@uluQulu hey man, can you help us with that issue?
thanks a lot for the support.

MrBeardedGuy:
I noticed you have a line of code that's out of place... `set_user_interact`.... 
Check https://github.com/timgrossmann/InstaPy#followunfollowexclude-not-working

` Essentially, all the set_* methods have to be before like_by_tags or like_by_locations or unfollow. This is also implicit in all the exmples and quickstart.py`

cormo1990:
@MrBeardedGuy I try to follow your suggestion but i have the same problem...any other solution?
thanks

uluQulu:
ello @cormo1990 
You have to loop until gets desired amount
I have posted a solution similar to this[ here ](https://github.com/timgrossmann/InstaPy/issues/1493)

cormo1990:
Thanks for the answer @uluQulu, I understand what you wrote in the suggested post but honestly I m worry to wrote something wrong in the actually script...and I don't know where...can you help me please? thanks and have a good weekend dude

cormo1990:
[...]

Issue #2011, opened 2018-05-13
Apostrophe in Comments causes script to fail even when escaping the apostrophe


axlright:
## Expected Behavior

Comments with escaped apostrophe are allowed and function

## Current Behavior

Comments with escaped apostrophe cause script to exit

## InstaPy configuration

```
session.set_comments([
    u'I\'d rather not live like there isn\'t a God, then die and find out there really is - think about it #kendricklamar :fire: :collision: :fire: true dat {}',
    u'You don\'t need dollars to make you feel like you have a place in this world #kendricklamar :dollar: :money_with_wings: it\'s true {}'
    ])
```

Error
```
raceback (most recent call last):
  File "user-atxmusicvideos-hashtag-kendricklamar.py", line 48, in <module>
    session.like_by_tags(['kendricklamar'], amount=46)
  File "/Users/XXX/Documents/Code/InstaPy/instapy/instapy.py", line 1087, in like_by_tags
    self.logfolder)
  File "/Users/XXX/Documents/Code/InstaPy/instapy/comment_util.py", line 52, in comment_image
    "arguments[0].value = '" + rand_comment + " ';", comment_input[0])
  File "/Library/Python/2.7/site-packages/selenium-2.53.6-py2.7.egg/selenium/webdriver/remote/webdriver.py", line 465, in execute_script
    'args': converted_args})['value']
  File "/Library/Python/2.7/site-packages/selenium-2.53.6-py2.7.egg/selenium/webdriver/remote/webdriver.py", line 236, in execute
    self.error_handler.check_response(response)
  File "/Library/Python/2.7/site-packages/selenium-2.53.6-py2.7.egg/selenium/webdriver/remote/errorhandler.py", line 192, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.WebDriverException: Message: unknown error: Runtime.evaluate threw exception: SyntaxError: Unexpected identifier
  (Session info: chrome=66.0.3359.139)
  (Driver info: chromedriver=2.38.552518 (183d19265345f54ce39cbb94cf81ba5f15905011),platform=Mac OS X 10.13.4 x86_64)
```
I'm more than happy to just use bad grammar and remove the apostrophe. Any tips on getting the script to run with an apostrophe in the comments would be really appreciated.

You all are the bomb. Great tool. Thank you.

uluQulu:
hi @axlright,
It's an obvious result of a quote conflicting.
in **comment_util.py** file,
[...]

PR #2012, opened 2018-05-13
Multidimensional revisions & enhancements; Mainly- moved relationship bounds verification into `validate_username` definition AND redesigned `follow_user_followers` & follow_user_following` features to support relationship bounds verification & have on-the-go interactions

uluQulu:
As of @DimIsaev's second request at #1858,
**Added** **relationship bounds** _verification_ support to `follow_user_follower` & `follow_user_followers` which _costed_ the **redesign** of `follow_through_dialog` definition.
* **Now**, it gets the list of usernames in the desired amount from the dialog box and then does the rest of the procedure through `follow_by_list`.
**It has** several **benefits**:
**1**. having **relationship bounds** _verification_
**2**. having **on-the-go** interactions (_if enabled_) **rather** than following **all of the** users **and then** _starting_ interactions
**3**. **more reliable** and **efficient** control **over sleep delay**, now it takes **a break time** **per** _overall follows_ in the called feature, **rather than** break times **per user**
* **Also**, enhanced the **process of** getting _links of usernames_, **now** it will **scroll as required** and **will get links regardless of the amount** (_used_ the previously **shared** method at https://github.com/timgrossmann/InstaPy/issues/1548#issuecomment-372803884)

* **For relationship bounds** _verification_ to be _used for other features_, I **did pull** it _into a new_ definition, actually, it's **more friendly** and already **existing** definition- `validate_username`, and did some **other improves** in it, too.
* Now, `check_link` definition _does_ **only its job**- checking the links **only**, _rather than_ checking links and _verifying users_.
* Also, at `check_link`, `validate_username` and `get_relationship_counts`, added **a new layer** of **current URL** checking, which is helpful:
**1**. if the **current URL** is the same with _to be_ navigated, it **will not navigate** to it again (_saving server requests_)
**2**. if the URLs **are different**, then it **will navigate** to the required URL.
_**it is very helpful** at certain **situations**, e.g. if **relationship bounds** _verification_ is **disabled**, check_link **does not need to** navigate to the _post page_ **again**, **OR** if **relationship bounds** _verification_ is **enabled**, `follow_given_user` does **not need to** navigate to the _profile page_ **again**, cos it **already navigated** in `get_relationship_counts`_

* **Also, moved** `get_relationship_counts` from `unfollow_util.py` to `util.py` **to avoid circular dependant imports** and **it is more** like a _general utility_ to be _there_.

* **Interesting thing**: added **simulated follows**- after changing the _mechanism_ of following through dialog to following on-by-one **separately**, I realized **it would be** pretty _non-humanly_ activity if you just scroll a huge list **and do** nothing, that's why _I did add_ a **simulated following** _mechanism_, which will follow a bit of users (_max of ~4 per 100_) in _random intervals_ during the scrolling until getting _sufficient amount_ of links and it will pass those simulatedly followed users to main processer and _if any of those_ users is **not valid**, it **will unfollow** them (**simulated** users are _added_ to the main list **in random** order) immediately.
**I think** this idea of **simulated actions** _should be_ applied to other features, **as well**, _cos_ scrolling a list (_even a huge list_) and **doing nothing** seems pretty _weird_ _to be smartly handled_.

* **Enhanced** `unfollow_users` function a bit and added missing **capability** to remove unfollowed user from followed pool CSV file (_if exists_).

* **Prior break time** handling in `follow_user_followers` & `follow_user_following` based on 10 follows is **now replaced** with smarter randomized **range** of `[7, 14)` and sleep delay is **also gonna** be randomized **between** `85` & `100` percents of user-provided or default (`600`) time and these values are gonna change **periodically**.

* **Alongside**, `follow_user_followers` and `follow_user_following`, I've also added **relationship bounds** _verification_ support for `follow_by_list` feature.
* Also some modifications is **done in syntax** for more **generalized** structure and **good** looking _output_.

The **enhancements done** for `follow_user_followers` were also **recently requested** by @OGHARRY (_**relationship bounds** _verification_ support_) in #2003 and by @FranklinClinton (_**relationship bounds** _verification_ support_ and _on-the-go_ interactions) in #2009
> **I** have tested several times, if there are still, some **logical** or **mechanical** errors, please **notify**!

### DIFF
--- a/instapy/comment_util.py
+++ b/instapy/comment_util.py
@@ -51,7 +51,7 @@ def comment_image(browser, username, comments, blacklist, logger, logfolder):
             comment_input = get_comment_input(browser)
 
             browser.execute_script(
-                "arguments[0].value = '" + rand_comment + " ';", comment_input[0])
+                "arguments[0].value = arguments[1];", comment_input[0], rand_comment+' ')
             # An extra space is added here and then deleted.
             # This forces the input box to update the reactJS core
             comment_input[0].send_keys("\b")
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -417,8 +417,8 @@ class InstaPy:
         return self
 
     def set_ignore_users(self, users=None):
-        """Changes the possible restriction to users, if user who postes
-        is one of this, the image won't be liked"""
+        """Changes the possible restriction to users, if a user who posts
+        is one of these, the image won't be liked"""
         if self.aborting:
             return self
 
@@ -427,7 +427,7 @@ class InstaPy:
         return self
 
     def set_ignore_if_contains(self, words=None):
-        """ignores the don't likes if the description contains
+        """Ignores the don't likes if the description contains
         one of the given words"""
         if self.aborting:
             return self
@@ -550,7 +550,7 @@ class InstaPy:
                 # This way of iterating will prevent sleep interference between functions
                 random.shuffle(commenters)
                 for commenter in commenters[:amount] :
-                    followed = self.follow_by_list([commenter], self.follow_times, sleep_delay, interact)
+                    followed, scope = self.follow_by_list(commenter, self.follow_times, sleep_delay, interact)
                     if followed > 0:
                         followed_all += 1
                         followed_new += 1
@@ -578,13 +578,13 @@ class InstaPy:
 
     def follow_likers (self, usernames, photos_grab_amount=3, follow_likers_per_photo=3, randomize=True, sleep_delay=600, interact=False):
         """ Follows users' likers """
-        self.logger.info ("Starting to follow likers..")
+        self.logger.info("Starting to follow likers..")
 
         if not isinstance(usernames, list):
             usernames = [usernames]
 
         if photos_grab_amount>12:
-            self.logger.info ("Sorry, you can only grab likers from first 12 photos for given username now.\n")
+            self.logger.info("Sorry, you can only grab likers from first 12 photos for given username now.\n")
             photos_grab_amount = 12
 
         followed_all = 0
@@ -601,8 +601,9 @@ class InstaPy:
                 likers = users_liked(self.browser, photo_url, follow_likers_per_photo)
                 # This way of iterating will prevent sleep interference between functions
                 random.shuffle(likers)
+
                 for liker in likers[:follow_likers_per_photo] :
-                    followed = self.follow_by_list([liker], self.follow_times, sleep_delay, interact)
+                    followed, scope = self.follow_by_list(liker, self.follow_times, sleep_delay, interact)
                     if followed > 0:
                         followed_all += 1
                         followed_new += 1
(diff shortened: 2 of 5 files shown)


timgrossmann:
@uluQulu You sir are incredible!
Thank you very much for all those improvements.

DimIsaev:
@uluQulu  

very cool 
very speed  testing

uluQulu:
Thank you my friends @timgrossmann, @DimIsaev.


**Recently**, I thought of **renaming**  `follow_given_user_followers`  and `follow_given_user_following` to `get_given_user_followers` and `get_given_user_following`. :.)

**Cos** now those 2 definitions _are not following_, **instead** _returns links_ for further follows, so renaming **would be cool**. 
 **But then** **I saw that** there are already existing definitions at those names which **are used** for `interact_user_following` & `interact_user_followers` methods to return the links **from a dialog box**.


**And** I **noticed that** those guys- `get_given_user_followers` and `get_given_user_following` are doing **the same thing** as NEW `follow_given_user_followers`  and `follow_given_user_following`.
**So**, I did **remove** _prior_ `get_given_user_followers` and `get_given_user_following`. **And then** **renamed** `follow_given_user_followers`  and `follow_given_user_following` **to** `get_given_user_followers` and `get_given_user_following`.


**Now** `interact_user_followers` and `interact_user_following` methods **are using** those **smarter** definitions. I have also added **support for** _simulated follows_ from a dialog box for these _interaction based methods_.

**Also**, added **URL checker** to `get_link_for_username` function, now it **will not navigate** to the already loaded webpage **again**.

_**The result** of last commit is **enhanced versions** of `interact_user_followers` and `interact_user_following` features and **more cleaner** source code with `2` **lesser** duplicated definitions._

>**BONUS**: fixed **quote conflicting** while trying _to send a comment_ which _itself contains_ a quote (_**single** or **double**_, _**escaped** or not_) issued at #2011.

timgrossmann:
@uluQulu I tested it with some basic testing.
I think we can merge this now.

Thank you very much for this!

joanroig:
Awesome work @uluQulu !

jeremycjang:
@uluQulu Thank you for these changes and for also keeping awesome documentation! In the ```like_util.py``` the ```from .unfollow_util import get_relationship_counts``` needs to be changed to ```from .util import get_relationship_counts``` to support your changes. Small detail, but would be good to merge!

uluQulu:
Thanks guys 😁,
@jeremycjang looks like you are deep into details, cool, but now `get_relationship_counts` is a part of `validate_username` definition rather than previous `check_link` definition AND both `validate_username` and `get_relationship_counts` definitions are located inside **util.py** file which means there is no need to import it into **like_util.py** file.

It worths to notify that, the message I've included in the description comment of #1969 PR:
  
change-to-util.py
is updated as,
```erlang  
from instapy.unfollow_util import get_relationship_counts
```
is changed to
```erlang  
from instapy.util import get_relationship_counts
```
_in case of you would like to access those counts from **quickstart** scripts_

And to be honest, I could not find a place to put it into README, maybe in future there can make `Tools` sections in there to include such small but useful standalone tool.

And a good style in README belongs to this project, really, although I am very new in GitHub, I have seen lots of of project and to say, this project's documentation is simply excellent and I tried to follow that tradition :)

PR #2034, opened 2018-05-16
Single instance check

redsector72:
As I see a lot of people locking for this, I've added a check that there is no other instance of quickstart.py running with a lock file.
I suppose is cross-platform, but for sure any posix will work.

### DIFF
--- a/quickstart.py
+++ b/quickstart.py
@@ -2,6 +2,16 @@ import os
 import time
 from tempfile import gettempdir
 
+# Check that there is only one instance running
+import fcntl
+x = open('locked.txt', 'w+')
+
+try:
+    fcntl.flock(x, fcntl.LOCK_EX | fcntl.LOCK_NB)
+except:
+    print "There is another Instance alredy Running! Delete Locked.txt if this is not true!"
+    exit(0)
+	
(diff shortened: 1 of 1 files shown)


CharlesCCC:
uhhh, never had this issue before, I'm in doubt if we should merge this.

sionking:
Great I really need it!
Can you explain how it works? When will the  lock release?

CharlesCCC:
@sionking I would like to know under what scenario this error could happening ?

redsector72:
Is not a patch for an error.
I've simply added a file lock check at the start.
If the script is already running, then the file is locked and the script quit.

In my system I use bash, but I have developed a python equivalent. Maybe it can be transformed in a function to be called optionally

CharlesCCC:
@redsector72 Gotcha, I would agree with you on add this as optional. for those people who had the issue, they can turn this `on`, otherwise default to `off`. thank you for the effort. 

In addition, I see you import additional library, please make sure add this to the install script as well.

sionking:
I think this feature is not multi user compatible.
Can it be not in quickstart (or will it get into race)?

Issue #1862, opened 2018-04-16
Theme: disable the info bar in chrome


sionking:
Just add it to the chrome options:
chrome_options.add_argument("--disable-infobars")

uluQulu:
Hey @sionking,
It is great I have even thought of removing that annoying banner, too and tried your suggestion which works perfetly
But there is a **drawback** using it (esp. _for mostly Chrome browser users_), **cos** once you have more than 1 Chrome windows opened, **that banner** in that selenium-controller window **reminds you** of _"Hey, it is controlled by selenium! DON'T touch this window! (otherwise, in a new tab opened or navigation to other webpage, selenium will most probably crash out of not finding the requested element)"_

>But I have recently tried latest Firefox browser started by selenium, and I see that **that banner** is shown right in the adress bar. Chrome should also do it, cos that banner captures a big space and sometimes it really is annoying...

PR #913, opened 2017-11-13
Record activity generate statistic (sqlite)

converge:
Hey guys,

the CSV approach for this functionality has caused an issue for Windows users and a workaround had to be done to "fix" it until we have a better solution.

I wrote the update_activity() function again using sqlite. Can you guys test it and leave a review ?

* docs/erm.mwb (MySQL workbench ERM diagram)
* db/instapy.py (sqlite DB)

it fixes #873

PR #2051, opened 2018-05-18
Optional ignore check image; optional use profile path for chrome; re…

sionking:
List of new/small improvements:

- Remove infobars from chrome.
- Mute sound in chrome.
- Optional ignore check image from Clarifi:
   e.g. If we want t to ignore dogs but not if a person is also in the picture.
  (if not used the list is empty so no changes needed)
- Optional provide image url for image check Clarifi (e.g. profile picture).
- Optional use profile path for chrome and not only for Firefox.
- log following of profile

I run one test but please check it as it is not my main repo.

SCK.

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -533,7 +533,7 @@ class InstaPy:
             self.use_clarifai = False
         elif tags:
             self.clarifai_img_tags.append((tags, comment, comments))
-            self.clarifai_img_tags_skip.append((tags_skip, comment, comments))
+            self.clarifai_img_tags_skip = tags_skip
 
 
         return self
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -32,6 +32,7 @@ from .like_util import get_links_for_username
 from .login_util import login_user
 from .print_log_writer import log_follower_num
 from .settings import Settings
+from .print_log_writer import log_following_num
 from .time_util import sleep
 from .time_util import set_sleep_percentage
 from .util import get_active_users
@@ -66,6 +67,7 @@ class InstaPy:
                  nogui=False,
                  selenium_local_session=True,
                  use_firefox=False,
+                 browser_profile_path=None,
                  page_delay=25,
                  show_logs=True,
                  headless_browser=False,
@@ -98,7 +100,7 @@ class InstaPy:
         self.page_delay = page_delay
         self.switch_language = True
         self.use_firefox = use_firefox
-        self.firefox_profile_path = None
+        self.browser_profile_path = browser_profile_path
 
         self.do_comment = False
         self.comment_percentage = 0
@@ -109,6 +111,7 @@ class InstaPy:
         self.followed = 0
         self.liked_img = 0
         self.already_liked = 0
+        self.already_Visited = 0
         self.inap_img = 0
         self.commented = 0
         self.followed_by = 0
@@ -137,6 +140,7 @@ class InstaPy:
         self.use_clarifai = False
         self.clarifai_api_key = None
         self.clarifai_img_tags = []
+        self.clarifai_img_tags_skip = []
         self.clarifai_full_match = False
         
         self.potency_ratio = 1.3466
@@ -210,9 +214,9 @@ class InstaPy:
             return self
 
         if self.use_firefox:
-            if self.firefox_profile_path is not None:
+            if self.browser_profile_path is not None:
                 firefox_profile = webdriver.FirefoxProfile(
-                    self.firefox_profile_path)
+                    self.browser_profile_path)
             else:
                 firefox_profile = webdriver.FirefoxProfile()
 
@@ -236,6 +240,8 @@ class InstaPy:
         else:
             chromedriver_location = Settings.chromedriver_location
             chrome_options = Options()
+            chrome_options.add_argument("--disable-infobars")
+            chrome_options.add_argument("--mute-audio")
             chrome_options.add_argument('--dns-prefetch-disable')
             chrome_options.add_argument('--no-sandbox')
             chrome_options.add_argument('--lang=en-US')
@@ -263,6 +269,9 @@ class InstaPy:
             # add proxy extension
             if self.proxy_chrome_extension and not self.headless_browser:
                 chrome_options.add_extension(self.proxy_chrome_extension)
+
+            if self.browser_profile_path is not None:
+                chrome_options.add_argument('user-data-dir={}'.format(self.browser_profile_path))
 
             chrome_prefs = {
                 'intl.accept_languages': 'en-US'
@@ -325,6 +334,7 @@ class InstaPy:
             self.logger.info('Logged in successfully!')
 
         self.followed_by = log_follower_num(self.browser, self.username, self.logfolder)
+        self.following_num = log_following_num(self.browser, self.username, self.logfolder)
 
         return self
 
@@ -458,8 +468,8 @@ class InstaPy:
         if self.aborting:
             return self
 
-        if os.name == 'nt':
-            raise InstaPyError('Clarifai is not supported on Windows')
+        #if os.name == 'nt':
+        #    raise InstaPyError('Clarifai is not supported on Windows')
 
         self.use_clarifai = enabled
 
@@ -514,7 +524,7 @@ class InstaPy:
         self.smart_hashtags = list(set(self.smart_hashtags))
         return self
 
-    def clarifai_check_img_for(self, tags=None, comment=False, comments=None):
+    def clarifai_check_img_for(self, tags=None, tags_skip=None, comment=False, comments=None):
         """Defines the tags, the images should be checked for"""
         if self.aborting:
             return self
@@ -523,6 +533,8 @@ class InstaPy:
             self.use_clarifai = False
         elif tags:
             self.clarifai_img_tags.append((tags, comment, comments))
+            self.clarifai_img_tags_skip.append((tags_skip, comment, comments))
+
 
         return self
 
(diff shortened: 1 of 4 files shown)


instapy/print_log_writer.py
```diff
@@ -27,6 +30,31 @@ def log_follower_num(browser, username, logfolder):
...
+        following_num = browser.execute_script(
+            "return window._sharedData.""entry_data.ProfilePage[0]."
+            "graphql.user.edge_follow.count")
+    except WebDriverException:
+        browser.execute_script("location.reload()")
+        sleep(10)
+        following_num = browser.execute_script(
+            "return window._sharedData.""entry_data.ProfilePage[0]."
+            "graphql.user.edge_follow.count")
+        # raise exception if number can't be read second time
```
uluQulu: have you forgotten to catch the last possible handle?
cos since `following_num`  is not initialized and if not catched it will return 'not defined' error

uluQulu:
Looks great, @sionking,
Regarding the infobar removal, I would add the same opinion as in https://github.com/timgrossmann/InstaPy/issues/1862#issuecomment-382003325
infobar

Also, it would be great if `followed_by` and `following` numbers could be stored in same file for more clear `log` folder

Also, I see these lines commented out,
```erlang
if os.name == 'nt': 
    raise InstaPyError('Clarifai is not supported on Windows') 
```
It's a topic of discussion, cos I also always comment out them and use clarifai successfully.  
But what is the problem with clarifai for Windows OS?
Is it not supported in some version(s) or some architecture(s)? 
 Anybody knows the reason of disabling it?
>In installation instructions for Windows, it says to comment out `from .clarifai_util import check_image`, too and this issue must be resolved correctly

instapy/print_log_writer.py
sionking: Thanks for code review!
no, I have not forgotten.
Basically if this value does not present - you might not logged in or other bed scenarios.
So I prefer to receive the exception and handle it.

sionking:
Lets make this infobar removal optional? maybe it can somehow be detectable (I don't think so).
I am using latest clarifai_util on windows 10 with no issues. ( it is very simple API...)

"Also, it would be great if followed_by and following numbers could be stored in same file for more clear log folder" same folder .. same file is just confusion and hard to parse(many do parse this file already for graphing) !

ebolamonk3y:
selenium.common.exceptions.WebDriverException: Message: unknown error: Cannot re ad property '0' of undefined

Using previous version and working fine.

sionking:
@ebolamonk3y hi  can you provide line number ?
I am pretty sure that it is the following_num  not read right and usually instapy return None but I prefer to have the exception sinceyou may got network error / login error ....

instapy/print_log_writer.py
sionking: I think the version for instapy should include the none to avoid not defined (again it is a problem that we avoid instead of handling)

uluQulu: It happens either out of temporary outage from server side or after structure change (can be fixed).
 And those are the only options available to get data: `entry_data` & page elements, if both fails there is nothing to do.

sionking:
@ebolamonk3y this one fixed

uluQulu:
@sionking 
>Lets make this infobar removal optional? maybe it can somehow be detectable (I don't think so).

It's a property of browser itself and can't be accessed by a webpage. Well **predictably**, there can be other ways to detect it but it's a nasty idea to be implemented that's what I think.
Optional could be fine to add its boolean parameter- `remove_infobar` with the default value of `False` inside `__init__` will be pretty.

>I am using latest clarifai_util on windows 10 with no issues. ( it is very simple API...)

@timgrossmann, @CharlesCCC 
**Maybe** you know why Clarifai is recommended to be disabled on Windows platform?
I couldn't understand the problem and it works pretty well on my Windows 10 machine and the same for @sionking.
**Also**, the warning `"Record activity is not working on Windows."`
What is record activity and why it is pretendably not working on that platform.
If you explain those situations, maybe we- Windows users could find solutions rather than ignoring it.

>"Also, it would be great if followed_by and following numbers could be stored in same file for more clear log folder" same folder .. same file is just confusion and hard to parse(many do parse this file already for graphing) !

I like simplicity and minimalism, and sometimes improvements cost breaking things :)

_BTW works on `StaleElementReferenceException` is amazing and in future there seems to be much more problems raised by it._

sionking:
@uluQulu thanks.
right the issue with StaleElementReferenceException happens to me a lot so I deep dig into it.

timgrossmann:
@sionking @uluQulu Unfortunately I cannot test the windows clarifai problem. It sounds like there were some people who encountered problems on windows but I never had the same problem on my windows pc.
Looks like it might only be a OS thing that only affects older windows versions?

timgrossmann:
@sionking Those improvements sound awesome, hopefully we can figure out how to merge them safely as soon as possible!

uluQulu:
I see @timgrossmann 
I looked back some issues and couldn't find that particular reason of the possible misbehavior by `clarifai` module but I will look again.
And it is possible that it is platform version specific where we can check the version easily and disable that not-working module from usage rather than disabling on every Windows and brand-new version 10 which most of people use now.
Further, I am sure that if the root of the issue for that not-working platform version is found, the solution can also be produced to make it work, cos as @sionking said, clarifai has a simple API and at which we only use the basic model.

Also for the warning `"Record activity is not working on Windows."`, have you any idea @timgrossmann about it?
That warning was loaded when I met Instapy this December and for a long time I thought it really is a problem but actually it is kind of a mystic :D
But old users would know about the born of that problem, also I did check issues and couldn't find a one which explained the specific reason of that possible issue.

Cheers >:

timgrossmann:
@uluQulu Well, this `"Record activity is not working on Windows."` seems to only be a legacy problem that might as well has been fixed now.
There were some changes that someone made and could not make it work on a windows laptop so he disabled it.

uluQulu:
@timgrossmann 
It's clear. But I still don't understand what is `Record activity`.
Does that refer to logging activity (**general.log**) blacklist campaign (**blacklist.csv**) or sth else?
Also, if something does not run well on Windows or in its some version, there is no any disabled state related to so-called Record activity and actually in source code there exists no similar line to turn off something for Windows machines about which I am very sure.

So at this state, I can clearly say that, for now that _warning_ is on the fly and has no effect. There are 2 options:
1. Make sure **it** really is to be avoidable in Widnows and either turn it off completely by checking platform or turn it off on some platform versions.
2. Just remove that warning cos it has no any appliance.

>I got a guess that it recalls to **blacklist.csv** but I'm not sure :>


Solving that situation for good is very important cos users (_mostly new users_) do panic that,
_"Hey! Is there something wrong?"_
 _"Heh! I missed a perfect piece- Record activity, by having a Windows box :("_
:D
and from my experience from here and there helping out people, **almost** 4/5 of users are running MS Windows OS who see that warning in every start.

Also, having that warning at the very start of InstaPy, makes another point,
_"Hmm..InstaPy has a problem"_ :D

Good luck!

timgrossmann:
@uluQulu As far as I know it's part of the update to the sql db in the `update_activity` methood in

https://github.com/timgrossmann/InstaPy/blob/afdf91122afb01f9fed08958eed1d35457caba95/instapy/util.py#L128

But I think it dates back to the implementation of the blacklist thing or alternatively the `campaign` feature. 

**We should really start creating and maintaining a changelog**

The part about the error and people thinking that there is a problem is true.
I would prefer to have some more experience if it really only is some legacy thing and there are no problems before simply ignoring to give out a warning.

uluQulu:
Thought about searching again the root of that problem to refine the situation. 
Last time I checked it was the time I first installed InstaPy and wondered what was **Record activity** and but this time I grabbed the cause of it. And well it's kind of a missing change with a huge impact!

* In #853 @converge [added](https://github.com/timgrossmann/InstaPy/pull/853/files#diff-870c589a51f82fb50caf6a78df9e4273) a great feature- `update_activity` (**_really great_**) with a CSV approach which was relied _atomic write_ (_kind of_) using the `tempfile` module.

* In #875 he did find out out a problem with using the CSV method [relied on](https://github.com/timgrossmann/InstaPy/pull/875/files) _atomic write_ on Windows platform AND for that reason, he added that warning- `Record activity is not working on Windows` message and **disabled** the `update_actvitiy` if the `os.name=='nt'` (_Windows_).
The original PR description is as follows:
"_There is a problem on windows when we try to delete a temporary file. To fix this, the solution is ugly, needs extra win32 libs. I think it's a good idea for now to not provide this functionality to windows users._"

* In #913 he changed the design of `update_activity` from scratch to use a DB by `sqlite3` module. :)
Now, problem is clear, I guess...
That problematic CSV _atomic write_ method was completely replaced with DB which didn't need `tempfile` module and is fully (_almost_) supported in Windows and other platforms.

And at and after #913, it was [forgotten](https://github.com/timgrossmann/InstaPy/pull/913/files) to remove the line of famous warning `Record activity is not working on Windows.`

>Looks like Windows people will chill now without that famous warning :)

Cheers xD (_BTW personally thank again to @converge for that `update_activity` feature_)

sionking:
As I said there is a repo https://github.com/untitaker/python-atomicwrites we should use.

sionking:
Can I commit ?

instapy/instapy.py
```diff
@@ -458,8 +468,8 @@ def set_use_clarifai(self, enabled=False, api_key=None, full_match=False):
         if self.aborting:
             return self
 
-        if os.name == 'nt':
-            raise InstaPyError('Clarifai is not supported on Windows')
+        #if os.name == 'nt':
+        #    raise InstaPyError('Clarifai is not supported on Windows')
```
uluQulu, line 473: This commenting out line will pop the possible problem again with Windows `Clarifai` users **and** at that time, it must be solved quickly otherwise this line will have to be reverted back.

Also. There are a few lines in Windows installation instructions which says to comment out `Clarifai` imports which must also be resolved after this change to avoid confusion.

PR #2080, opened 2018-05-23
ISSUE-2079: set custom BASE_DIR for instapy settings

jaybocc2:
if you install instapy through pip and don't have the ability or desire to edit settings.py directly you only need set environment var for `BASE_DIR` before running your instapy (you'll need copies of db and proper db structure in base_dir)

### DIFF
--- a/instapy/settings.py
+++ b/instapy/settings.py
@@ -5,6 +5,7 @@ BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
 
 class Settings:
+    BASE_DIR=os.environ.get('BASE_DIR', BASE_DIR)
     log_location = os.path.join(BASE_DIR, 'logs')
     database_location = os.path.join(BASE_DIR, 'db', 'instapy.db')
     chromedriver_location = os.path.join(BASE_DIR, 'assets', 'chromedriver')


timgrossmann:
@jaybocc2 Thank you very much for this improvement.

@uluQulu @sionking @CharlesCCC Could you guys double check this real quick, looks good from my site.

sionking:
Thanks.

CharlesCCC:
change looks fine to me, but can you please add some wiki instruction on how/where/when to use this ?

uluQulu:
Hi guys
Finally got time to share my review 🙋🏼‍♂️
@jaybocc2 I have to say that either you are about to use a different path for data files or have a `pip` situation, if you would like, you can use an already existing easier method to achieve that :)

While back @lucanello wrote an amazing program-wide accessible `Settings` class with which you can set your desired data locations right inside your **quickstart** script like this,
```python
""" Such a quickstart script"""
##import Settings class
from instapy import Settings

##set custom locations
Settings.database_location = './db/instapy_guliyevshahriyar.db'
Settings.browser_location = '/path/to/chromedriver'
```
**This is more** appropriate cos, considering you will have _multiple_ InstaPy installations & **quickstart** scripts you can set custom data locations for them (_although, webdriver has no effect, but e.g. DB file **instapy.db** should be unique for each different kind of quickstart script_ (_e.g. of different users..._))
_As for simple  `BASE_DIR`, you must have set those different locations **every time** with another environmental variable (`BASE_DIR2`, `BASE_DIR3` sth like that,..) in order to have unique data locations which is a typical mess._
>**It is** just 3 lines to **add** in you quickstart script to be added only once, rather than 1 line of setting system-wide environmental variable and also having your system settings get bloated xD

**Also it has** difficulties if you think the **user side**, e.g., not all users have _administrative privileges_ to set environmental variables OR a little of mistake can mess their other system environmental variables.

@CharlesCCC 
**It's easier** in UNIX-based systems to set it but **for Windows** it is a bit of trouble _making it work_ xD
_E.g., to add a permanent environmental variable, one can use `SETX` command_
```cmd
SETX BASE_DIR C:\Users\Qulu\BASE_DIR
```
**It will make** a new environmental variable named `BASEDIR` to be used system-wide (it will work in the next opened consoles but not in current one).
>And SETX command has a limit of max 1024 char allowed for environmental variable (_well, not much big problem in this case..._)

#### **What I RECOMMEND is** to add an **understandable documentation** for the _usage of `Settings` class_ and setting data files' location and use it for the **efficiency**, **ease** of modification and **portability** of the program.


BTW @timgrossmann
I have _always requested_ up-to-date files GitHub when downloaded the program, but it could be a lot easier to use `pip`, in that matter, I have some **questions**:
+ Is the files downloaded from `pip` is identical with the up-to-date files in the GitHub repo?
+ From the **Issues** by users, `pip` method complains about `2.34` chromedriver to be used, so why it is not updated?
>It would be great to make `pip` support fully functional and up-to-date.

Good luck!

timgrossmann:
@uluQulu Thank you for the deep insights into your opinion, it sounds very reasonable and clear. I like the approach you're proposing.

Regarding your questions, I have to admit that I have no idea about how to make a tool like this "easily" available through pip...
The problem I see here, even if we make sure it's also clear that you're able to download the tool directly as a zip, is that you have to "put the chromedriver" somewhere and link it. That can be a huge pain for a lot of people that don't know where the tool goes if you do `pip install .....` since there won't be a folder in your download directory. 
I'd be glad if we could find a solution for this, absolutely glad if it would be as easy as creating a single file on the Desktop and simply doing `from instapy import InstaPy` in there and everything would be resolved automatically.
Do you think you could get that working @uluQulu ? Seems like so 😉 

There is a repo called instapy on pip but it has nothing to do with this tool. If you do the `pip install git+....` thing, then it takes the tool from github with the current version, so yes that would be the latest one. 

Thanks guys for all the help and reviews, that's pure gold!
You make this tool great!

uluQulu:
It's a pleasure, @timgrossmann 
It's clear, gotta take a good care to `pip` installations.

And I am grateful about the assignment and I wish I could do it right away but pity that I have very less knowledge about `pip's operational mechanism AND these days have not much time to conduct a research. 

_But I keep it in mind, whenever gets a suitable condition, will try to have a look at that very essential installation & execution specific issue._
>And I have some initial ideas about making it work which would cost a few design changes, if someone else is interested, I would tell.

Cheers

jaybocc2:
@uluQulu  Thanks i did not realize the Settings class was publicly available.  That solution works great, even in a `pip install git+` installation and is a good bit more elegant, though i think it would also be easy enough to run with many different `BASE_DIR`s that is not necessary for my use case.  Closing this PR.

jaybocc2:
Solution:

```
from instapy import InstaPy
from instapy import Settings

base_dir = '/home/jaybocc2/myinsta'
Settings.log_location = os.path.join(base_dir, 'logs')
Settings.database_location = os.path.join(base_dir, 'db', 'instapy.db')
Settings.chromedriver_location = os.path.join(base_dir, 'assets', 'chromedriver')
session = InstaPy(...)
...
session.end()
```

Issue #2076, opened 2018-05-22
Follow_User_Followers not interacting with users


jeremycjang:
After the major code update that was made to change how ```follow_user_followers``` and ```follow_user_following``` fundamentally function, I have been unable to get the program to interact with users as they are followed. I have currently concluded that the updated program is not pulling from the ```interact``` function correctly as I have tested this with this code (the media parameters not working is being tracked in a separate issue):

```
session.set_do_comment(enabled=True, percentage=100)
session.set_comments(['comment1', 'comment2', 'comment3'], media='Photo')
session.set_comments(['comment4', 'comment5', 'comment6'], media='Video')
session.set_user_interact(amount=1, randomize=False, percentage=100)

session.follow_user_followers(['user'], amount=5, randomize=False, interact=True, sleep_delay=600)

```

I set the comment and interact percentages to 100% so that the program would like/comment on every user photo during the test run and it did not comment on any while following.

@uluQulu were there any other missing parameters related to this issue that should be added to make the code interact with users?

Thank you!

uluQulu:
@jeremycjang 
Earlier those changes, those features could not have on-the-go interactions but now they can.
I have used `interact_by_users` method to do the standalone interactions for them.
And for particularly `follow_user_followers` feature, it will consider to interact only after following a user.
So you have 2 interaction types available: _likes_ and _comments_

Fine, and in `interact_by_users` feature, it does try to comment only after liking a picture which is I like and which is you are missing.
You have not defined `set_do_like` in your quickstart script in order to handle _liking_ and if it is not defined, `self.do_like` stays in disabled state, so it will not like any picture, that means will not also comment, too regardless of the commenting is disabled or enabled.
What you could  do is defining `set_do_like` in your quickstart script.


>I set the comment and interact percentages to 100% so that the program would like/comment on every user photo during the test run and it did not comment on any while following.

And from those words, I think you also want to like, so adding the missing `set_do_like` will be okay.
_I said this cos there are people who would want to comment without liking and it is not how `interact_by_users` work and that kind of commenting is pretty annoying and inefficient, that's why I specifically chose `interact_by_users`._

Read https://github.com/timgrossmann/InstaPy/issues/2019#issuecomment-388889341 for more.

jeremycjang:
@uluQulu That makes sense; I think previously the ```do.like``` function did not exist, or at the least I had never used it while using the legacy ```follow_user_followers```.

[...]

Issue #2074, opened 2018-05-21
the web page instagram does not load


CamiloB:
## Expected Behavior
when i go to load the web page of instagram is in white 

## Current Behavior
this appear on the console 

`Traceback (most recent call last):
  File "/home/ubuntu/InstaPy/sagrada.py", line 30, in like
    session.set_dont_unfollow_active_users(enabled=True, posts=5)
  File "/home/ubuntu/InstaPy/instapy/instapy.py", line 1740, in set_dont_unfollow_active_users
    self.logger)
  File "/home/ubuntu/InstaPy/instapy/util.py", line 118, in get_active_users
    "//span[contains(@class,'_t98z6')]//span").text)
  File "/usr/local/lib/python2.7/dist-packages/selenium-2.53.6-py2.7.egg/selenium/webdriver/remote/webdriver.py", line 293, in find_element_by_xpath
    return self.find_element(by=By.XPATH, value=xpath)
  File "/usr/local/lib/python2.7/dist-packages/selenium-2.53.6-py2.7.egg/selenium/webdriver/remote/webdriver.py", line 752, in find_element
    'value': value})['value']
  File "/usr/local/lib/python2.7/dist-packages/selenium-2.53.6-py2.7.egg/selenium/webdriver/remote/webdriver.py", line 236, in execute
    self.error_handler.check_response(response)
  File "/usr/local/lib/python2.7/dist-packages/selenium-2.53.6-py2.7.egg/selenium/webdriver/remote/errorhandler.py", line 192, in check_response
    raise exception_class(message, screen, stacktrace)
NoSuchElementException: Message: no such element: Unable to locate element: {"method":"xpath","selector":"//span[contains(@class,'_t98z6')]//span"}
  (Session info: chrome=65.0.3325.181)
  (Driver info: chromedriver=2.36.540471 (9c759b81a907e70363c6312294d30b6ccccc2752),platform=Linux 3.13.0-143-generic x86_64)
`

## InstaPy configuration

`from instapy import InstaPy
import schedule
import time

insta_username = 
insta_password = 

def like():

    session = InstaPy(username=insta_username,
                      password=insta_password,
                      headless_browser=False,
                      bypass_suspicious_attempt=False,
                      multi_logs=True)
    session.login()

    session.set_do_comment(True, percentage=50)
    session.set_do_like(True, percentage=70)
    session.set_do_follow(True, percentage=70)
    session.set_upper_follower_count(limit=12500)
[...]

Issue #1732, opened 2018-03-24
Follow followers but not follow user that already follow me


cormo1990:
Hello friend, how can i prevent to follow user that already follow me?
thanks for the support.
I hope that everything is explain well

cormo1990:
UP

@uluQulu help please

uluQulu:
hi @cormo1990 
1. in which feature(s) you have intended to make changes? (_e.g. `follow_user_followers`_)
2. in the opened dialog or page, if somebody follows you (_and that you do not follow her/him_), does IG show you "**Follow Back**" or just "**Follow**" button?

cormo1990:
hi @uluQulu the answers to your questions are:

1. Follow_user_followers

2. IG show **Follow Back** button

uluQulu:
Great! @cormo1990 
_2nd step_ is very important, if you really see _Follow Back_ button it will be so much easier for writing the desired change
To make sure, please share the screenshot of that button in the same dialog box. (_to guarrantee that we are gonna modify the right way_)

> I think in the future every user will see _Follow Back_ button in web browsers, too, but at this moment it is possibly in beta test. _E.g._ I still see _Follow_ button in the same dialog box, whilst some buddy from Italy see _Follow Back_ everywhere...

cormo1990:
@uluQulu I am italian, and after what you wrote in the previous answer, i did some checks with changing browsers and PCs and it is incredible...i don't understand why but not in all browser and pc i can see that button of **follow back**...so i suppose that edit is helpful just for a few user...the other changes is so difficult?

uluQulu:
Super job! @cormo1990 
I meant the simplicity of the modifications, none of them is difficult
Well I thought again and realized that we can make it possible with the same set of accuracy using the other method. And will even make it much faster than the prior "Follow Back" method by just adding a new single variable inside InstaPy class ^^;

cormo1990:
@uluQulu your job is the hardest bro! thanks again for support all the community...

uluQulu:
_i've caught seasonal grip that's why it took lots of time to test within ill state, but it is ready_
you will edit several places in **3** files, @cormo1990 (_mostly small changes 🔩_ )
**File** 1: **instapy.py**
**#1** **import** new function `get_watchers` near other imports
```python
from .util import validate_username
from .util import get_watchers   ##you will add this new line
```
**#2** **add** a new parameter to `follow_user_followers` method, called `non_watchers`
```python
[...]

PR #2042, opened 2018-05-18
Fix `following_link` layout change in `unfollow_users` and add graphql approach to get `likes_count` and `possible_posts` values

uluQulu:
* Now `like_by_locations` feature will also use _possible posts_ amount which will provide great accuracy while getting links (_solves https://github.com/timgrossmann/InstaPy/issues/1825#issuecomment-380301233_).

* Adds graphql approach of getting values for `possible_posts` and `likes_count` (_their values are numbers_) which sometimes have troubles getting by elements resulting an empty string either is a result of a layout change or something related to the visibility setting such as, `display:block` that I could not determine and it happened in some users only and also is not persistent, yet which will definitely be solved by using this graphql approach, hopefully.  
Mentioned in https://github.com/timgrossmann/InstaPy/issues/1924#issuecomment-385247307.  
Affects #1924, #1957, #2032, #2039, #2047

* Adds correct XPATH of `following_link` to be used with `unfollow_users` in `onlyInstapyFolloed=False` **and** `onlyNotFollowMe=False` condition.  
Mentioned in https://github.com/timgrossmann/InstaPy/issues/2031#issuecomment-389608594.  
Affects #2031, #1977, #2024, #2026, #2044

PR #2086, opened 2018-05-23
Take cautions to provide healthy followRestriction.json usage. [PROPOSED to use DB]

uluQulu:
There is a situation when a user presses `CTRL+C` (or `CTRL+Break`) in order to terminate the program.
**Well** it is what most of us do. But there is a problem.

#### INSPECTION
**followRestriction.json** is being dumped to the local file at `end` method which happens right after termination (**procedural** or **forcibly** (_e.g. by a user_)). Well, the problem raises here:
**At any** case, if that _JSON_ is not properly written (e.g. missing some double-quote or a comma and a closing curly brase) it will not be able to load in the next run cos it is corrupt.

**This problem has been** around so long and I had suggested a solution to fix it in many posts, (e.g., here https://github.com/timgrossmann/InstaPy/issues/1732#issuecomment-383390727)
But that solution **is to handle** that error, but not solve.

**Having it raised** by some guys lately and most of time they just empty that _JSON_ instead of fixing corrupt element, so I decided to solve it definitely.

#### INITIAL ANALYSIS
First I thought about implementing a **_JSON_ fixer** definition to do the job automatically. But this solution has drawbacks.
Cos once your data **gets corrupted**, even after you fix it to work, you lose old data which couldn't been dumped to the _JSON_.

#### SOLUTION
So going **into the** root of the process. I have thought about this:
**When a** user externally interrupts the program, handle that interruption and let _JSON_ to dump and after dump finished successfully, release the handle.

**This** solution is so **robust** cos I have used `signal` module to catch **SIGINT** (_external interrupt signal_) and passed it to **SIG_IGN** (_ignore signal_). That's where, during the process of dumping, if user presses `CTRL+C` million times it will ignore it and continue to dump, even though user keeps `CTRL+C` pressed and not released :D

**It changes** to have a healthy _JSON_ file after all and nothing more!
And also included a condition to check if the _JSON_ file is changed any (_if any user was followed_), if there is no any user followed, then it will skip dumping :)

#### DEBUG DATA 
To see results, I did test 2 scernarios: **middle** and **huge**
##### MIDDLE TEST: auto-generated _JSON_ file with 68595 unique elements
**test1**
load time: '0.18832683563232422' seconds
files are identical  |  compare time passed: 0.019001007080078125
**test 2**
load_time1: 0.16599178314208984  |  load_time2: 0.1660010814666748
Successfully dumped newly followed users in 1.6 seconds  ~ 7 new users

##### HUGE TEST: auto-generated _JSON_ file with 999990 (_~nearly 1 million_) unique elements
**test 1**
len folRes: 999990  |  load_time1: 3.670952558517456  |  load_time2: 3.618953227996826
**test 2**
len folRes: 999990  |  load_time1: 3.5599496364593506  |  load_time2: 3.584954023361206
check_time: 0.6769866943359375
**test 3**
len folRes: 999990  |  load_time1: 4.496942758560181  |  load_time2: 4.072948932647705
Successfully dumped newly followed users in 42.18 seconds  ~ 7 new users
**test 4**
len folRes: 999997  |  load_time1: 3.6659462451934814  |  load_time2: 4.09894871711731
Successfully dumped newly followed users in 23.49 seconds  ~ 1 new users
**test 5**
len folRes: 999998  |  load_time1: 3.535956382751465  |  load_time2: 3.524956226348877
Successfully dumped newly followed users in 23.41 seconds  ~ 1 new users.
**test 6**
load_time1: 3.5369532108306885  |  load_time2: 3.6459503173828125
Successfully dumped newly followed users in 23.31 seconds  ~ 1 new users.
**test 7** 
load_time1: 3.5379552841186523  |  load_time2: 3.623953342437744
Successfully dumped newly followed users in 23.07 seconds  ~ 8 new users.

#### FINAL WORDS
**The only thing changed** is making user wait _JSON_ file to dump and not allow any interruptions during that time. Else than that change I haven't touched any other mechanism.

_I have attached those middle and huge _JSON_ files used in tests, feel free to have a look (_beware of HUGE file, it has 1 million elements which can crash your text editor :D_)
[followRestriction JSONs](https://github.com/timgrossmann/InstaPy/files/2031405/followRestriction_JSON.zip)_

**This PR also includes** fixes shared in #2074 and #2076.

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -173,10 +173,6 @@ class InstaPy:
         if selenium_local_session:
             self.set_selenium_local_session()
 
-        if os.name == 'nt':
-            error_msg = ('Sorry, Record Activity is not working on Windows. '
-                         'We\'re working to fix this soon!')
-            self.logger.warning(error_msg)
 
     def get_instapy_logger(self, show_logs):
         """
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1,5 +1,4 @@
 """OS Modules environ method to get the setup vars from the Environment"""
-import time
 import csv
 import json
 import logging
@@ -39,7 +38,6 @@ from .time_util import set_sleep_percentage
 from .util import get_active_users
 from .util import validate_username
 from .util import web_adress_navigator
-from .util import interruption_handler
 from .unfollow_util import get_given_user_followers
 from .unfollow_util import get_given_user_following
 from .unfollow_util import unfollow
@@ -122,7 +120,6 @@ class InstaPy:
         self.not_valid_users = 0
 
         self.follow_restrict = load_follow_restriction(self.logfolder)
-        self.follow_restrict_init = load_follow_restriction(self.logfolder)
 
         self.follow_times = 1
         self.do_follow = False
@@ -2406,19 +2403,7 @@ class InstaPy:
 
     def end(self):
         """Closes the current session"""
-        # Dump followed users to followRestriction.json file at program exit.
-        self.logger.warning("Please wait a second to dump followed users to followRestriction.json :>")
-        with interruption_handler():
-            if self.follow_restrict != self.follow_restrict_init:
-                 ts_before_dump = time.time()
-                 dump_follow_restriction(self.follow_restrict, self.logfolder)
-                 ts_after_dump = time.time()
-                 self.logger.info("Successfully dumped newly followed users in {} seconds"
-                            "  ~ {} new users.\n".format(
-                                float("{0:.2f}".format(ts_after_dump-ts_before_dump)),
-                                    len(self.follow_restrict)-len(self.follow_restrict_init)))
-            else:
-                self.logger.info("Nobody was followed during the session.\n")
+        dump_follow_restriction(self.follow_restrict, self.logfolder)
 
         try:
             self.browser.delete_all_cookies()
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1,4 +1,5 @@
 """OS Modules environ method to get the setup vars from the Environment"""
+import time
 import csv
 import json
 import logging
@@ -37,6 +38,7 @@ from .time_util import set_sleep_percentage
 from .util import get_active_users
 from .util import validate_username
 from .util import web_adress_navigator
+from .util import interruption_handler
 from .unfollow_util import get_given_user_followers
 from .unfollow_util import get_given_user_following
 from .unfollow_util import unfollow
@@ -117,6 +119,8 @@ class InstaPy:
         self.not_valid_users = 0
 
         self.follow_restrict = load_follow_restriction(self.logfolder)
+        self.follow_restrict_init = load_follow_restriction(self.logfolder)
+
         self.follow_times = 1
         self.do_follow = False
         self.follow_percentage = 0
@@ -2381,7 +2385,20 @@ class InstaPy:
 
     def end(self):
         """Closes the current session"""
-        dump_follow_restriction(self.follow_restrict, self.logfolder)
+        # Dump followed users to followRestriction.json file at program exit.
+        self.logger.warning("Please wait a second to dump followed users to followRestriction.json :>")
+        with interruption_handler():
+            if self.follow_restrict != self.follow_restrict_init:
+                 ts_before_dump = time.time()
+                 dump_follow_restriction(self.follow_restrict, self.logfolder)
+                 ts_after_dump = time.time()
+                 self.logger.info("Successfully dumped newly followed users in {} seconds"
+                            "  ~ {} new users.\n".format(
+                                float("{0:.2f}".format(ts_after_dump-ts_before_dump)),
+                                    len(self.follow_restrict)-len(self.follow_restrict_init)))
+            else:
+                self.logger.info("Nobody was followed during the session.\n")
+
         try:
             self.browser.delete_all_cookies()
             self.browser.quit()
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -365,7 +365,7 @@ def get_links_for_username(browser,
     sleep(2)
 
     # Get Links
-    main_elem = browser.find_element_by_tag_name('main')
+    main_elem = browser.find_element_by_tag_name('article')
     link_elems = main_elem.find_elements_by_tag_name('a')
     total_links = len(link_elems)
     # Check there is at least one link
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -742,8 +742,8 @@ def get_given_user_followers(browser,
                     "ProfilePage[0].graphql.user.edge_followed_by.count")
             except WebDriverException:
                 try:
-                    allfollowers = format_number(browser.find_element_by_xpath(
-                                    "//li[2]/a/span[contains(@class, '_fd86t')]").text)
+                    allfollowers = format_number(browser.find_elements_by_xpath( 
+                        "//span[contains(@class,'g47SY')]")[1].text)
                 except NoSuchElementException:
                     logger.error("Error occured during getting the followers count of '{}'\n".format(user_name))
                     return [], []
@@ -816,8 +816,8 @@ def get_given_user_following(browser,
                     "ProfilePage[0].graphql.user.edge_follow.count")
             except WebDriverException:
                 try:
-                    allfollowing = format_number(browser.find_element_by_xpath(
-                                        "//li[3]/a/span[contains(@class, '_fd86t')]").text)
+                    allfollowing = format_number(browser.find_elements_by_xpath( 
+                        "//span[contains(@class,'g47SY')]")[2].text)
                 except NoSuchElementException:
                     logger.error("\nError occured during getting the following count of '{}'\n".format(user_name))
                     return [], []
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -700,18 +700,31 @@ def get_given_user_followers(browser,
 
     # check how many people are following this user.
     try:
-        allfollowers = format_number(browser.find_element_by_xpath(
-            '//li[2]/a/span').text)
-
+        allfollowers = format_number(browser.find_element_by_xpath("//a[contains"
+                                "(@href,'followers')]/span").text)
     except NoSuchElementException:
-        # todo check if private account?
-        logger.error('Could not determine if {} has followers'.format(user_name))
-        return []
+        try:
+            allfollowers = browser.execute_script(
+                "return window._sharedData.entry_data."
+                "ProfilePage[0].graphql.user.edge_followed_by.count")
+        except WebDriverException:
+            try:
+                browser.execute_script("location.reload()")
+                allfollowers = browser.execute_script(
+                    "return window._sharedData.entry_data."
+                    "ProfilePage[0].graphql.user.edge_followed_by.count")
+            except WebDriverException:
+                try:
+                    allfollowers = format_number(browser.find_element_by_xpath(
+                                    "//li[2]/a/span[contains(@class, '_fd86t')]").text)
+                except NoSuchElementException:
+                    logger.error("Error occured during getting the followers count of '{}'\n".format(user_name))
+                    return [], []
 
     # skip early for no followers
     if not allfollowers:
         logger.info('{} has no followers'.format(user_name))
-        return []
+        return [], []
 
     elif allfollowers < amount:
         logger.warning('{} has less followers than given amount of {}'.format(
@@ -727,11 +740,11 @@ def get_given_user_followers(browser,
 
     except NoSuchElementException:
         logger.error('Could not find followers\' link for {}'.format(user_name))
-        return []
+        return [], []
 
     except BaseException as e:
         logger.error("`followers_link` error {}".format(str(e)))
-        return []
+        return [], []
 
     person_list, simulated_list = get_users_through_dialog(browser, login, user_name, amount,
                                                  allfollowers, randomize, dont_include,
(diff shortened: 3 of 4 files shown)


sionking:
Hi. First of all you are doing a great job!
Second, I think we are missing something, atomic write. We must protect ourself from data loss! Atomic wrote can be done by:
- Writing to a temp file. 
- sync the Chach in disk to make sure data is on disk and not on Chace. 
- os.replace with original file name. 

Better we can use repo atomic_write since atomicy on windows is a bit harder to get, this repo unsure it. 

We can combine both solutions, for generic reliable files update.

sionking:
Please make this solution valid also for csv writer, this is more important to maintain valid follower list ( i did implemented kind of temporary file solution but we may need protection from ctrl+c and outage power)

uluQulu:
Thanks @sionking 🤩
I also thought of **tmp** file idea. Let me clear this,
1. Program is running. User presses `Alt+F4` >> `end` definition will not be executed, so user lost sessional data not been dumped to JSON and yet JSON is healthy cos it has not changed a bit.
2. Program is running User presses `CTRL+C` and before `follow_restrict` **dict** was properly dumped to JSON immediately presses `Alt+F4` >> user possibly lost data + now possibly has corrupt data

Considering the 2nd situation, **tmp** file solution will come to help. But for other situations handling the `SIGINT` will do the job without a leak. But I can't have a look at it any soon having busy days.

Well, CSV has a different situation. Cos it does not write entire data at the end of the program instead it write in a new line and if there were a corruption, it would be easier to solve it without all those. And **followedPool.csv** file is not included in `end` method, so in CTRL+C interruption, it is hardly is affected and only is only affected when it is the code being executed at that moment, etc., etc.

But yet if you want to apply `SIGINT` handling to any other place, it is very easy!
```python
from .util import interruption_handler

with interruption_handler():
    do_something()
    print(something)
    time.sleep(sometime)
```
I mean just importing and adding `with interruption_handler()` it will ignore any `SIGINT` signals during execution of the code with it :)

so particularly for **follwedPool.csv**,
```python
from .util import interruption_handler

with interruption_handler():
    with open('{0}{1}_followedPool.csv'.format(logfolder, login), 'a+') as followPool:
        followPool.write('{} ~ {},\n'.format(logtime, followed))
```
could do the job.

sionking:
Atomic writing is not only for ctrl+c situations, but also power fails for example.

sionking:
BTW, a better solution to close the app is to close the chrome browser that will lead to exception. no data harm.

uluQulu:
Yes it is the same thing as Alt+F4:
if power fails **or** if OS crashes **or** user presses Alt+F4, program does not reach `end` method
so the JSON file will be healthy, instead it will miss the sessional data and with this JSON structure it is impossible to make it work efficiently.
While implementing it, I thought about changing JSON to a DB structure. Cos JSON needs to write entire data while each dump, and if JSON data was to be written after each new action (_follow_) it would be pretty inefficient, cos from tests e.g. ~66k elements take 1.6 secons to dump...
But with a DB, it could be **safely** written in realtime at every new action. And more importantly, SQL engines like SQLite has a good open/close strategy and if it fails to write an element, it ignores that element and keeps DB healthy. And also, after a `ALT+F4` kind of situation, the only thing can be lost is the last element which was to be written.

Well, although DB method is amazingly efficient, it also is a long code (_than JSON load/dump_) and the job of followRestriction is not so crucial, but any time it would be great to use DB structure :)

uluQulu:
@sionking I like to close from console cos it stops all of the processes initiated by that session :)

uluQulu:
#### UPDATE 2
Raised in #2089, I noticed a **bug** in return values which now should return 2 values in certain conditions.

**Added** 1 more XPATH and a graphql layers of getting **followers**/**following** counts (_the same set used in `get_relationship_counts` definition_)

**Also**, removed `Record activity is not working on Windows.` warning message after verifying that it was already solved last year.

sionking:
@uluQulu  what about just using:
    except KeyboardInterrupt:
        session.end()
        quit()

This will work if press on ctrl+c only once

uluQulu:
`KeyboardInterrupt` exception. Well...
Before using signals, I used it,
```python
        followRes_dumped = False
        followRes_identical = False
        self.logger.warning("Please wait a second to dump followed users to followRestriction.json :>")
        
        while followRes_dumped==False and followRes_identical==False:
            try:
                if self.follow_restrict != self.follow_restrict_init:
                    before_dump = time.time()
                    dump_follow_restriction(self.follow_restrict, self.logfolder)
                    after_dump = time.time()
                    self.logger.info("Successfully dumped newly followed users in {} seconds"
                                "  ~ {} new users.\n".format(
                                    float("{0:.2f}".format(after_dump-before_dump)),
                                        len(self.follow_restrict)-len(self.follow_restrict_init)))
                    followRes_dumped = True
                else:
                    self.logger.info("Nobody was followed during the session.\n")
                    followRes_identical = True
            except KeyboardInterrupt:
                pass
```
And soon I realized that it can't catch every interruption: it doesn't catch anything in the except `block` and while you send SIGINT during execution of that part, it does immediately crash and leaves your file corrupt. (_just use that code and keep CTRL+C pressed for a few seconds_)
And much worse, with that try/except block it always starts from scratch to dump dict to JSON after **every** interruption.

And that's where I decided to use signals, **WHICH** can handle any SIGINT anytime and moreover it does not re-write data from scratch after every interruption, instead it ignores those interruptions during the write of the data :)

timgrossmann:
@uluQulu @sionking Especially since, from the linux kernel definition, every interrupt that is not handled leads to the program getting killed, the approach with the signals seems to be the correct one.

sionking:
But this not solving the main problem with this repo - ACID db.
Not getting reliable values when storing/retrieving info.

We need to figure it out, maybe use mongoDB?

uluQulu:
DB engines like SQLite works smoothly and is amazingly safe (consider `update_activity`)
But using a DB for such a simple data structure is much of complexity to avoid.

For reliability of data, with JSON approach used with that handler, the only unwanted situation is what I mentioned above:
If user first presses CTRL+C and JSON starts to emergency dump, if during that time user presses ALT+F4 or closes that window somehow (_shutdown, power outage_) the JSON will most likely be corrupt.

Else than that situation, JSON will be healthy. e.g. situation like,
User just presses ALT+F4 (_without CTRL+C_), for this, user will miss the sessional data of followed users to be dumped into JSON and yet will have the JSON file perfectly healthy.

Would consider the situation this deep and re-write with DB? - I think not...
_Also consider that there will be lots of other data read/write in the future and if sticked to DB approach on every difficulty, the code will be harder to maintain :)_
>@sionking, I have never used mongoDB, is it better than SQLite for this case?  And is it ultimately reliable?
_I still would choose SQLite which is used in almost every little DB need_.

Moreover if a stable **tmp** file solution added to it, that one scenario of data corruption will be solved.

uluQulu:
Also
Terminating session with ALT+F4 instead of keyboard interruption like CTRL+C brings unwanted behaviours.  
Just because of it does not execute `end` method, where, else than dumping **followRestriction.json**, there exists other several important blocks, too :
+ closing webdriver (**important**),
+ clearing cookies (**important**),
+ stopping possible gui display (**important**),
+ storing session end time in logger (_good for further data analysis_),
+ writes `self.followed` _value_ into **followed.txt** local file (_is useful_)

So, it says users should not use ALT+F4 at all (_for normal scenarios_) to terminate the program.
The rest is power outage or forcibly termination out of lag, well it is sad but it happens in ~1-2% of times and in any of those situations result is a corrupt file, rather, only sessional data will not be added the local JSON...

What are your final words?
@timgrossmann, @sionking 

Cheers

CharlesCCC:
Here are my 2 cents, I would think use DB is the best-way. And I do agree with @uluQulu it will add some complexity to the code. so the main challenge is how to implement DB module in a way that is easy to use and easy to maintain. 

I'm propose create a DB access module that handles every single activity update (like UpdateActivity) and then we can just call the coreleated method as necessary across the codebase.

uluQulu:
@CharlesCCC 
My final comments are my final thoughts, but for your suggestion I must say,
+ For making DB maintenance easier, alongside the DB, we must also include it's `making` module. It's a double-work and double-code. (e.g. `update_activity` came along as a ready-to-use inside already made **instapy.db** file by @converge, and it has no it's `making` code)
I mean in case of maintenance, the implementer will have to reverse-detect the way that DB was made if its making code is not included.

Well, for real need of a DB it can be considered but for this situation, I return to my previous comments and say long live JSON :D

BTW, I am implementing new features right now and which needs a few local data files and for the portability of the structure I used JSON again (`CSV` has version problems, `TXT` is so unstructured, etc.)
_I mean we can't refer to DB in every difficulty, it is also a disrespect to DBes ;)_

@CharlesCCC If you read my last comments, you will understand and support JSON, too 💪🏼

@timgrossmann @sionking @CharlesCCC I got new features waiting on the line, can we make a final decision on this?
features:
+ unified & sophisticated **followers**, **following** data grabber through graphql (READY)
+ smart _recent_ **unfollowers** grabber (READY)
+ with the help of those standalone grabbers, will come a new feature- **ghosts** grabber (FUTURE)

sionking:
I am not a database supporter as I have no idea in this field.
I just know that many situations can fail the existing data from been **reliable**.
We should handle that, this saying saving temp wil and using replace, or we can use the atomic_write repo (or any other) for file updating.

Well, let stick to JSON but let make a generic ACID method for that (temp file, os.repalce, flash to disk..)

timgrossmann:
I also think it's an extremely tough decision since DBs in lot of cases come with the "guarantee" of data integrity which is extremely important.
However, as @uluQulu we can't simply put everything in the same DB whenever we face some problems. 
I think the "hybrid" approach of keep using JSON but in combination with the @sionking mentioned atomic writing seems to be the be suitable approach for this.

@CharlesCCC I can absolutely understand your points here, too... But let's go with the atomic writing in combination with the JSON files.

@uluQulu Well those follow up features sound incredible! Looking forward to seeing them in action!

sionking:
well I had power off yesterday and got 3 files corrupted (I have backup) but I know only the files that are corrupted, the other I might miss:
image

What is the best way to avoid it ?

uluQulu:
@sionking 
Which files were affected or you know to be affected?

But **tmp** file solution is the way.

sionking:
We have the temp file solution for username_followedPool.csv but they are still affected all by power outage. since we don't do it right, I think we should use flush to disk or something ?

uluQulu:
@sionking, you said three files were affected, at least, which are them?

Anyways, I think I have said all my opinions and suggestions about this matter and anything more will be repetition.
And for **tmp** file usage, it is a separate work and must be done carefully cos of platform dependent troubles to be considered.

Cheers

sunnygill:
Hi All, 

I am new to github. I am wondering how do I pull or checkout this pull request on to my local git project. I have only cloned the master and want to pull these changes as I am facing the error.

uluQulu:
Hi @sunnygill 
Sure, have a read https://github.com/timgrossmann/InstaPy/pull/2042#issuecomment-390280939 I wrote recently to do what you want.
_Or just download my fork and use it._
But it is better to know that as of now, my fork is 5 commits behind the main repository so if you do the above, you will miss those 5 commits.
##### The best is to apply the changes in this PR manually to the same files.

And also, the changes in this PR will help to prevent further possible corruptions by keyboard interruptions but if you already have a corrupt JSON file, you have to fix it by hand.  
After applying these changes, you will not have that situation again.

Cheers

sunnygill:
Hi @uluQulu 
Thanks for a quick response.  How do I fix my followRestriction.json file manually. Do we need to fix both json files. One in the log folder and one in the insta profile folder within log

uluQulu:
It's a pleasure @sunnygill 
In many **Issues** I suggested the way to fix it, such as in https://github.com/timgrossmann/InstaPy/issues/1732#issuecomment-383390727

But basically, JSON file structure is as like below:
```javascript
{"user1":1, "user2":5, "user3":2}
```
Your corrupted file probably has some missing data from dump failure and it is sth like
```javascript
{"user1":1, "user2":5, "user3":
```
What you can do is to fix it by taking the last healthy value and closing it with curly brace
```javascript
{"user1":1, "user2":5}
```
_removed corrupt element `"user3":``_

And if you find it too hard to do it, just remove everything in that JSON file and keep only a pair of curly braces in there
```javascript
{}
```

And yes that JSON file is **followRestriction.json** located in your **logs** folder.

Cheers

uluQulu:
@sionking 
I really wonder how did you get three files (_at least_) corrupt at a power outage.
Cos since there is no any threads, so, you can end up with _only_ **one** corrupt file - one of the CSV files in the **logs** folder, nothing more.

>Maybe you mean having multiple corrupt files from multiple different sessions running?- Well that's not the case.

sionking:
@uluQulu "Maybe you mean having multiple corrupt files from multiple different sessions running?- Well that's not the case." TRUE many sessions in the same PC (kind of home server)

uluQulu:
Alright guys
Good that this is not merged.
I have came across to quit this approach fully because of good reasons.

Well, the reason is simple and I realized it the first time I looked at the method of writing of that JSON file which happens only at the call of `end` method.
Fine, with the help of `interruption_handler` and **tmp** file writing, the target file- JSON will always be healthy. There is no problem with it.

The problem is the loss of  sessional NEW data in case of `ALT+F4` kind of program terminations or unexpected forced terminations like from power outage or system (_program_) crash in which scenarios it will never reach method `end` and it will not write the NEW data.
The problem lies here is so huge, because user might have that session running for hours and have generated hundreds of new data to be dumped and once it  does not dump it, it simply is not good.

Those scenarios can be so less to be happened but they can happen, which is important.
E.g. some users would like `ALT+F4` kind of terminations, well you can recommend them to use `CTRL+C` in documentation but it is not a good approach, some users will not do it even reading docs, others will forget.
System crash & power outage also happens from time to time.

So once those reasons some up, the change in design is crucial.

The optimal solution is to use one-by-one data writing to the target file SO in any of the situations the data loss will not be more than 1.

one-by-one writing can be done with CSV style file formats but since this structure needs a file update the CSV format will struggle and it will be lots of code and harder to maintain.

That's where I recommend to use DATABASE for this crucial function.

And `interruption_handler` still can be and should be used in all other unsafe write methods.

I got another files who also dump to JSON as I said above, but I rethoughtagain and I see that those files are just helpful data files to do analysis and if one or some times they fails to dump data, it will not be a problem. The thing I mention is not all JSON files need to be re-written into DB, only the crucial ones like **followRestriction.json**.


I will try to rebase and remove those parts from code since it needs a re-design, but it can also be kept and replaced once new solution is available but I think it will be a waste, so it's better to not merge the `interruption_handler` into this part.

If I do not make to do it, I hope you these ideas will be any help for you to do it.

Cheers

uluQulu:
One more recommendation, while writing a DB approach, include it's making part in code, too and generate a DB file only by running the code.
It will make maintenance easier.
More importantly, when the DB is generated after running the program it has a great benefit:
e.g. DB like `update_activity` came along as a ready DB file to be pulled from the repo and it was not generated during the run of code. For that reason, each time users download repo and their that DB gets updated with new values, and then next time when they want to pull new files from the repo, they have to stash that DB in order to preserve their old data, otherwise if they force pull and ignore that DB file, they will download the zero-state DB file from the repo.
It causes loss of DB data in most of users who do not stash, and stashing is a pain.

But if it was auto generated, then git will see it as a NEWLY ADDED and UNSTAGED file so it will not make a conflict with the upstream repo.


This idea will also help while writing an auto-updater program for InstaPy.

Also, `update_activity` must be re-written like that and include it's making code.

BTW you may think so less people use `update_activity` for data analysis so it is not a problem but think about the future, e.g. Quota Supervisor or that kind of feature will need that file be in state.

Good luck!

uluQulu:
Hi guys
It has been a long time having this piece here to be reviewed. And I got another work waiting in line to be shared with the community.
So I wanted to know if you would like to merge this PR so that I can pull a new one.
>And yes, although, I can't promise but I will try to write a good DB for that error prone JSON file if gets some free time.

@timgrossmann, @CharlesCCC, @sionking 
Regards, Shahriyar.

timgrossmann:
@uluQulu Sorry for making you wait that much... Let's merge this!
Thank you for the hard work!

PR #2116, opened 2018-05-29
TypeError: makedirs() got an unexpected keyword argument 'exist_ok'

VerusK:
fixes bug when using proxy_extension
TypeError: makedirs() got an unexpected keyword argument 'exist_ok'


uluQulu:
Hi @VerusK 
Looks like you have an old **python** cos `exist_ok` parameter came with version 3.2.
And your code is perfect but if I were you I would use
```python
if not os.path.exists(dir_path):
    os.makedirs(dir_path)
```

VerusK:
I'm using python 2.7 :(

Issue #2129, opened 2018-05-31
Starting today Keep getting Too few images, skipping this tag


CharlesCCC:
## Expected Behavior
get targeted photos for like
## Current Behavior
```
INFO [2018-05-30 20:52:01] [username]  Tag [1/27]
INFO [2018-05-30 20:52:01] [username]  --> b'fitness'
ERROR [2018-05-30 20:52:25] [username]  Too few images, skipping this tag
INFO [2018-05-30 20:52:25] [username]  Tag [2/27]
INFO [2018-05-30 20:52:25] [username]  --> b'powerlifting'
ERROR [2018-05-30 20:52:49] [username]  Too few images, skipping this tag
INFO [2018-05-30 20:52:49] [username]  Tag [3/27]
INFO [2018-05-30 20:52:49] [username]  --> b'bodybuilding'
ERROR [2018-05-30 20:53:14] [username]  Too few images, skipping this tag
INFO [2018-05-30 20:53:14] [username]  Tag [4/27]
INFO [2018-05-30 20:53:14] [username]  --> b'weights'
ERROR [2018-05-30 20:53:37] [username]  Too few images, skipping this tag
INFO [2018-05-30 20:53:37] [username]  Tag [5/27]
INFO [2018-05-30 20:53:37] [username]  --> b'weightless'
ERROR [2018-05-30 20:54:01] [username]  Too few images, skipping this tag
INFO [2018-05-30 20:54:01] [username]  Tag [6/27]
INFO [2018-05-30 20:54:01] [username]  --> b'weight'
ERROR [2018-05-30 20:54:25] [username]  Too few images, skipping this tag
INFO [2018-05-30 20:54:25] [username]  Tag [7/27]
INFO [2018-05-30 20:54:25] [username]  --> b'loss'
ERROR [2018-05-30 20:54:49] [username]  Too few images, skipping this tag
INFO [2018-05-30 20:54:49] [username]  Tag [8/27]
INFO [2018-05-30 20:54:49] [username]  --> b'transformations'
```
starting this afternoon, every time it not getting any image to like. when I look at the run, the webpage showed all the top/main posts, but it is not liking anything at all. 

## Possible Solution (optional)
Not sure yet 
## InstaPy configuration
with default setting and like by tags along with following

CharlesCCC:
PR https://github.com/timgrossmann/InstaPy/pull/2130 for the fix.

tunidev:
same here when using  Authentication proxy

[...]

PR #2130, opened 2018-05-31
Fix issue that can't get the possible_posts amount

CharlesCCC:
Fix issue that can't get the possible_posts amount due to the css class changed to `g47SY`

### DIFF
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -216,7 +216,7 @@ def get_links_for_tag(browser,
     except WebDriverException:
         try:
             possible_posts = (browser.find_element_by_xpath(
-                                "//span[contains(@class, '_fd86t')]").text)
+                                "//span[contains(@class, 'g47SY')]").text)
             if possible_posts:
                 possible_posts = format_number(possible_posts)
             else:
--- a/instapy/util.py
+++ b/instapy/util.py
@@ -494,7 +494,7 @@ def get_relationship_counts(browser, username, logger):
             except WebDriverException:
                 try:
                     followers_count = format_number(browser.find_element_by_xpath(
-                                    "//li[2]/a/span[contains(@class, '_fd86t')]").text)
+                                    "//li[2]/a/span[contains(@class, 'g47SY')]").text)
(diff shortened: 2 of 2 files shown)


timgrossmann:
@CharlesCCC is this global? Or is it some incremental A/B-Testing change?

@uluQulu Could you double check that. Looks like it's the same change on my side.

uluQulu:
Hi guys
This problem stuck my mind. xD
Cos in the issue raised by @CharlesCCC in #2129 which lead to this PR, the error log shows:
skippingtag
But it is impossible, cos if it failed to get useful `possible_posts` count, it should still work without a problem and not show that error: `Too few images, skipping this tag`.
_Then, also wondered why it couldn't get that `possible_posts` count within **graphql** element..._

And yes it happened cos the files used in that setup is missing the commits came with #2042 
possibleposts

And since those changes are not existing, in case of `NoSuchElementException` raised while getting `possible_posts` count will be caught only by the outer except block in `InstaPy` class rather than the inner except block in the `get_links_for_tag` definition and will not be handled correctly.

That is not a problem if @CharlesCCC pulls those commits, it just did stuck my mind to see the `NoSuchElementException` caught by the outer except block in `InstaPy` class xD
error


And yes @timgrossmann, it is the same change in my side, too. And moreover, I checked, it has affected the **followers**/**following** counts elements, too, and there are 2 of them in **unfollow_util.py** and another 2 in **util.py** (_counts the changes with my latest PR_) which must all be renamed from `_fd86t` to `g47SY`.

Thanks to @CharlesCCC for super-quick catch 👍

CharlesCCC:
@uluQulu you are correct, my locally didn't have the later version (my bad). 
but the change should still valid. (of course, there are additional place need be changed as you pointed out).

Thank you for the code review and hopefully this is not a A/B test

bardascat:
why is this not merged yet ? Thanks !

sionking:
this is Y I am more of graphql fan

sionking:
i CHECKED BOTH CHANGES AND THEY ARE FINE

PR #2134, opened 2018-05-31
FIxed follow commenters

david-vendel:
Xpath changed fixed. Should work now.

### DIFF
--- a/instapy/util.py
+++ b/instapy/util.py
@@ -3,8 +3,10 @@ import datetime
 import os
 import re
 import sqlite3
+import time
 
 from selenium.common.exceptions import NoSuchElementException
+from selenium.common.exceptions import WebDriverException
 
 from .settings import Settings
 from .time_util import sleep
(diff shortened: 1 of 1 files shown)


hw711:
In get_relationship_count function, do you need two separate try blocks for getting follower and following counts?  Because if one fail, the other one will fail the same way so if you put them in the same block, you can cut the time the function takes when NoSuchElementException hits by 50%.

timgrossmann:
@justdvl Thank you very much for the PR.

Could you explain what all the changes are that you did?
When looking at the diff there is a lot of stuff going on, not just a "simple" XPath change.
Thanks

david-vendel:
You are right, I only made few changes but it displaied way too many things changed. Could it be because I didn't have the newest version of Instapy forked? I am not so good with Github, so it is bit confused now.

david-vendel:
The thing is, Github is showing that I changed the parts of code, which I didn't, -- it shows all the changes since April when I was last working on this. Before working on this code, I downloaded the new fresh version to my pc and verified that I am up to date, and changed only few things.
How should I fix this now?

timgrossmann:
@justdvl It actually should be as easy as doing a rebase on the current master branch from your branch.
You probably have to "update" your own fork with the latest changes.

david-vendel:
I will try to do it again. Can you delete this pull request?

PR #2154, opened 2018-06-02
Atomic write and atomic pickling by flush to disk and os.replace

sionking:
Please review.
Please compare to atomic_write:
https://github.com/untitaker/python-atomicwrites

If you think it won't be suitable to windows/linux.

Thanks.

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -510,7 +510,7 @@ class InstaPy:
                     for item in self.smart_hashtags:
                         print(u'[smart hashtag generated: {}]'.format(item))
             else:
-                print('Too few results for #{} tag'.format(tag))
+                print(u'Too few results for #{} tag'.format(tag))
 
         # delete duplicated tags
         self.smart_hashtags = list(set(self.smart_hashtags))
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -188,7 +188,7 @@ def get_links_for_tag(browser,
         media = [media]
 
     tag = (tag[1:] if tag[:1] == '#' else tag)
-    browser.get('https://www.instagram.com/explore/tags/{}'.format(tag))
+    browser.get(u'https://www.instagram.com/explore/tags/{}'.format(tag))
     # update server calls
     update_activity()
     sleep(2)
@@ -216,7 +216,7 @@ def get_links_for_tag(browser,
     except WebDriverException:
         try:
             possible_posts = (browser.find_element_by_xpath(
-                                "//span[contains(@class, '_fd86t')]").text)
+                                "//span[contains(@class, 'g47SY')]").text)
             if possible_posts:
                 possible_posts = format_number(possible_posts)
             else:
--- a/instapy/util.py
+++ b/instapy/util.py
@@ -537,17 +537,17 @@ def atomic_pickling(file_path, data, logger):
         file_path_Temp = file_path+".temp"
         try:
             with open(file_path_Temp, "wb") as f:
-                pickle.dump(data, f)
+                pickle.dump(data, f, pickle.HIGHEST_PROTOCOL)
                 # flush from memory to disk (might be cache)
                 f.flush()
                 # sync to disk from cache
                 os.fsync(f.fileno())
             # rename new temp file to filepath
-	        while os.path.isfile(file_path_Temp):
-               try:
-                   os.replace(file_path_Temp, file_path)
-               except OSError as e:
-                   logger.error("Can't rename file_path_Temp to filepath {}".format(str(e)))
-                   sleep(5)
+            while os.path.isfile(file_path_Temp):
+                try:
+                    os.replace(file_path_Temp, file_path)
+                except OSError as e:
+                    logger.error("Can't rename file_path_Temp to filepath {}".format(str(e)))
+                    sleep(5)
         except BaseException as e:
             logger.error("Error saving pickle file: {}".format(str(e)))
--- a/instapy/util.py
+++ b/instapy/util.py
@@ -11,6 +11,7 @@ from selenium.common.exceptions import WebDriverException
 from .settings import Settings
 from .time_util import sleep
 from .time_util import sleep_actual
+import pickle
 
 
 def validate_username(browser,
--- a/instapy/util.py
+++ b/instapy/util.py
@@ -337,7 +337,6 @@ def get_active_users(browser, username, posts, boundary, logger):
 
 def delete_line_from_file(filepath, lineToDelete, logger):
     try:
-        file_path_old = filepath+".old"
         file_path_Temp = filepath+".temp"
 
         f = open(filepath, "r")
@@ -350,28 +349,19 @@ def delete_line_from_file(filepath, lineToDelete, logger):
                 f.write(line)
             else:
                 logger.info("--> \"{}\" was removed from csv".format(line.split(',\n')[0]))
+        # flush from memory to disk (might be cache)
+        f.flush
+        # sync to disk from cache
+        os.fsync(f.fileno())
         f.close()
 
-        # File leftovers that should not exist, but if so remove it
-        while os.path.isfile(file_path_old):
-            try:
-                os.remove(file_path_old)
-            except OSError as e:
-                logger.error("Can't remove file_path_old {}".format(str(e)))
-                sleep(5)
-
-        # rename original file to _old
-        os.rename(filepath, file_path_old)
         # rename new temp file to filepath
         while os.path.isfile(file_path_Temp):
             try:
-                os.rename(file_path_Temp, filepath)
+                os.replace(file_path_Temp, filepath)
             except OSError as e:
                 logger.error("Can't rename file_path_Temp to filepath {}".format(str(e)))
                 sleep(5)
-
-        # remove old and temp file
-        os.remove(file_path_old)
 
     except BaseException as e:
         logger.error("delete_line_from_file error {}".format(str(e)))
@@ -494,7 +484,7 @@ def get_relationship_counts(browser, username, logger):
             except WebDriverException:
                 try:
                     followers_count = format_number(browser.find_element_by_xpath(
-                                    "//li[2]/a/span[contains(@class, '_fd86t')]").text)
+                                    "//li[2]/a/span[contains(@class, 'g47SY')]").text)
                 except NoSuchElementException:
                     logger.error("Error occured during getting the followers count of '{}'\n".format(username))
                     followers_count = None
@@ -516,7 +506,7 @@ def get_relationship_counts(browser, username, logger):
             except WebDriverException:
                 try:
                     following_count = format_number(browser.find_element_by_xpath(
-                                        "//li[3]/a/span[contains(@class, '_fd86t')]").text)
+                                        "//li[3]/a/span[contains(@class, 'g47SY')]").text)
                 except NoSuchElementException:
                     logger.error("\nError occured during getting the following count of '{}'\n".format(username))
                     following_count = None
--- a/instapy/util.py
+++ b/instapy/util.py
@@ -533,7 +533,7 @@ def web_adress_navigator(browser, link):
         sleep(2)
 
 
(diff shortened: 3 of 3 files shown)


uluQulu:
Cool @sionking 
I see you have used pickle and I know that you're a pickle master xD

The same procedure can be done easily with
+ write file in tmp filename
+ if process is successful
+ rename to ogirinal filename

It will do job well.


Fine it is amazing it will always avoid data corruption but it has drawbacks:
If program exists before renaming to original filename, the sessional data will be **LOST**.

For small data writes, it is not a problem (_e.g. CSV or TXT one line at a time_) cos small data lost is not much a problem.

But whilst huge data writes (e.g. JSON with ~1k elements) if it fails to do it and rename, the original file will still be healthy but the sessional data will be lost (_fully **or** some part can be found in tmp file maybe_)

And if that sessional data dump is really important (_consider that those ALT+F4 situations are so rare in normal scenarios_), then there is really no way to make it work with bluk-data writer filetypes (_like JSON_).

Instead if we use on-the-go data writing (_e.g. CSV, TXT, **DB** but not sth like JSON_) that problem can be solved.

And esp if we use DB it will do the job itself without the help of data corruption/lost preventors.

DB cannot be used in everyplace. So this situation is so sad, cos I liked JSONs :(

sionking:
@uluQulu 
1. I did not used pickle, it is just an helper function if you want to atomic write pickle.
2.  you are wrong, I get a lot of file corruption with simple implementation of temp file (delete_line_from_file method)
3.  rename in windows is not atomic as replace.
4. we have to make sure the data is on disk and not in buffers.
5. monjoDB is a JSON database. simple to use. python compatible.

SCK

uluQulu:
@sionking 
I wonder if you really get data corruption right from that innocent function, also if `os.rename` does not work portably well, there are plenty of alternatives to use, e.g. with `call` from `subprocess` module :)

sionking:
@uluQulu "call from subprocess module" ? I didn't get it. can you explain?
os.rename works but it is not atomic in windows.

sionking:
Recommand to you all:
image

I got tones of NULL bytes.

uluQulu:
I got 946 lines in **followedPool.csv** and 1114 lines in **record_all_followed.csv** (_those are biggest CSV files_) and got no any NULL character.

Actually, I will touch to some points.
You have said `os.rename` is not atomic in Windows. Yes that's write, it's mainly cos `os.rename` cannot overwrite an existing file and throws an OSError.
That's why the implementor of the `delete_line_from_file` definition used a smart algorithm to prevent this clash. And I have checked, that function is amazingly super and provides zero-corruption, but a bit long.
_Although, there is a condition you can end up with file.old and file.tmp if it fails at last step (_out of external interruption_) but you would never get a corrupt file_.

Okay, `os.rename` is not atomic to use directly cos it cannot not overwrite existing files in Windows platforms and only for that purpose Python guys implemented `os.replace` function which can overwrite existing file. And there is not any other differences (_mainly_) between those functions, except that.
**Details**: `os.rename` uses `MoveFileW` (_cannot overwrite_) and `os.replace` uses `MoveFileExW` (_can overwrite_) from Windows C++ API.

Also, you should know that `os.replace` was introduced in 2012 with python 3.3 release, that's why it is not available in previous versions and should be avoided for portability.

`call` function from `subprocess` module is kind of a wrapper around well-known `Popen` but is much safer than `Popen` cos it waits for command to complete then leaves function so it does not allow possible conflicts.
using `call`, we can send native OS commands out of the box, e.g. for Windows, we can do that job like this:
```python
from subprocess import call as move_it

move_it("move /Y exfile newfile", shell=True)
```
`move /Y exfile newfile` is a native Windows command to be run in command prompt.
`/Y` is set to overwrite the existing file if any.

Do we really need flush?
Consider that the temp and original files are both located in the same path.

Also, the corruption you got in your CSV files are possibly cos of unsafe **writer** methods which has no temp file methods.

uluQulu:
Also, having the **tmp** and  **original** file located in the same drive is highly crucial for performance.

Cos once they are in the same drive, `MoveFileExW` operation will operate using its `pointer` in no time, but if not in the same drive, it will first run `CopyFile` and then rename which will cost a performance (_even imagine file is huge_) while copying and in any case will cost slow down program cos _write_ operations are being requested too many times.

And, having multiple platforms using the program with various filesystems, it is better to located **tmp** files in the same path with **original** files.

Regarding the simple case in the write procedures of this program, what I offer again is to use the simple implementation of
+ write to **tmp**
+ if successful, overwrite original

sionking:
Hi I will make it short.
I have around 10 different windows pc/laptop that got NULL BYTES in CSV after power outage, some of them are the followers file which rewrite every delete line from file..
I beat you never experienced power outage, try using csv.writer and turn off the PC during the method ?
BTW, I am the one who wrote the delete line from file implemented with temp file, since lost of information.
Why are you against flush? this is done in every atomic write method I checked, it cost nothing.
I also use Dropbox to sync all accounts to one manager server, so it maybe the cause for the NULL but I doubt.

BTW, you are amazing, reading your comments and see your knowledge is inspiring.

uluQulu:
Thank you @sionking 
If not having too much programming knowledge, I have extensive experience with various kinds of software cos I have worked in almost any platform and used tons of software to day, that's why I can look to problems in compact view 🤩

If you say so and flushing and then syncing from requirement costs nothing, well I did some tests.

Max test took place as:
Having a string with 1771 characters (_integers and various chars_)
Did test writing to a CSV file in append `a+` mode without a time break (_non-stop_)
(_at final there were 2384 lines in the file_)

And I can say, from **Task Manager** monitoring,
It ate up to max of ~10% CPU and 38% DISK usage which is scary
_Computer usage was normal, not much loaded not so free_.

And the time spent to flush is really too few
minimum `0.0`
mean `0.0009703636169433594`
max ` 0.018550634384155273`

But for `fsync`, it is devastatingly high
min `0.018088102340698242`
mean `0.08024721814875017`
median `0.07421016693115234`
max `0.76101927757263184`

#### for InstaPy
The data to be written will not be more than ~50 characters
And this test results should be thought to not taken seriously, BUT, even in small tests, `fsync` takes huge time similar to above.
**Although** it does not eat up that much computer resources during small data writes.

And considering that data will not be written continuesly  the results can be brought down a bit, too.

#### OPINIONS

I think, for ultimate data integrity, we need `flush` and for `flush` we need `fsync`.

I also think, we may not need it, because in the worst case the data will be lost is so small considering the one-line-per-time writes.

It is up to you to decide, personally I like data integrity and I can recommend to even decrease default sleep time generator's stochasticity percent from `95` to e.g. `90` which will solve the little time spent to `fsync`.

#### To @sionking 
Your code is cool, I think, finalizing it and applying it to all of the writer definitions will be so good.
First you should solve the `os.replace` conflict. Using `subprocess` module it will be much easier to do it.  
Having known that the move (_with overwrite_) command is `mv` in almost any UNIX-like OS and `move /Y` for Windows, it can be like.
```powershell
move_command = "move /Y" if os.name=="nt" else "mv"
``` 
and running it with `call` function.
But applying the same procedure you did in `delete_line_from_file` is also a good option in a long a way.

**Summarizing**, adding the same set of **tmp** method plus `flush` & `fsync` (_with final decision_) into all of the writer definitions (_instead of making a separate definition for this purpose_) will solve all of possible problems (_well still, there are rare situations that **can't be solved** but they won't cause corruption_).

@timgrossmann @CharlesCCC 
Can you guys join this highly crucial conversation, too? ☕

@sionking 
Actually, I have tried it before :D
Writing to a CSV (_non-stop_) and closing it with ALT+F4 (_CTRL+C, too_) using a too large string to write and really in some cases it did stop at writing phase and interestingly, in all of those cases, the output CSV was pretty healthy :_)
Maybe power outage and ALT+F4 are behaving differently regarding the exit of the program...
It's kind of a mystery for me how you ended up with corrupt files :/

Also does you Dropbox service have write access to the data files, if so, then it can be piece of the problem too..


Cheers

timgrossmann:
@sionking @uluQulu Thank you very much for all the considerations you guys put into this problem.
It's in fact very important to resolve this correctly since there are a ton of people out there who don't know how to fix a broken JSON file. 

@uluQulu I think you "Summarising" part pretty much sums up the final decision I'd go for as well.
`adding the same set of tmp method plus flush & fsync` sounds like a reasonable approach, even though it seems that `fsync` is pretty compute intensive I think it should not be a problem here.

@sionking I'm actually really impressed by what I read about your setup every now and then 👍

sionking:
@uluQulu 
"Maybe power outage and ALT+F4 are behaving differently regarding the exit of the program..."
The big diff is flush to disk...., ALT+F4 will still flush to disk, while power outage can stop in a middle of flushing.
Scenario (no flush):
- Write to temp file
- os.replace (original file is dead but new file in buffers)
- during un-buffering of new file there is a power outage
- NULL BYTES (probably some space savers for dynamic memory allocation)

docs:
https://blog.gocept.com/2013/07/15/reliable-file-updates-with-python/
https://stackoverflow.com/questions/7433057/is-rename-without-fsync-safe
https://github.com/untitaker/python-atomicwrites

@timgrossmann 
Since I am missing proxy servers in my country I built my own servers and proxies.
Way more complicated :( but working pretty cool.

Note
Actually there is another solution called cache proxy (but lets not get into it)/

sionking:
@uluQulu 
Hi, why not using os.replace ?
It is python method that works.

uluQulu:
@sionking 
Have you read my third comment?
untitled

There are several methods to use.
**Such as**,
```python
import sys
import os
import suprocess

if sys.version_info >= (3, 3):
    os.replace()

elif sys.platform == "win32":
    subprocess.call("move /Y {} {}".format(exfile newfile), shell=True)

else:
    # POSIX rename() is always atomic
    os.rename()
```
_For less mess but long code._


**OR** use `call` for all.

**OR** you can remove original file before renaming, it possibly will not overwrite 🌘  but if process fails the original file will not be around.
The similar procedure in `delete_line_from_file` which renames original file to `.old` to prevent overwrites (_instead of removing_) also has a leak, cos in a failure, it can end up with `.old` and `.tmp`.

Well there are too many methods to use...

sionking:
I will read all later. But I think the methods that have issues are only the one that updates the file in the middle. Writing to end of file can corrupt file but not to loss data.

timgrossmann:
Unnecessary due to #2563

timgrossmann:
@sionking @uluQulu Found it, if that is the PR you were talking about.

sionking:
@timgrossmann this is the PR

sionking:
@timgrossmann @uluQulu 
Are we going to stop using python 2.7? if so we can merge this.

timgrossmann:
@sionking Good question. At the end of the year, Python2 will officially get deprecated: https://pythonclock.org

So yes, I just don't know when.
We probably should discuss this in the core_team group on discord.

I've starte a discussion there.
Let's see, we even have a issue for that in the roadmap 😉 
https://github.com/timgrossmann/InstaPy/issues/3580

uluQulu:
<h1>😥</h1>

image



<h6>It was a great experience..</h6>

timgrossmann:
We will remove text files that hold information completely in the future and therefore will use the atomic behaviour of DB actions.
https://github.com/timgrossmann/InstaPy/issues/3924

sionking:
@timgrossmann 
I still think this should enter for now, as the db usage going to be in couple of months.
Anyway I use it for last few months and it works much better.

timgrossmann:
@sionking Hmm, true It will take much more time until this Issue gets implemented.
Have you encountered any instances where the file was corrupt due to problems with the `os.replace` method?
Also, we haven't deprecated python 2, yet.
We still need to figure out how to provide good documentation on python 3 for raspi, @JoeFertig is currently working on it.
Once everything is resolved, we can deprecate python 2 and switch all documentation to python3.

Then we can talk about implementing this PR again.
Thanks for the comment though, I'll put a note to this PR to keep it in mind.

Issue #2157, opened 2018-06-02
New account - not filling the username.


sionking:
It only fills the password and throw wrong login data.
anyone managed to solve it ?

solehpolysas:
@sionking  already done here https://github.com/timgrossmann/InstaPy/issues/2133#issuecomment-394098747

using open PR #2146  but need to ADD small delay `sleep(1)` at new line 147.
Mine working fine. 👍 

Feel free to check & update.

ghost:
works for me on ubuntu 
added sleep(1) at new line 147 in instapy/login_util.py

axlright:
Had this same issue and after banging my head against the keyboard for an hour or two, PR #2146 and https://github.com/timgrossmann/InstaPy/issues/2133#issuecomment-394098747 worked perfect. 👍 🥇

sambrucejoseph:
Still doesn't work after changing the class name and adding the sleep(1)
It still doesn't key in the username

Issue #2153, opened 2018-06-02
Wrong login credentials


Baniyal:
## Expected Behavior
normal functioning of the script
## Current Behavior
in spite of the chrome window opening and the automatically logging into my Instagram.i can view it on the chrome window that is generated. Although soon after that the window abruptly closes and the error displayed is "Wrong login credentials".Yes i have verified the credentials ! 

Thankyou in advance

imbarismustafa:
Can you share the error with us. Thanks.

sionking:
R u using any server like AWS / DO ?

solehpolysas:
Can you try this #2133 
Maybe its related to your issue.

Hope that help.

PR #2146, opened 2018-06-01
Language class changed to hztqj

Armax:

Issue #2133, opened 2018-05-31
[FIX]IG Change Login Class Element - Cannot Creating Cookies or New Login Issue When Deleting Cookies


solehpolysas:
## Expected Behavior
Should be working & creating new cookies for new account in the System.
## Current Behavior
Error when creating cookies. Stuck at login page, not login anymore.
`Traceback (most recent call last):
  File "quickstart.py", line 20, in <module>
    session.login()
  File "C:\Users\admin\Desktop\Tools\Backup InstaPy\InstaPy-master\instapy\instapy.py", line 322, in login
    self.bypass_suspicious_attempt):
  File "C:\Users\admin\Desktop\Tools\Backup InstaPy\InstaPy-master\instapy\login_util.py", line 136, in login_user
    "//select[@class='_fsoey']/option[text()='English']").click()
  File "C:\Users\admin\AppData\Local\Programs\Python\Python36\lib\site-packages\selenium\webdriver\remote\webdriver.py", line 385, in find_element_by_xpath
    return self.find_element(by=By.XPATH, value=xpath)
  File "C:\Users\admin\AppData\Local\Programs\Python\Python36\lib\site-packages\selenium\webdriver\remote\webdriver.py", line 955, in find_element
    'value': value})['value']
  File "C:\Users\admin\AppData\Local\Programs\Python\Python36\lib\site-packages\selenium\webdriver\remote\webdriver.py", line 312, in execute
    self.error_handler.check_response(response)
  File "C:\Users\admin\AppData\Local\Programs\Python\Python36\lib\site-packages\selenium\webdriver\remote\errorhandler.py", line 242, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.NoSuchElementException: Message: no such element: Unable to locate element: {"method":"xpath","selector":"//select[@class='_fsoey']/option[text()='English']"}
  (Session info: chrome=66.0.3359.181)
  (Driver info: chromedriver=2.39.562718 (9a2698cba08cf5a471a29d30c8b3e12becabb0e9),platform=Windows NT 10.0.16299 x86_64)`

## InstaPy configuration
Basic quickstart, just need to creating cookies, but FAIL.

[20180531-132628.zip](https://github.com/timgrossmann/InstaPy/files/2056794/20180531-132628.zip)

Anyone facing the same issue?
If you already have cookies, i dont think you will get this error.

But if you change or delete cookies or use new account, this error will come.
Someone please verify this.

Thanks in advance.

[...]

PR #2175, opened 2018-06-04
Instagram update fix

cmaixen:
Solves username in wrong field during login and handles update of instagram.

### DIFF
--- a/instapy/login_util.py
+++ b/instapy/login_util.py
@@ -133,7 +133,7 @@ def login_user(browser,
     # Might cause problems if the OS language is english
     if switch_language:
         browser.find_element_by_xpath(
-          "//select[@class='_fsoey']/option[text()='English']").click()
+          "//select[@class='hztqj']/option[text()='English']").click()
 
     # Check if the first div is 'Create an Account' or 'Log In'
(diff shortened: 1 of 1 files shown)


timgrossmann:
@cmaixen Thank you very much for this fix and your first contribution.
However I do not really understand the point with "Solves username in wrong field during login"... You're not changing anything in terms of the input field for the login credentials.
You are merely changing the language selection class

Could you quickly explain what you mean, please?

cmaixen:
@timgrossmann This is a fix for #2133,#2153,#2146,#2157 . When you add a new user or you delete your cookies and you need to login again, the script is too fast. It will enter the username in the field for making a new account and not in the username field to login. The password is entered in the correct field. As a result, the script will tell you in the logs: "Wrong login credentials". This is because the username field of the login is left empty. By adding a sleep, you solve this issue. Not completely my idea, @solehpolysas suggested it. :)  I've just merged two solutions in one.

timgrossmann:
@cmaixen Thank you very much, just double checked the changed class for the language selector and everything seems to be correct.

Looking forward to future contributions 😉 
Thank you

PR #2177, opened 2018-06-05
Update of xpath + num_od_posts

david-vendel:
Updated follow commenters procedure following recent Instagram updates.
Minor update in number of posts to scroll through.


david-vendel:
OoOps it seems like sambrucejoseph already pulled this fix... My work is redundant, still one_pic_element is not needed to declare twice, and num_of_posts = (min (num_of_posts,max_pic)) is reasonable update.

timgrossmann:
@justdvl Unfortunately, I don't really understand your second comment.
Is this still valid in terms of changes?

PR #2205, opened 2018-06-09
Binary cookie for login

Armax:
Two new options for login_user function, use_file=True, passed_cookie=None if use_file is False it will return a cookie instead of creating a file, passed_cookie can be a Binary type cookie and if is not None will be used instead of the file

### DIFF
--- a/instapy/login_util.py
+++ b/instapy/login_util.py
@@ -90,7 +90,10 @@ def login_user(browser,
                password,
                logfolder,
                switch_language=True,
-               bypass_suspicious_attempt=False):
+               bypass_suspicious_attempt=False,
+               use_file=True,
+               passed_cookie=None):
+
     """Logins the user with the given username and password"""
     assert username, 'Username not provided'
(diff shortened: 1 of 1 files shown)


timgrossmann:
@Armax Really cool, thank you very much for the contribution!

Could you please do two things here?

1. Could you please explain what this feature is useful for
2. Add documentation to the README that helps people understand how to use this feature and why it's useful

Armax:
This feature is useful if you don't want to store the passwords in cleartext inside the code or a file, after the first login you can save the cookie in binary format

CharlesCCC:
@Armax I believe we are already doing this. if you logged in the first time, the cookie is stored (not necessary in binary format). and you don't need the password anymore (you can put 123456789 as password if you want) until your cookie is expired.

Armax:
@CharlesCCC yes but is stored in a file, if you run on a remote server is not the best thing to use files as storage, if you have binary format you can also store it in a DB

PR #2231, opened 2018-06-13
Avoid stop of execution if there are too few images

redsector72:
This raise will stop execution when there are few images than expected.
I don't understand why is there

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -796,7 +796,6 @@ class InstaPy:
                                                skip_top_posts)
             except NoSuchElementException:
                 self.logger.warning('Too few images, skipping this location')
-                raise
                 continue
 
             for i, link in enumerate(links):


timgrossmann:
@redsector72 You are absolutely right. I think since this feature is not used too much, it's some left over code from quite a while ago, so thank you very much for the catch!

uluQulu:
I am not sure but I can be the one who did that change 😂
Cos there were a similar situation where the thrown exception message was really needed to find and fix the error occurred WHERE error was not really out of _less images_, but from some _change in page layout_.

>Back then I was a hardcore `raise`r but not I know that it is good only for test purposes and some rare situations.

So, if not to `raise` which will terminate the program, it must be catched wisely.
Also the current logger message is so misinforming which should be changed too.

I suggest to write it like this,
```python
except NoSuchElementException as exc:
    self.logger.warning("Error occured while getting images from location: {}  "
            "~maybe too few images exist\n\n{}\n\n".format(location, exc.encode("utf-8")))
    continue
```


Regards

PR #2235, opened 2018-06-13
more clear logs; add aborting;

sionking:
change logs to be more clear;

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -666,7 +666,7 @@ class InstaPy:
                                            self.min_following,
                                            self.logger)
             if validation != True or acc_to_follow==self.username:
-                self.logger.info("--> not a valid user: {}".format(details))
+                self.logger.info("--> Not a valid user: {}".format(details))
                 not_valid_users += 1
                 continue
 
@@ -830,7 +830,7 @@ class InstaPy:
                                                        self.min_following,
                                                        self.logger)
                         if validation != True:
-                            self.logger.info("--> not a valid user: {}".format(details))
+                            self.logger.info("--> Not a valid user: {}".format(details))
                             not_valid_users += 1
                             continue
                         else:
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1344,7 +1344,7 @@ class InstaPy:
                                            self.min_followers,
                                            self.min_following,
                                            self.logger)
-            if valid_user is not True:
+            if validation is not True:
                 self.logger.info("--> not a valid user: {}".format(details))
                 not_valid_users += 1
                 continue
@@ -1530,7 +1530,7 @@ class InstaPy:
                                            self.min_followers,
                                            self.min_following,
                                            self.logger)
-            if valid_user is not True:
+            if validation is not True:
                 self.logger.info("--> not a valid user: {}".format(details))
                 not_valid_users += 1
                 continue
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1344,7 +1344,7 @@ class InstaPy:
                                            self.min_followers,
                                            self.min_following,
                                            self.logger)
-            if validation is not True:
+            if not validation:
                 self.logger.info("--> not a valid user: {}".format(details))
                 not_valid_users += 1
                 continue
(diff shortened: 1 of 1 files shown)


instapy/instapy.py
```diff
@@ -1343,8 +1344,8 @@ def like_by_users(self, usernames, amount=10, randomize=False, media=None):
                                            self.min_followers,
                                            self.min_following,
                                            self.logger)
-            if validation != True:
-                self.logger.info(details)
+            if validation is not True:
```
uluQulu: @sionking 😂
In python, `is` and `is not` is better suited for checking againist `None`,
For _Boolean_ checking, it is best to use `==` and `!=` OR just `if somebool:` and `if not somebool:`
_And yes there is another value which comes into the scene: `0`.._

That's why it is best to check Booleans directly with _equal_ and _not equal_ expressions for ultimate value matching.

>I also really liked easier `is` &  `is not` early days but it is **python**, it has its own rules 🚦

instapy/instapy.py
```diff
@@ -830,7 +830,7 @@ def like_by_locations(self,
                                                        self.min_following,
                                                        self.logger)
                         if validation != True:
-                            self.logger.info(details)
+                            self.logger.info("--> not a valid user: {}".format(details))
```
uluQulu: This is great @sionking!
Bcos the `details` string has different messages according to the validity level, having a general `Not a valid user` message on top of specific message is better and reader easily can see and detect if the user _is_ or _is not_ valid from console log..

Please uppercase write `Not` for better readability.

instapy/instapy.py
timgrossmann: What is the reason we are not simply using...
`if validation` and `if not validation` here?

uluQulu: Looks simple, right, @timgrossmann.

E.g.
`zx=0`, `zx=[]`, `zx={}`, `zx=()`, `zx=None`, `zx=False`, `zx=[          ]`, etc.
```python
if zx:
    print("chocolate")
```
all of those guys will not give chocolate. Not cos of they are greedy, just cos they all are empty (literally).

Well, I have used such statements in some parts of the code e.g. relationship bounds verification lines uses too many such kind of checking, and only for the clearance of the code I used that confusing mixed art cos there were too many checks (null, Boolean, etc.).

But in elsewhere, everywhere, it is best to use what is recommended by python. And for definitely precise result, it is best to use `equal` and `not equal` in that particular expression.

The idea is to have a max safe source code. It's important, cos now validation may only return `True` or `False` and it may look safe to use other styles, but in future it can return other values which can confuse implementer.

What I do right now is, to follow ultimate type safety whenever possible and accommodate with the recommended rules by the language itself- python here.

Cheers

sionking: Anyway we only return True/False from this method.
But I also consider != safer to use as it checked equality and not identity.

sionking: Anyway the "right" method by python is the one I suggested:
https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_true.html
when comparing to True.
But I am a c guy, nothing confusing in c as in those elite languages.

timgrossmann: @uluQulu Thank you for the explanation. I'm aware of those equality comparisons with empty structures, it's comparable in JS
But like @sionking already mentioned, I think we're only expecting a boolean value here that's why I thought we should use the python way... Only doing `if <boolean>:` if we want to check for true. 
But doing`== True` here might be good to do if we are not 100% sure that this will be a boolean value

uluQulu: Yeah right guys
I guess I forgot I was talking to the masters.  
>Unlike you, I can't be even called a programmer cos all I know is so basic C++ and some bits of InstaPy-python. 😂  

You are also right about that, python team may offer PEP stuff but when I called python recommended way I meant the best practices I've learnt, heard, read so far.

sionking: Don't look at me, I am experts on networking and memory management with C.
Python is in my best efforts...
Sometimes I like it since it simple sometimes I found it confusing :()

timgrossmann: Haha, don't look at me either 😂
I'm mostly doing python in my free time.
I just like to sometimes read the style guides 😉

sionking: @uluQulu 
Don't be modest you are really not a basic programmer.
Unleash the expert behind the mask!

uluQulu: Thanks @sionking, it's really an honor to hear it from a real programmer!
I wish I could be a one one day 😊

Issue #2206, opened 2018-06-09
Console Logging for Interact with User Followers


gitpatrickhub:
## Expected Behavior
In prior versions of InstaPy,  interaction with a Users Followers provided a console log of the aggregated total of all the interactions. 

<img width="718" alt="screen shot 2018-06-09 at 9 01 35 am" src="https://user-images.githubusercontent.com/29474212/41192562-47f79510-6bc5-11e8-92c5-ccbcb9987f78.png">

## Current Behavior
In the current version of InstaPy, the console log only shows the total number of interactions.

<img width="738" alt="screen shot 2018-06-09 at 9 02 44 am" src="https://user-images.githubusercontent.com/29474212/41192566-56d69fea-6bc5-11e8-85b8-4522c7049d35.png">

## Desired Behavior
Is it possible to revert back to the previous console logging format? It is helpful to see how effective each run is via the console summary.

uluQulu:
Good idea! @gitpatrickhub 
Can you implement it?
If not, if I implement, can you PR it? 😏

gitpatrickhub:
@uluQulu I can try to implement it, but need a bit of guidance on where this part of the code is managed. Just let me know which you prefer. Thanks!

uluQulu:
That's amazing! @gitpatrickhub 
There is always only one optimal solution to problems try to find it, challenge yourself! ^.^

Hint:
+ Always see other people's code which is doing the same (_or similar_) thing you want to write. By this, I mean look at source files and see if such kind of code already exist to use quickly.
+ To write code in correct place, you must find where that correct place is, to do it, use "GO TO ROOT" idea, it is done by following every function call to until finds your correct place.

If you have any detailed question, feel free to ask.
Good luck!

gitpatrickhub:
@uluQulu after hitting send, I kind of figured your answer would be as such. I'll see what I can do.

uluQulu:
Yes @gitpatrickhub 😅
It's because I am quite a self learner.

But I will share a few more tips to help you get started
+ There are existing lines to use in `InstaPy` class
e.g.
```python
self.logger.info('Liked: {}'.format(liked_img))
self.logger.info('Already Liked: {}'.format(already_liked))
self.logger.info('Commented: {}'.format(commented))
self.logger.info('Followed: {}'.format(followed))
self.logger.info('Inappropriate: {}'.format(inap_img))
self.logger.info('Not valid users: {}\n'.format(not_valid_users))
```
[...]

PR #2243, opened 2018-06-14
Update instapy.py for CLI logging

gitpatrickhub:
@uluQulu Here is the PR for the adding back the summary logging info to the CLI for interact_user_followers & interact_user_following that I suggested. #2206

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1722,8 +1722,8 @@ class InstaPy:
 
         interacted_all = 0
         simulated_unfollow = 0
+        not_valid_users_initial = self.not_valid_users
         not_valid_users = 0
-        #initialize counts for summary logging
         liked_img = self.liked_img
         already_liked = self.already_liked
         commented = self.commented
@@ -1805,14 +1805,14 @@ class InstaPy:
         commented_sum = self.commented - commented
         followed_sum = self.followed - followed
         inap_img_sum = self.inap_img - inap_img
-
+        not_valid_users_sum = self.not_valid_users - not_valid_users_initial
 
         self.logger.info('Liked: {}'.format(liked_img_sum))
         self.logger.info('Already Liked: {}'.format(already_liked_sum))
         self.logger.info('Commented: {}'.format(commented_sum))
         self.logger.info('Followed: {}'.format(followed_sum))
         self.logger.info('Inappropriate: {}'.format(inap_img_sum))
-        self.logger.info('Not valid users: {}\n'.format(not_valid_users))
+        self.logger.info('Not valid users: {}\n'.format(not_valid_users_sum))
 
         self.not_valid_users += not_valid_users
         
@@ -1833,8 +1833,8 @@ class InstaPy:
 
         interacted_all = 0
         simulated_unfollow = 0
+        not_valid_users_initial = self.not_valid_users
         not_valid_users = 0
-        #initialize counts for summary logging
         liked_img = self.liked_img
         already_liked = self.already_liked
         commented = self.commented
@@ -1917,14 +1917,14 @@ class InstaPy:
         commented_sum = self.commented - commented
         followed_sum = self.followed - followed
         inap_img_sum = self.inap_img - inap_img
-        
+        not_valid_users_sum = self.not_valid_users - not_valid_users_initial
         
         self.logger.info('Liked: {}'.format(liked_img_sum))
         self.logger.info('Already Liked: {}'.format(already_liked_sum))
         self.logger.info('Commented: {}'.format(commented_sum))
         self.logger.info('Followed: {}'.format(followed_sum))
         self.logger.info('Inappropriate: {}'.format(inap_img_sum))
-        self.logger.info('Not valid users: {}\n'.format(not_valid_users))
+        self.logger.info('Not valid users: {}\n'.format(not_valid_users_sum))
 
         self.not_valid_users += not_valid_users
 
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -791,8 +791,9 @@ class InstaPy:
                                                self.logger,
                                                media,
                                                skip_top_posts)
-            except NoSuchElementException:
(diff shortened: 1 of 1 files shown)


uluQulu:
That's great, @gitpatrickhub!

Hope gets a time to test it.

Keep up good work!

instapy/instapy.py
```diff
@@ -99,7 +99,7 @@ def __init__(self,
             os.makedirs(self.logfolder)
 
         self.page_delay = page_delay
-        self.switch_language = True
+        self.switch_language = False
```
uluQulu: hm how could you decide to switch this variable?

uluQulu:
@gitpatrickhub 
Just took a quick glance on your code and it seems pretty clean 🏅 good job!
Will try to look into details and take some tests.

Cheers

instapy/instapy.py
gitpatrickhub: @uluQulu instapy fails with it set to True for me, the only method to make it run successfully is to set it to False. I believe it only needs to be set to True if running outside the US.

gitpatrickhub: @uluQulu just realized that there is a set_switch_language function that will allow me to change this setting via the quickstart.py file instead of the instapy.py file. Do I need to make the change to instapy.py for these changes to be merged?

uluQulu: Yes @gitpatrickhub 
You should remove that change from the PR.

uluQulu:
@gitpatrickhub 
I've just took a detailed look at your work and have found this.
gitpatrick

>behave `not_valid_users` carefully, @gitpatrickhub.

If you can solve the problems, it will be great!

instapy/instapy.py
```diff
@@ -1884,6 +1911,19 @@ def interact_user_following(self, usernames, amount=10, randomize=False):
         self.logger.info(
             "--> Interacted total of {} people\n".format(interacted_all))
 
+        #add summary logging to the console
+        liked_img_sum = self.liked_img - liked_img
+        already_liked_sum = self.already_liked - already_liked
+        inap_img_sum = self.inap_img - inap_img
+        commented_sum = self.commented - commented
+
```
uluQulu: `followed_sum` seems to be missing here

uluQulu:
You have done a great job @gitpatrickhub!

I think getting the count of `not_valid` users from class variable and adding it to the count of `not_valid` users per that feature would be much better..
_e.g._,
```python
##start of feature
not_valid_users_initial =self.not_valid_users
not_valid_users = 0

##some activity
not_valid_users += 1  #etc.

##final
self.not_valid_users += not_valid_users
not_valid_users_sum = self.not_valid_users - not_valid_users_initial
print("Not valid users per this feature: ", not_valid_user_sum)
```
This kind of approach will return more up-to-date value of `Not valid users`.

Good luck!

gitpatrickhub:
@uluQulu ok, from your previous comment, I thought you wanted me to use the value directly from the feature. I can make this change now and push a new commit.

instapy/instapy.py
```diff
@@ -1794,6 +1799,21 @@ def interact_user_followers(self, usernames, amount=10, randomize=False):
...
+        not_valid_users_sum = self.not_valid_users - not_valid_users_initial
+
+        self.logger.info('Liked: {}'.format(liked_img_sum))
+        self.logger.info('Already Liked: {}'.format(already_liked_sum))
+        self.logger.info('Commented: {}'.format(commented_sum))
+        self.logger.info('Followed: {}'.format(followed_sum))
+        self.logger.info('Inappropriate: {}'.format(inap_img_sum))
+        self.logger.info('Not valid users: {}\n'.format(not_valid_users_sum))
+
         self.not_valid_users += not_valid_users
```
uluQulu: Great,
```python
self.not_valid_users += not_valid_users
``` 
please bring this line before

```python
#add summary logging to the console
```
so that, while calculating the **sum** value after it, it also can evaluate the amount of `not_valid_users` variable.

You're doing this so quickly, it's amazing!

instapy/instapy.py
```diff
@@ -791,9 +791,8 @@ def like_by_locations(self,
                                                self.logger,
                                                media,
                                                skip_top_posts)
-            except NoSuchElementException as exc:
-                self.logger.warning("Error occured while getting images from location: {}  "
-                                    "~maybe too few images exist\n\t{}\n".format(location, exc.encode("utf-8")))
+            except NoSuchElementException:
+                self.logger.warning('Too few images, skipping this location')
```
uluQulu: _also this change must be reverted back to as it is.._

gitpatrickhub: Hmmm...I didn't make that change. Perhaps I didn't pull the latest instapy.py file. Not exactly sure how that happened. I'll make sure to correct it in my next push.

instapy/instapy.py
gitpatrickhub: ah ok, that makes sense. Thanks for all the help!

uluQulu:
Looks great @gitpatrickhub!

Maybe you can add this support to `follow_user_followers` & `follow_user_following` pairs, eh? 🧐

gitpatrickhub:
@uluQulu I can definitely add the summary, but I was thinking it didn't make too much sense to add totals for Likes, Comments, and Inappropriate Images, since these 2 features only follow accounts.  Whereas, the other 2 features can Like, Comment, and Follow. Let me know your thoughts and if you still think it makes sense to keep it uniform, I will add code and push the commit.

uluQulu:
@gitpatrickhub 
Of course currently, `follow_user_followers` & `follow_user_following` features do have interact capabilities.
Actually they are very similar to `interact_user_followers` & `follow_user_following` after that capability was added.

The only difference is, `follow_user_followers` attempts to interact only after following whereas `interact_user_followers` interacts freely.

uluQulu:
Good job @gitpatrickhub

uluQulu (approved):

PR #2247, opened 2018-06-14
update example

ghost:
there was an error , changed the argument from "random" to "randomize"


timgrossmann:
@ilmetu Awesome catch!
Thank you very much

PR #2248, opened 2018-06-14
modify the chromedriver name in assets directory

ghost:


timgrossmann:
@ilmetu Awesome, thank you very much for the contribution!
Just for my understanding, what is the reason you're needing several versions of the chrome driver on your host system?

ghost:
@timgrossmann hi, 
i have many development environment, in my office i run on windows and osx and at home i run on linux , so i need to switch the chromedriver everytime i move from office to home and viceversa.

So i just use this little trick, i download all the os versions of the chromedriver and save in the assets directory with the different names, and i wrote that small function to identify the os and use the right chromedriver depending on the os.

what do you think?

(sorry for my bad english)

timgrossmann:
@ilmetu I think it's an interesting idea. However I kind of wonder if things go as they do right now when the user does not rename the chromedriver and keeps it at the default name `chromedriver`.

ghost:
you can look at the code [here](https://github.com/timgrossmann/InstaPy/pull/2249/commits/bfc27203b503145162402d730037314a6fb4af0e)

it's very easy code

timgrossmann:
@ilmetu Ahh ok, thank you for pointing it out. 
What is the reason this is a separate PR?

ghost:
@timgrossmann i am sorry this is the first time i modify an opensource project ... my fault ... next time i will do everything in one pull request ... :+1:

timgrossmann:
#2971

PR #2249, opened 2018-06-14
use different chromedriver for os environment

ghost:
if you need more then one chromedriver (for development purpose) you can rename it as :

chromedriver_linux
chromedriver_osx
chromedriver_windows

and you can put all in assets directory

### DIFF
--- a/instapy/settings.py
+++ b/instapy/settings.py
@@ -1,13 +1,34 @@
 import os
 
-
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
+def os_env():
+    from sys import platform
+    if platform == "linux" or platform == "linux2":
+        return 'linux'
+    elif platform == "darwin":
+        return 'osx'
+    elif platform == "win32":
+        return 'windows'
 
 class Settings:
     log_location = os.path.join(BASE_DIR, 'logs')
     database_location = os.path.join(BASE_DIR, 'db', 'instapy.db')
+
     chromedriver_location = os.path.join(BASE_DIR, 'assets', 'chromedriver')
+
+    if os_env() == 'osx':
+        chromedriver_location = os.path.join(BASE_DIR, 'assets', 'chromedriver_osx')
+
+    if os_env() == 'linux':
+        chromedriver_location = os.path.join(BASE_DIR, 'assets', 'chromedriver_linux')
+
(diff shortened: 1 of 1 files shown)


uluQulu:
You have done an amazing job, @ilmetu 
Instead, I suggest once needs to use other webdriver, rename others it to _e.g._ `chromedriver_old.exe` and use new (_desired_) one as `chromedriver.exe`. This approach will avoid confusion cos it is such a basic approach everybody understands.

Or to go further, it could be amazing to have this:
Put multiple (_more than 2_) `webdrivers` into the assets folder and add a new parameter to choose the webdriver, such as:
```powershell
dir ~/assets folder/
chromedriver.exe
chromedriver_old.exe
chromedriver3.exe
chromedriver_latest.exe
chromedriver_best_working.exe
```
have that parameter to do the choosing:
```python
webdriver_selection = "oldest"    ## ["oldest", "best", "latest"]
#or choose directly with the version
webdriver_selection = 3.38
```

Well it can be done by getting the list of available executables in that directory using e.g. `glob` from `glob` module and getting their version running native OS commands using `call` from `subprocess` module.

The difficult is to select the "best" version, it will be a challenge to find it :)
But got some idea to do it, if you like to know.


Cheers

sionking:
@uluQulu 
chromedriver_best_working.exe ?
What did you modify? please tell me and I will tell you :)

uluQulu:
Yep @sionking
It's the idea in my mind for quite a long time since started using InstaPy.
There have to be a mechanism, either written manually or maybe some external source to give the information of the BEST compatible webdriver  version for its web browser.
It's a big issue cos there are lots of updates coming from 2 different sources (webdriver & it's web browser) and from time to time they do have incompatibility issues which cause to real problems.

E.g. as of today there is a live (_unfixed_) problem with the latest Google Chrome web browser and a user from this community hit that problem. It's kind of a beta update which is not supported by chromedriver.
What I mean is, it could be great to get that information of the BEST webdriver & it's web browser combo and have a good mechanism to integrate it into this program.


Well yes, I have done nothing about this matter, it's all in quickly generated words 😄

sionking:
I need this, is it working ?

ghost:
@sionking i use it every day since i started to use instapy

sionking:
Tested and working

PR #2260, opened 2018-06-18
Clarifi url correction

marcomokastyle:


uluQulu:
Charmful adds, @marcomokastyle 
But why removed XPATH layer?
It's class was just changed from `_2di5p` to `FFVAD`.

Cheers

marcomokastyle:
Yes, i noticed that the new class is FFVAD, but when I used with XPATH layer i had a lot of Bad requests in Clarifi (null string), using graphql instead I always got the right url.

uluQulu:
@marcomokastyle 
It's always good to have more options cos there will be times when you will encounter lots of `entry_data` errors from using **graphql** out of temporary outage from servers and there the only help you will get is from XPATH.

Add a new layer to check against empty string, most of similar tools does check it already, e.g. `total_posts` variable in `get_active_users` definition.

>BTW I use `FFVAD` class last few days without a noticeable problem

marcomokastyle:
@uluQulu Ok, I think you're right, is just safer have multiple ways to get the url. I'll made this edit as soon I can and commit again :) just for be sure, the xpath is the same, just changed the class name? 
so, //img[@class = "FFVAD"] is this the new one?

sionking:
What really on with graphql not updating ? I think I don't see it anymore ...

uluQulu:
@marcomokastyle, yes that 's right.
@sionking, right. It is more stable than before but that situation has been happening regularly (_just checkout **Issues** opened on it- I saw from 2017 July to last days_)

uluQulu:
Oops guys using Github in mobile phone is crazy😅
Sorry for closing 😉
>I think @Github must have a verification on closing, reopening, merging, etc.

timgrossmann:
@uluQulu thank you for pointing out the duplication (error handling) part.
@marcomokastyle Could you please add the proposed addition so that if graphql is not working we are trying to access the XPATH for the element, please? 👍 
Thank you very much for this fix

marcomokastyle:
@uluQulu @timgrossmann I'll made this edit as soon as I can, thanks for all this very helpful feedbacks

marcomokastyle:
Just a little bit of patience, still have conflicts, now I sync again with all the latest edits, and try again :)

timgrossmann:
@marcomokastyle Awesome, thank you very much!

marcomokastyle:
Done, let me know if is all ok :)

timgrossmann:
@marcomokastyle Looks good, only one last question.
The part about the `skip tags`, does it also work if the user does not provide any `skip tags`?

Could you maybe also add a short explanation to the part where you show how to use it, please?

marcomokastyle:
@timgrossmann maybe not, this part wasn't written by me, but I'll do some checks

timgrossmann:
@marcomokastyle Ahh good to know that... Saw it the first time, actually 🙈
Thank you very much!

marcomokastyle:
by the way, using an empy list, like [] its working, is the way i'm using right now. But i'll experiment with skip tags, because could be very helpful, i just don't know yet what tags i wanna skip :D.
Based on the position of the parameter i guess if is skipped in clarifai_check_img_for should be generate an error. But if present an empy list, works as aspected.

marcomokastyle:
@timgrossmann @sionking skipping tags I think was originally written by @sionking.

If you want explain better what you have done :) you are the author of this edit so maybe you are the right one to do that :)

timgrossmann:
@marcomokastyle @sionking Ahh good to know that, I really like the idea of it. I just wonder whether we simply need to provide the default value of an empty array in order to make the code more error prune and not force users to provide a skip list

marcomokastyle:
@timgrossmann indeed it could be the solution to the problem.

sionking:
@marcomokastyle 
Hi, skipping tags mean (example) that if you find a man in the picture that you want to ignore you also check the skipping tags, so if woman also found in picture it will not return fail.
tags = man
skip_tags = woman
only man in picture = fail
man in picture but also woman = pass

ghost:
i think we should change variable names to
skip_tags = man
dont_skip_if_contains = woman
only man in picture = fail
man in picture but also woman = pass
@sionking

dfundingsland:
Just wanted to add a bit to this disscusion. You will get an error if, instead of an image, the post is a video. The class Instagram uses for keyframe images of videos is `_8jZFn`which is different from that used on standard image posts (i.e. `FFVAD`). To mitigate the `Unable to locate element` errors, I suggest changing the xpath to `//img[@class="FFVAD" or @class="_8jZFn"]`

PR #2269, opened 2018-06-18
Specify type of action in summary

redsector72:
I find particularly interesting the summary at the end of an action, but I miss the specification of that tipe of action was, so I added it.


timgrossmann:
@redsector72 Did you make sure you've added that logging to every one of the "action"-Methods?

Thank you for your contribution.
Looking forward to future PRs.

Best
Tim

redsector72:
I've added to the methods is possible (according to my understanding of python).
Some methods are not clear to me, so I can't figure out what to log :)

timgrossmann:
@redsector72 Awesome, thank you very much!

PR #2278, opened 2018-06-19
fix follow button check and remove one extra update

sionking:


timgrossmann:
@sionking Nice! Thank you

PR #2279, opened 2018-06-19
Use O(1) in searching within dont_include (to be named whitelist)

sionking:


timgrossmann:
@sionking Yes! Nice and clean fix, love it!
Thank you very much.

Just to be really sure here (I know you most likely did), but did you make sure that the list does not get altered somewhere else in the codebase?

sionking:
I hope not, we shall check it better I think.

timgrossmann:
@sionking Just checked the whole repo, looks like you covered all the places it's altered.
Thank you very much

PR #2286, opened 2018-06-20
make code nicer and reuseable by exporting following check and elemen…

sionking:
…t to function
also remove double remove from line


sionking:
I need to do some changing please don't dommit

PR #2304, opened 2018-06-23
follow pool "all" got the same user all the time

sionking:
since splitting is part of eligible only
also optional pool added to function


sionking:
@uluQulu Is it ok the change ?

timgrossmann:
@sionking Looks good, thank you!

PR #2313, opened 2018-06-25
Added a "Required Tag" feature

dankelley2:
Adds the ability to set one or more tags that must appear in a post to be like-able (via check_links function in like_util). Originally created for a friend that wants to use like_by_location features as well as narrowing the pool of potential posts with required tags. The list of tags is boolean OR, so if any one tag matches, then it passes the filter.

Usage example:

` session.set_required_tag(['river', '#cityofbridges'])`
` session.like_by_locations(['213470852/pittsburgh-pennsylvania/'], amount=200, media='Photo')`

I'm going to have a hearty laugh if there was already an easier way to do this.

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -143,6 +143,7 @@ class InstaPy:
         self.smart_hashtags = []
 
         self.dont_like = ['sex', 'nsfw']
+        self.required_tag = []
         self.ignore_if_contains = []
         self.ignore_users = []
 
@@ -425,6 +426,22 @@ class InstaPy:
             self.aborting = True
 
         self.dont_like = tags or []
+
+        return self
+
+    def set_required_tag(self, tags=None):
+        """Changes the possible restriction tags, if the image does not contain
+         one or more of these tags, the image won't be liked but user 
+         still might be unfollowed. Useful when paired with geo-locoation features"""
+        if self.aborting:
+            return self
+
+        if not isinstance(tags, list):
+            self.logger.warning('Unable to use your set_required_tag '
+                                'configuration!')
+            self.aborting = True
+
+        self.required_tag = tags or []
 
         return self
 
@@ -830,6 +847,7 @@ class InstaPy:
                         check_link(self.browser,
                                    link,
                                    self.dont_like,
+                                   self.required_tag,
                                    self.ignore_if_contains,
                                    self.logger)
                     )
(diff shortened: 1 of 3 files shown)


timgrossmann:
@dankelley2 Thank you very much for your contribution!
This sounds awesome.

As far as I know the only option right now is to ignore posts if they contain something specific.
So there was nothing already implemented. You probably could've somehow used `set_dont_like` and `ignore_if_contains` in some hacky manner to achieve the same, but this is way cleaner.

However, I'd find it more interesting and appealing if it was, let's call it, more general.
Only requiring tags might not be the only thing users want to do. 
Having something like `set_do_like` which does the same like the `set_dont_like` just the other way around would probably be nicer.

Let's wait and see what the other contributors say.

Thank you very much for this contribution!

dankelley2:
@timgrossmann Thanks! This definitely helped us search very general locations for something very specific without hacking a solution together.

I agree that something more general would be more appropriate, I only avoided 'set_do_like' because that might sound like a *must have* or *important* parameter for potential users, confusing someone if their use case doesn't fit this specific template.

I would gladly change this to set_do_like or whatever else fits the most common use case! If this is to be a general option, it might be prudent to add a use_smart_tags kwarg (since this is a logical OR, and other similar tags would allow you to *slightly* broaden the match), or even something like type='OR'/type='AND'.

I'm very new to creating pull requests for others' projects so let me know if this is something that I should have brought up in the Issues section first. 

Thanks again!

**edit:** And if it didn't seem like I was new enough, I just accidentally closed this!

CharlesCCC:
@dankelley2 Thank you for the contribution, this feature is awesome. I bet a lots more people would love to have it.  github/git take couple practice to get the handle of it, but I think you already did :) 

The change looks good for the most part, one thing I would suggest you to add is the `Read.me` instruction. Can you please add this use case(e.g.) so people can learn how to use it in the future. (don't want this feature get merged in without nobody noticed it).

CharlesCCC:
also, there is merge conflicts need resolved, you can resolve it by merge latest master branch to your feature branch and then push up all the change.

timgrossmann:
@CharlesCCC Thank you very much for the feedback.
I'm not really sure if we should keep this open since we merged #2612 a while ago which basically does something similar.

What do you think?

CharlesCCC:
close this PR as similar function has already been implemented #2612

PR #2336, opened 2018-07-01
avoid infinite loop when "get_links_from_feed" return 0

ghost:
i modify the while loop to check the numbers of the obtained links.
if the function return consecutively 0 for almost 10 times the loop will terminate.
in my case i stuck on the loop for hours before i realize that ...


sionking:
Thanks ! I was stuck too but had no time to investigate

sionking:
Thanks ! I was stuck too but had no time to investigate

Issue #2322, opened 2018-06-26
After update, session.end doesn't close Chrome and CPU is 99%  (no-sandbox option related?)


ortegafernando:
## Expected Behavior
session.end() would close and end session.
After that, python would close cmd window and chrome close perfectly.
## Current Behavior
I can see chrome closed (no windows), but I can see chrome in tas manager, using 99% of CPU
## InstaPy configuration
maybe related with chromedriver? I am using last version of chrome, chromedriver and Instapy and Selenium

Thanks

converge:
is your script setup using the try/catch approach ?

ortegafernando:
Hi @converge 
This is my script:
```
def main():
	try:
		.....
		session.logger.info('Finish')
		session.end()
	except Exception as e: 
		print('Error: {}'.format(e))
		if need_send_email:
			send_email(.....)		
```

I am calling the script by Window's task manager (cmd /c), and after `session.logger.info('Finish')` I could see also the "session end" message in log file.
I haven't changed anything in my script since a lot of time, so I suppose that is not script related. I don't know. 
May be "session.end()" doesn't close properly ?

But after that, I can see the cmd windows still open (as "cmd /c python ..." doesnt finish), chrome closed, but still in task manager using 99% of cpu.

I don't know where to start to get more information about this problem. It seems to be related with chrome driver (2.40v), Chrome (last version) or  ......

Any clue?
Thanks

sionking:
I suffered a lot from this in Windows.
You should turn off the no-sandbox option for chrome (why we need it ?)

ortegafernando:
> I suffered a lot from this in Windows.
You should turn off the no-sandbox option for chrome (why we need it ?)

Hi, I can't find sandbox option in chrome. I have found these settings that may be related to.. 

Continue running background apps when Google Chrome is closed: is ON by default
Use hardware acceleration when available: is ON by default

May be any of two is related to ?

Thanks a lot

sionking:
in the code at __init__ there are chrome pref comment the line for sandbox:
file: instapy.py

ortegafernando:
Hi @sionking , in instapy.py I have this:

```
chromedriver_location = Settings.chromedriver_location
chrome_options = Options()
#chrome_options.add_argument("--disable-infobars")
chrome_options.add_argument("--mute-audio")
chrome_options.add_argument('--dns-prefetch-disable')
chrome_options.add_argument('--no-sandbox')
[...]

PR #2347, opened 2018-07-03
Solve issue 2322

ortegafernando:
See [https://github.com/timgrossmann/InstaPy/issues/2322](https://github.com/timgrossmann/InstaPy/issues/2322)

PR #2363, opened 2018-07-04
#2322: no sandbox option in settings

douwevandermeij:
Revert --no-sandbox option but make it configurable in Settings module.


timgrossmann:
@douwevandermeij I think we'll go with the proposed way of xvx29a here #2347.
Seems like the sandbox is only needed for headless browser usage

douwevandermeij:
Sounds good.

PR #2364, opened 2018-07-04
update sleep time for slower connection

ghost:
i modify a sleep time before and after click on unfollow for slower connections to show up and close the popup for unfollow confirmtion

### DIFF
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -282,7 +282,7 @@ def unfollow(browser,
 
                         try:
                             browser.find_element_by_xpath("//button[contains(text(), 'Unfollow')]").click()
-                            sleep(1)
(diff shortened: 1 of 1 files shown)


timgrossmann:
@ilmetu Sounds reasonable, thank you

marcomokastyle:
@timgrossmann  just an idea, and I don't know yet how to do it, but, if we somehow measure the speed of connection and create sleep time proportional to a ratio, taken from a speed test?

timgrossmann:
@ilmetu It's a really good idea, I thought about that some time before already. However the problem is that connection speeds might also fluctuate so we should really get a good mean estimation for it.

If you have any ideas on how this could be implemented (best without any big external dependency), I'd be eager to hear about it.

Thank you very much

PR #2366, opened 2018-07-04
better code for identify like button

ghost:
better code for identify like button


marcomokastyle:
@ilmetu it seems that is exactly what i suggested too.

timgrossmann:
@ilmetu Awesome! Thank you

PR #2374, opened 2018-07-04
align logs to correct place

sionking:
per location print


sionking:
logic is all wrong in this, we shall print the whole list and not only the last tag/location.

PR #2401, opened 2018-07-09
Updated Unfollow to remove unfollowed or removed IG accounts from followedpool.csv

ximhotep:
Accounts removed by IG aren't being removed from followedpool.csv causing unfollow to iterate through those accounts every cycle.  Recommended adding the two additional calls to delete_line_from_file to remove users from followedpool.csv


sionking:
This is not the right place to fix.
the get_following_status function return the status, used by other function.
The solution is to retrun also follow_)element as None. I will add the code here:

sionking:
I did some modification ! thanks for contributing !

PR #2402, opened 2018-07-09
Updated Unfollow to remove unfollowed or removed IG accounts from followedpool.csv; cancellation of ctrl+c when updating a file

sionking:
Thanks to @ximhotep

Accounts removed by IG aren't being removed from followedpool.csv causing unfollow to iterate through those accounts every cycle. Recommended adding the two additional calls to delete_line_from_file to remove users from followedpool.csv

** This is a diffrent fix based on the return following is False, so we will go to the else.

----------------
Added also the cancellation of ctrl+c when updating a file

### DIFF
--- a/instapy/print_log_writer.py
+++ b/instapy/print_log_writer.py
@@ -61,7 +61,8 @@ def log_followed_pool(login, followed, logger, logfolder, logtime):
     a seperate file"""
     try:
         with open('{0}{1}_followedPool.csv'.format(logfolder, login), 'a+') as followPool:
-            followPool.write('{} ~ {},\n'.format(logtime, followed))
+            with interruption_handler():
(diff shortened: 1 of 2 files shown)


timgrossmann:
@sionking Looks like `interruption_handler()` is not defined in that context which leads the travis-ci checks to fail.

sionking:
Fixed...

sionking:
I do think we should change the order:
first call interruption_handler later open file.

timgrossmann:
@sionking Awesome, now all the checks and the CI work 👍 

What's the reason you would prefer the order to be changed?

sionking:
- open file than ctrl + c == not closing the file right.

but I don't really think it matter.

timgrossmann:
Now it makes sense, awesome addition to InstaPy, @sionking!
Thank you very much

Issue #2295, opened 2018-06-21
Document all(most) methods/functions in machine readable json.


Malusovium:
This will simplify thirdparty integration, for people like me who want to hack on it without the need
for the project to be maintained everyday. Good for projects like: electron-instapy-gui.
This should then enable gui and documentation generation.

By means of a json file (api.json?) where all the methods and input(args) are documented.
Maybe the json is generated from the code? Should produce less maintenance that way :smile: 

example implementaiton:

// api.json 
```json
{
  "git-hash":  "<commit-hash>",
  "api":
  {
    "methods":
    [
      {
        "name": "set_do_comment",
        "description": "Wheter to enable or disable comments and specify the frequency.",
        "args":
        [
          {
            "name": "enabled",
            "type": "Boolean",
            "default": false,
            "description": "Either enabled or disable do-comments",
          },
          {
            "name": "percentage",
            "type": "Int",
            "description": "Percentage, e.g. '25' means that every /4 will be commented.",
          }
        ]
      },
      {
        "name": "set_comments",
        "description": "List of comments to randomly comment when a comment is invoked.",
        "args":
        [
          {
            "type": "Array<String>",
            "description": "List of comments"
          },
          {
            "name": "media",
            "type": "Tuple<'Video'|'Photo'>",
            "description": "Set comments only on a specific media type."
          }
        ]
      }
    ]
  }
}
```

PR #2418, opened 2018-07-13
WIP: Documented api in machine readable JSON

Malusovium:
This an how api.json could be constructed, now generated from README.md.
As the described in: #2295.
Only the "Instapy features section".

### DIFF
--- a/api.json
+++ b/api.json
@@ -1,0 +1,725 @@
+{
+  "args": [
+    {
+      "name": "enabled",
+      "type": [
+        "Boolean"
+      ],
+      "examples": [
+        "True",
+        "False"
+      ]
+    },
+    {
+      "name": "percentage",
+      "type": [
+        "Number"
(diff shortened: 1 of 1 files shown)


timgrossmann:
@Malusovium This is an awesome idea!
I thought of something similar a while ago when the GUIs started to be abandoned. 

Like you stated, this would be a great thing to be used as a base to build a GUI-Creator on top. 
This would remove the problem with updating the elements of the GUI by the GUI-Dev if done right.

Kudos to you for setting this up.

Do you think you will provide your GUI tool to the community at one point? 😉

Malusovium:
Didn't thought about making a Instapy-gui creator lib, Great idea!
Will work on it at: Malusovium/instapy-tools.

timgrossmann:
@Malusovium Epic, will follow that around!
Thank you very much

PR #2426, opened 2018-07-14
Faster way to check if user is private

mehdichekori:
Updated how we check if the user is private because it took 25 seconds with the find_element_by_xpath function.
It takes less than a second with this edit.

### DIFF
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -324,7 +324,8 @@ def get_links_for_username(browser,
     abort = True
 
     try:
-	is_private = browser.execute_script("return window._sharedData.entry_data.ProfilePage[0].graphql.user.is_private")
+	is_private = browser.execute_script(
+		"return window._sharedData.entry_data.ProfilePage[0].graphql.user.is_private")
     except:
         logger.info('Interaction begin...')
     else:
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -324,8 +324,7 @@ def get_links_for_username(browser,
     abort = True
 
     try:
-        is_private = body_elem.find_element_by_xpath(
-            '//h2[@class="_kcrwx"]')
+	is_private = browser.execute_script("return window._sharedData.entry_data.ProfilePage[0].graphql.user.is_private")
     except:
         logger.info('Interaction begin...')
     else:


sionking:
There use to be a bug not having the graphql correct or exist.
I anyway use this method long time ago and it working.

timgrossmann:
@mehdichekori @sionking I think we should, again, have a combination of both.

The graphql checking is way faster but it might not return the wanted result. If that fails we should have a backup and still use the find_element part.

@mehdichekori Could you please add that and make sure that all the CI-Checks work?
We'll merge this then 👍 

Thank you very much for this

sionking:
We need to make a function to check if a user is private, and maybe instead of using the xpath fin we can try locate the "physical place" it should be much faster.

```
def is_private_profile(browser, logger, following=True):
   is_private = None
    is_private = browser.execute_script(
        "return window._sharedData.entry_data."
        "ProfilePage[0].graphql.user.is_private")
    # double check with xpath that should work only when we not follwoing a user
    if is_private is True and not following:
        body_elem = browser.find_element_by_tag_name('body')
        is_private = body_elem.find_element_by_xpath(
            '//h2[@class="_kcrwx"]')
 
    return is_private
```

uluQulu:
Hey @mehdichekori 
Thanks a lot for this catch!
I always noticed that big time during that phase but couldn't imagine that finding XPATH would be so long.
Then checked time, it is taking more than 25 seconds and the graphql approach you proposed takes only 0.015 seconds.
This is gonna be one of the most efficient changes so far!
Thanks again and hope this gets merged soon.

sionking:
I posted this time duration long long time ago... in issues

HCWcoder:
works perfect for me

timgrossmann:
@mehdichekori @sionking @uluQulu Incredible speed up! Love it 👍 
Now we only need to modify and merge it.

Thank you all for the hard work on our project

bianpratama:
@mehdichekori @sionking @timgrossmann,

It works perfect for me 🎉 

Initially, it always showed:
```
INFO [2018-07-29 03:12:27] [username]  Getting username image list...
INFO [2018-07-29 03:12:52] [username]  Interaction begin...
ERROR [2018-07-29 03:13:17] [username]  Element not found, skipping this username
```

And I made the change from:
```python
is_private = body_elem.find_element_by_xpath(
     '//h2[@class="_kcrwx"]')
```
to:
```python
is_private = browser.execute_script(
    "return window._sharedData.entry_data."
         "ProfilePage[0].graphql.user.is_private")
```

Now, it works smoothly yet it doesn't show the `Interaction begin...` info log. But I think it doesn't matter, does it? 🙏

HCWcoder:
@bianpratama 

this is working really fast for me
Thanks for sharing !

timgrossmann:
@mehdichekori Do you think you can fix the CI-fail and adjust your check to use the, by @sionking proposed, changes?

Thank you very much

guivinicius:
Looks like a easy fix

```
./instapy/like_util.py:327:115: E999 TabError: inconsistent use of tabs and spaces in indentation
	is_private = browser.execute_script("return window._sharedData.entry_data.ProfilePage[0].graphql.user.is_private")
                                                                                                                  ^
1     E999 TabError: inconsistent use of tabs and spaces in indentation
```

timgrossmann:
Unfortunately it seems that @mehdichekori is not able to do the changes so there is a new PR #2609

@mehdichekori @sionking Thank you for the work

PR #2429, opened 2018-07-15
Update index.md

jvanremoortere:
Links get double "docs" folder


timgrossmann:
@goeroe Good catch, thank you very much!

PR #2440, opened 2018-07-17
Set of efficient revisions and fixes

uluQulu:
Hey guys,
Tonight could find some time and decided to share some useful bits with the community.
#### The changes are stated in the commit names clearly.  

* Sorry for being late, @timgrossmann, but finally could prepare the promised change regarding _simulation_ and also improved it a little bit.

> As it's a quick preparation, will be great to have a look and ask questions on any unclear line.

Cheers 😁

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,8 @@ Table of Contents
   * [Windows Task Scheduler](#windows-task-scheduler)
   * [cron](#cron)
   * [Schedule](#schedule)
-* [Extra Informations](#extra-informations)
+* [Extra Information](#extra-information)  
+  * [Simulation](#simulation)
 
 ## Getting started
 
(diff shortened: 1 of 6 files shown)


sionking:
Welcome back.
The only reason I canceled the simulation is that unfollow_user don't look for pop-up confirmation window.
We need to make one function for unfollowing a person and not 2 (list vs one person).

thank.

uluQulu:
Thank you, @sionking!
That's right, `unfollow_user` should be integrated into `unfollow` definition.

timgrossmann:
@uluQulu Awesome to see you're still active with InstaPy! Your changes always make InstaPy way better.
Thank you for adjusting the part with the simulation, now people have an easy way to turn it off if they definitely don't want to follow any account.

Documentation is clear as well, great one!

Thank you
Best
Tim

PR #2457, opened 2018-07-18
fix #1589

### LINKED ISSUE #1589: [Question] Adding sleep delay to "Follow by Tags"
## Question/Feature request

Would it possible to integrate the same delay used in the "follow user followers" function?
By this I mean, since the follow user followers function stops @ every 10 users, it sleeps for a set amount of time. Would it be possible to add that same rule to the "follow by tags" function? That way we don't risk getting banned for following too many users, therefore we can  avoid surpassing the Instagram threshold.

zlita:
this add the feature requested by @walachewaka (issue #1589).

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -791,7 +791,7 @@ class InstaPy:
         if (enabled not in [True, False] or
                 percentage not in range(0, 101)):
             self.logger.info("Invalid simulation parameter! Please use correct syntax with accepted values.")
-        
+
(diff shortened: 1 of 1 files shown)


bgunsberger:
Thanks, but it would be better if it took the same sleep parameter as the other follow methods.

timgrossmann:
@zlita Are you able to adjust your PR to the changes @bgunsberger mentioned?

Issue #2359, opened 2018-07-03
Unfollow user that no longer exists -> behaviour analysis


joanroig:
After some checks, I can see the followedPool.csv full of non-existing users. 
## Expected Behavior
If a user does not longer exists, or its name has changed, I think it should be removed from the followedPool. It will prevent having a dead followedPool full of users on top of the list that will be checked on every run.
## Current Behavior
The bot shows that the user no longer exists and skip it, leaving the user in the followedPool.
## Possible Solution (optional)
Original code from unfollow_util.py:

```
except:
    logger.error(
        '--> Unfollow error with {},'
        ' maybe no longer exists...'
            .format(person.encode('utf-8')))
    continue
```

Changes:
```
except:
    logger.error(
        '--> Unfollow error with {},'
        ' maybe no longer exists...'
            .format(person.encode('utf-8')))

    delete_line_from_file('{0}{1}_followedPool.csv'.format(logfolder, username),
                          person + ",\n", logger)
    print('')
    sleep(2)
    continue
```

## InstaPy configuration
Try to unfollow a non existing user with the quickstart.py file

HansDaigle:
Just pointing out that a check is in O(n) even if the check is N+1 it wont really impact the performance. But I agree that the expected behavior would be to delete the user from the list! 👍

dbmathis:
I have this same issue, however, your code above does not seem to help me, if I am experiencing the same issue.  The code iterates through hundreds of people I have already unfollowed:

Or do I need to manually remove these and the code you wrote above removes them going forward?

```
WARNING [2018-07-04 08:52:33] [xx]  --> Cannot Unfollow From InstaPy 0, maybe zz1 was unfollowed before...

WARNING [2018-07-04 08:52:38] [xx]  --> Cannot Unfollow From InstaPy 0, maybe zz2 was unfollowed before...

WARNING [2018-07-04 08:52:43] [xx]  --> Cannot Unfollow From InstaPy 0, maybe zz3 was unfollowed before...

WARNING [2018-07-04 08:52:48] [x]  --> Cannot Unfollow From InstaPy 0, maybe zz4 was unfollowed before...
```

joanroig:
@dbmathis you are experiencing a different error, it says that the user was probably unfollowed, so those users exist on Instagram. The problem seems the same, but in another part of the code, look for "was unfollowed before" and add the same delete_line_from_file after the print, as the same as I've done in the possible solution.

dbmathis:
[...]

PR #2459, opened 2018-07-18
Delete user from followedPool if no longer exists (account suspended / removed / name changed)

joanroig:
[Discussed here: #2359] Delete user if followed with InstaPy, and when trying to unfollow it doesn't exist and remains in the followedPool.
Removed some unnecessary encodes (tested for weeks with no issues).

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1027,7 +1027,7 @@ class InstaPy:
                             already_liked += 1
                     else:
                         self.logger.info(
-                            '--> Image not liked: {}'.format(reason.encode('utf-8')))
+                            '--> Image not liked: {}'.format(reason))
                         inap_img += 1
                 except NoSuchElementException as err:
                     self.logger.error('Invalid Page: {}'.format(err))
@@ -1209,7 +1209,7 @@ class InstaPy:
                             already_liked += 1
                     else:
                         self.logger.info(
-                            '--> Image not liked: {}'.format(reason.encode('utf-8')))
+                            '--> Image not liked: {}'.format(reason))
                         inap_img += 1
                 except NoSuchElementException as err:
                     self.logger.error('Invalid Page: {}'.format(err))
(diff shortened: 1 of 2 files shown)


sionking:
I can tell you, you ahve many issues with this.
- first you will never reach this line since the function have error handling inside.
- What if instagram web is down ? you will remove a valid user.
There are so many options that you will remove a vaild followed user!

joanroig:
@sionking Hi, thanks for the review. Well, I've tested this line and checked during weeks that is reached when the user page is not available. 

I cannot confirm what happens when Instagram web is down, but I suspect it will throw a timeout because will not retrieve data (not tested). Have you got a better idea on this?

As said, I've tested this with some accounts where a lot of lost follows were stuck in the followedPool, occasionating an accumulation of accounts that makes the bot lose time when unfollowing (and if you try to unfollow X users on different bot sessions every Y minutes, you'll be trying to unfollow the same users forever if the number of lost follows is greater than X, which is a major problem).

ortegafernando:
HI,

> What if instagram web is down ? you will remove a valid user.
There are so many options that you will remove a vaild followed user!

May be, if you could put this as an option, it will be easy to run it one a month and avoid the risk ?

regards,

shariqmoosa:
Hi I am facing the same issue:

ERROR [2018-08-29 13:31:54] [them] --> Unfollow error with b'emilykonopka', maybe no longer exists....
ERROR [2018-08-29 13:31:54] [them] --> Unfollow error with b'emilykonopka', maybe username has changed or he/she blocked you...
ERROR [2018-08-29 13:31:54] [them] --> Unfollow error with b'emilykonopka', maybe no longer exists...

I have tried the above code and got this error:

File "quickstart-merch.py", line 1, in 
from instapy import InstaPy
File "[path]\instapy_init_.py", line 2, in 
from .instapy import InstaPy
File "[path]\instapy\instapy.py", line 44, in 
from .unfollow_util import get_given_user_followers
File "[path]\instapy\unfollow_util.py", line 287
except:
^
TabError: inconsistent use of tabs and spaces in indentation

Please help I am new to Python and GIT

PR #2478, opened 2018-07-19
fix get_links_for_username always shuffling the order

sionking:
while trying to remove duplicated previous method (list(set(...
Reordered the user links.
Fixed.

### DIFF
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -15,6 +15,7 @@ from .util import add_user_to_blacklist
 from .util import click_element
 from .util import web_adress_navigator
 from .util import get_number_of_posts
+from .util import remove_duplicated_from_list_keep_order
 
 
 def get_links_from_feed(browser, amount, num_of_search, logger):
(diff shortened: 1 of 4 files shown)


instapy/like_util.py
```diff
@@ -355,14 +356,13 @@ def get_links_for_username(browser,
 
     while len(links) < amount:
         initial_links = links
-        browser.execute_script(
-            "window.scrollTo(0, document.body.scrollHeight);")
+        body_elem.send_keys(Keys.HOME)
```
uluQulu, line 359: hmm, why needs it?

uluQulu:
Hey @sionking 
_I noticed it after a while and I was gonna share it in my next PR but you're fast :D_

But instead of that new function, you can use `sorted` to remove duplicates while preserving the order **faster** and **simpler** 😅
```python
links = sorted(set(links), key=links.index)
```
>Side **note**: list comprehension (_you have used_) always consumes more resources.  

Cheers 😁

sionking:
Ok, first thing I found on google.
Lets merge this and you can fix it in your next PR.

Issue #2484, opened 2018-07-20
Getting user image list


Tachenz:
On interacting with other user followers - 

Script just stops randomly after message:
“Getting “username” image list...
Interaction begings...”

And then nothing happens

torbenberger:
same for me

torbenberger:
@uluQulu could you have a look into this? i saw that you worked on the get_links_for_username function in like_util.py 4 days ago

i think this problem occours when the desired amount of liked images is higher than 0 but lower than the available amount of images of the user.

the while loop will only break by its own once len(links) is >= the desired amount.

if this is not possible because there are not enoug images there is only one other place where it breaks, and that is inside this `if len(links) == len(initial_links):` if clause. but initial_links is always empty and links is only empty if there are no posts at all.

so both of these break conditions are not met when there are less pictures available than desired but not 0

torbenberger:
added PR for this

Tachenz:
Amazing - will try this tonight!

Tachenz:
Not sure if it is related at all - but a similar problem, when user is private (and hence there are no pictures at all) it also spends quite a bit if time on it, returning “page does not contain a picture” for about 10 times ish...

uluQulu:
Hey @torbenberger 
I think the logic is correct I used in there.
At worst case, it will break after the 7th attempt (_7*0.66 seconds_) in case of either there are less posts than it is written in `Posts count` in there (_regardless of the amount you gave_) **OR** in situations like request hangs causing scroll or sth else misfunctional.

_But if you clearly explain the problem you face I will understand it much better. But I personally use it right now without problems._

@Tachenz, all I have done in my latest PR changing some lines in that function WAS to make it return the desired amount of links all the time and I did not touch anything other there.
The problem you have outlined in upper comment is related to the _check_ of `private`ness.

There is a live PR touching that issue of checking `private`ness, so I will not talk much, but for now, you can do this:
**replace**
```python
is_private = body_elem.find_element_by_xpath(
    '//h2[@class="_kcrwx"]')
```
**with**
```python
is_private = browser.execute_script(
    "return window._sharedData.entry_data."
        "ProfilePage[0].graphql.user.is_private")
```

Tachenz:
Thank you @uluqulu - do you mean to replace it in that PR (for privateness)?

As for the original topic - it just hangs on “interaction begins” for hours...   Even though user has about 10 posts ...

uluQulu:
[...]

PR #2486, opened 2018-07-20
FIX: intial links in get_links_for_username

torbenberger:
added initial links to array in get_links_for_username function Issue: #2484

@uluQulu could you check if i am missing smth?


uluQulu:
Your work is correct @torbenberger.
Although I do not see the need of that change, I have outlined my points in #2484

Cheers 😁

torbenberger:
sorry, i missed smth. you are right, it's not needed. but still its hangig. didnt had a chance to debug it. the problem is that it only occures in very few occasions... so would have to debug through a lot of users until one fails..

PR #2493, opened 2018-07-21
FIX: variable reference issue in get_links_for_username

torbenberger:
Issue: #2484

Fixed by @uluQulu


timgrossmann:
@torbenberger @uluQulu Awesome, thank you very much!

Issue #2271, opened 2018-06-19
Could be an UTF8 Bug in don't like regex structure?


redsector72:
I am not that good in python, but I suppose there is a bug in the don't like check.
If I have a tag like "#snapchat👻" it seems to quit without any warning.
I leave this check to you.
If the problem is not in there, I will try to debug better.

jeremycjang:
https://github.com/timgrossmann/InstaPy/blob/master/README.md#emoji-support

redsector72:
@jeremycjang I am supposing there is a bug, I have no problem with emoji.
If there is a tag in the comments of a picture, and the tag include an UTF-8, like in the example I have shown, Instapy quits.
I am not able to debug where and why quits.

the-unknown:
same here, can confirm this bug. 
When a hashtag includes an emoji or an UTF-8 character, InstaPy quits. 
I will post the exact error, next time I run into it.

the-unknown:
This is the error I get:

Traceback (most recent call last):
  File "./quickstart.py", line 24, in <module>
    session.like_by_tags(['stayandwander], amount=100, media=None)
  File "/home/user/InstaPy/instapy/instapy.py", line 1315, in like_by_tags
    '--> Image not liked: {}'.format(reason.encode('utf-8')))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 53: ordinal not in range(128)

timgrossmann:
The issue is bound to your terminal output and the standard decoding that is set (seems like Ascii). 
If you change the decoding, it should work.

kushpvo:
If anyone can help by explaining how to change the decoding in terminal output?
My Rasp Pi is setup as UTF-8 but the terminal is still outputting ASCII.

PR #2494, opened 2018-07-21
Fix a bug: printing emoji encoded with utf-8

### LINKED ISSUE #1830: UnicodeDecodeError: ASCII code can't decode byte...
Continuously getting this error. Any idea?
CentOS 7.x, Python 2.7

  File "./quickstart.py", line 39, in <module>
    session.like_by_tags(['tag1', 'tag2', 'tag3', 'tag4'], amount=100)
  File "/home/my_name/InstaPy/instapy/instapy.py", line 987, in like_by_tags
    '--> Image not liked: {}'.format(reason.encode('utf-8')))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xf0 in position 49: ordinal not in range(128)
`

IdanDavidi:
When there is an emoji in an inappropriate tag, the app crashes. Now it is fixed.

Picture of the bug:
image

closes #1830, #2271.

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1331,7 +1331,7 @@ class InstaPy:
                             already_liked += 1
                     else:
                         self.logger.info(
-                            '--> Image not liked: {}'.format(reason.encode('utf-8')))
+                            '--> Image not liked: {}'.format(reason))
                         inap_img += 1
                 except NoSuchElementException as err:
                     self.logger.error('Invalid Page: {}'.format(err))


timgrossmann:
@IdanDavidi This does not fix the issue, the issue is bound to your terminal output and the standard decoding that is set (seems like Ascii). 
If you change the decoding, it should work.

Removing the encoding part only leads to more issues

Thank you very much for helping out and contributing

Apao123:
@IdanDavidi - hello, i am new to python. Can you show me the code how to solve this? Thanks a lot!

IdanDavidi:
@Apao123 You better try to upgrade to python3 first, apparently this bug fix is relevant only for python2

Issue #2539, opened 2018-07-26
interaction not work anymore due to an instagram change to the DOM


ghost:
Hi, this morning instagram change the DOM again :

Invalid Page: Message: no such element: Unable to locate element: {"method":"xpath","selector":"//article/header/div[2]/div[1]/div/a"}

it looks like the interaction doesn't work anymore.

the code isn't working is this:

`username = (self.browser.find_element_by_xpath('//article/header/div[2]/div[1]/div/a'))`

i try to fix it as soon as possible.

ghost:
it looks like this works:

`username = (self.browser.find_element_by_xpath('//article/header/div[2]/div/div[1]/a'))`

PR #2540, opened 2018-07-26
update getting username

ghost:
Due to a DOM modification by instagram , the username of the user to follow/commenting cannot be retrived.

please see my issue here : https://github.com/timgrossmann/InstaPy/issues/2539

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1241,7 +1241,7 @@ class InstaPy:
                                 username = (self.browser.
                                     find_element_by_xpath(
                                         '//article/header/div[2]/'
-                                        'div[1]/div/a'))
+                                        'div/div[1]/a'))
 
                                 username = username.get_attribute("title")
                                 name = []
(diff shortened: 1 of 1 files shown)


timgrossmann:
@ilmetu Looks good to me, let's wait for at least one more approval though. 
Thank you for contributing!

sionking:
this kind of xpath are disasters.

ghost:
@sionking do you mean i can do it better ? 
i just update the xpath, to make it work again, i didn't change the logic ...
if you want you can improve it or let me know how to do better.
thanks

sionking:
Not blaming you, but I think we should have more generic xpath for elements and not physical structure.
Also we have to maintain one page with all xpath with description for each.

ghost:
@sionking i think that also , with a little bit of refactoring this will be very helpful for other instagram future changes ...

timgrossmann:
@sionking You're absolutely right!
I thought of maybe doing a diff on every opening of a page (only first time of each run) and then comparing the used paths whether they are still the same or if they already changed. 
If yes, we could update the, external, table of elements which are used to find certain elements.

What do you guys think about this?

sionking:
@timgrossmann 
This could be the best thing this application will ever have.
This way the bots will not fail during run time, but only at init.

Can someone find time for this ?
1. xpath and structure global class under settings file.
2. test page before every run start (and maybe also during run)


Thanks.

PR #2543, opened 2018-07-26
Output correction

vleliseev:
I suppose it should be "was sent" instead of "wast sent".


timgrossmann:
@shmel3 Nice catch!
Thank you very much for the fix

And congrats for your first contribution! 🎉

Issue #2526, opened 2018-07-24
 likers_from_photo always returns zero users


meinert:
## Expected Behavior
When calling the method (from session.follow_likers(....)) with a url of an image that have many likes I expect that the requsted amount of users are returned.

## Current Behavior
The method always returns zero users

## Possible Solution (optional)
The error is related to that this line always returns a list of length 0
`       follow_buttons = dialog.find_elements_by_xpath("//div/div/span/button[text()='Follow']")"`

## InstaPy configuration
```
session.set_relationship_bounds(enabled=True,
                                    potency_ratio=-1.0,
                                    delimit_by_numbers=True,
                                    max_followers=4590,
                                    max_following=5555,
                                    min_followers=40,
                                    min_following=50)
    session.set_do_comment(False, percentage=10)

    session.set_dont_include(dont_include)
    session.set_dont_like(['pizza', 'girl'])

    session.set_user_interact(amount=1, randomize=True, percentage=50, media='Photo')

    session.follow_likers(['salomon_danmark', 'northcoastultra', '9000running'], photos_grab_amount=2, follow_likers_per_photo=3, randomize=True,  sleep_delay=6, interact=False)
```

meinert:
I have create a solution for this error. A simple change to an xpath......would like to commit to a branche and make a pull request, but do not seem to have the rights to commit a new branche. How do you normal handle this, @timgrossmann ?

HCWcoder:
@meinert i tetsed the change you made

return result 👎 

Getting likers from user:  example

shuffling links
Got  12 , returning  1  links:  ['https://www.instagram.com/p/example/']
opening likes
Scrolling down...  22 -> 0  /  25
Scrolling finished
Like window closed
Got  0  likers shuffled randomly, who you can follow:
 []

meinert:
Works fine here. Could someone else test my suggested change? And confirm wheter it is working...

rangabears:
I applied the edit on both line 283 and 294 of commenters_util.py and it seems to be working again for me.

swznd:
You should edit line 298 too after deleted  a path in 282 and 293 

check this https://github.com/timgrossmann/InstaPy/pull/2476/

meinert:
Think we are all done with this

PR #2556, opened 2018-07-27
Update xpath in likers_from_photo method

meinert:
Relates to #2526


timgrossmann:
@meinert @joanroig Is this a global change or does this only occur for some accounts?

joanroig:
@timgrossmann testing it on four independent accounts that were failing for over 2 days, today after those changes all seems to be back to normal.

PR #2563, opened 2018-07-29
Dedicated Data Base to provide robust 'record activity' and 'follow restriction' and support future DB operations easily

uluQulu:
Hey there! 🙂

#### The work I talked about has been written and I introduce the new set of inner tools and also the agreement about the bonds of follow restriction has also been accomplished.  

_I will talk about the major changes coming with this PR._

**1-)** Now there is a database engine which is making a fresh new database at request
+ **previously**
instapy.db (_came with the repo_)
____________________________________________________________________________________________________________
|           _TABLES_           |                                                 _COLUMNS_                                                        |
| **statistics**                |  `likes`, `comments`, `follows`, `unfollows`, `server_calls`, `created`                      |
_____________________________________________________________________________________________________________
+ **now**
instapy.db (_written at request_)
____________________________________________________________________________________________________________
|           _TABLES_          |                                                 _COLUMNS_                                                        |
| **profiles**                 |  `id`, `name`                                                                                                         |
| **recordActivity**       |  `profile_id`, `likes`, `comments`, `follows`, `unfollows`, `server_calls`, `created`  |
| **followRestriction**  |  `profile_id`, `username`, `times`                                                                          |
____________________________________________________________________________________________________________

**2-)** Priorly called **instapy.db** file held `statistics` table which is now renamed to `recordActivity` and the content of that table is the same, except a new column- `profile_id`.

**3-)** There is only 1 database- **instapy.db** holding three tables `profiles`, `recordActivity`, `followRestriction`. Please see the **database_engine.py** file to see their column types and more.

**4-)** `profiles` table is holding the names of the unique users (_"`owner`" users- the users who started a **quickstart** script_) and their `id` is being **referenced** by other tables. This kind of referencing saves up tons of space , provide clearer structure and powerful functionalities and some  speed.

**5-)** The database engine is written to handle every worse situation.  e.g.,
+ you can remove a database at the time program running, there, it will make a database again
+ you can give a custom location of database using Settings class and there if the location does not contain a database, then it will make a new database
+ etc.

**6-)** Now follow restriction is fully handled by the realtime database data which is being written one by one and has no any data leak.

**7-)** To provide speed in follow restriction querying, I used a good `WHERE` clause which is fetching only for the _username_ rather than fetching whole data and comparing it. Speed is so cool here.

**8-)** I renamed `statistics` to `recordActivity` cos due to smart naming policy.  

**9-)** Now, follow restriction and record activity data is also being dumped to local files in logs folder for the user to view them easily.
+ I used the structure based on the username to solve the possible misbehaviours during multi logs turned off. That's why the data will be dumped at the username level.  
+ Note that those JSONs are for viewing and has no any internal usage.

**10-)** Right, having considered that the users might have existing data in previous **followRestriction.json** file, I made a design choice. The thing is while it dumps new data to _follow restriction_ and/or _record activity_ local JSONs files, it will preserve the existing data in there 🤝🏼.
* E.g. you have 2 elements in **followRestriction.json** before merging this PR,
```javascript
{"old_user1":3, "old_user2":1}
```
* then you merge this PR **and** start/finish a good session, then that file will be like,
```javascript
{"old_user1":3, "old_user2":1, "YOURUSERNAME":{"new_user1":1, "new_user2":4, "new_user3":2}}
```
* _so that you can put that **old** data into correct username label or just remove it._
>If it gets hard to understand, start using it you will see how it works.

**11-)** Database tables' integrity will be checked at the session start and once the database file does not exist. So, the only worse situation can be, you drop the _old structured_ **instapy.db** file inside that location once the program is running, but I think of somebody doing it is crazy 😂  
The other times, it will just pass back the address of the database in use.

**12-)** For speed, I used simple data base write model- _a row at a time_ for writing follow restriction data.
It is fast for both querying and writing data.  

**13-)** Removed the default loaded instapy.db file from the repo cos of it is gonna written from scratch at database request. As a result, users will no more have to stash the **modified** instapy.db file while pulling new commits. Cos now, **instapy.db** file will be an **unstaged** file 🙂.  

**14-)** Matters I forgot to say! 😀 Please see the changes and ask for unclear lines in questions.

#### There are also other changes in several different parts came with this PR:
+ using **sorted** algorithm to remove duplicates faster at `get_links_for_username`
+ solve the case of empty strings while getting followers and following counts
+ fix follow button in `follow_given_user` definition to work properly in `Follow Back` buttons, too
+ some other slight changes for performance and good looks

#### **PROS**:
+ Previously dumping json files which held the very important data of follow restriction was problematic and once that JSON was getting huge, it was failing to dump in some session exit and your whole previous data was lost.  I don't even talk about corruption. 
But now, with this dedicated DB structure, it will always keep a robust data in there.
+ With this Data Base model, new tables or databases can be written **easily** and be served to required functions.

#### **CONS**:
In future there have to be written a new module to remove **so old** records in the database (_e.g. 6 months old data_), vice verse it is obviously gonna cost an inefficiency in querying time.

#### LAST NOTES
A casual user will not notice any of these changes cos they are intended mostly for  inner use.
Except, now there will be a noticeably new file- **recordActivity.json** holding record activity data in logs folder which will be a very useful file for everybody.
_I want to thank @converge once more for his amazing DB work in here which held the light on the road preparing this._

Writing this tool was easy but the phase of deciding the design was the most time consuming but hopefully I found all solutions for it at its best and wish it is gonna be useful to you! 

Cheers 😁


timgrossmann:
@uluQulu I'm absolutely impressed by this changes. Carefully reading your awesome documentation and the clean code really helped me understand the changes.
I love them! Can't express how valuable your work is to this project.

Thank you very much!

sionking:
Thank you for the hard work.
But, you make big changes, and many ppl have modified version of instapy, having those big changes while not been testing enough, but committed to trunk, make a lot of mismatches.

Maybe we should have a method of testing before merging ?

timgrossmann:
@sionking You are absolutely right about the testing part.
However, taking into account the versions of InstaPy that have been modified locally will be extremely hard. Maybe for those people it makes most sense to not pull changes right after the updates but slowly integrate the changes into their version.

I know that you have an insanely great setup with personal adjustments and modifications and I really appreciate that but I think most people use the "most" basic setup with the tool.

Do you have any proposal on how he could handle this?

CharlesCCC:
@uluQulu great job to put all those together. This will dramatically enhance the stability of instaPy in the future. I have had few ideas that I would like implement after this as well, I'm plan to implement it after I got back from a 2 months business trip. 

@sionking great point. 

My suggestion is to have two branches. 

`master` branch  which is the release version, that everybody gonna use.

`staging` branch which is the beta release, that all the power-user/pilot group will be use. 
this way, after everything confirmed with 'staging` branch, and then we can merge the PR to `master`

GangsterEmpire1:
@uluQulu Is it possible to get the data from the followRestriction.json to this new instapy.db database? I just noticed that followRestriction.json is not anymore used in this update and now InstaPy follows all the users who are in that followRestriction.json file.

CharlesCCC:
@GangsterEmpire1 I actually just noticed this in the past weekend while I was plan to do a upgrade. I think all the existing user from `followRestriction.json` should be migrated over to the DB `or` it should at least be looked at first before checking on the DB.

uluQulu:
@GangsterEmpire1 
I think you have not read the parts in PR description relating your situation.
But I like your point. Let me clear it up.

"_I just noticed that followRestriction.json is not anymore used in this update_"
- "Note that those JSONs are for viewing and has no any internal usage." (_from description_)

"_.. now InstaPy follows all the users who are in that followRestriction.json file_"
- Well, since that JSON has no any internal usage it does not matter what it has for that purpose.
- What matters now is, what the DB's `followRestriction` table has cos it is what follow times parameter rely on.

" _Is it possible to get the data from the followRestriction.json to this new instapy.db database?_"

sekil
That `10th` point of the description clearly explains it.
If you still could not understand THIS PART clearly, let me know.

**What more**?
Yes, the reason I made that design choice is right for the purposes of you- "somebody might need that old data..." and it is simply up to you now to do whatever you want with the old data.
_E.g. you can write a few SQL `INSERT INTO` lines in order to load that old data to currently used `followRestriction` table in the Data Base file_.
>If you really want this option and finds it hard to realize it, let me know to help you achieve it.


Why I did not make it an internal option to load old data in follow Restriction JSON file into new DB table? 
Well, it is not what I did not think about. The thing is,
+ people have a messy JSON and making that capability would opt out lots of problems for newbies
+ more importantly, here is the thing:
+ + Generally, people forget the `names` of profiles who followed them **after a few days**  and that few days have not much data (_optimally_) AND it is what matters. 

So, for the ultimate safety of the DB switch and also regarding the previous point, it is the way I realized personally and I comfortably decided to make a new BLANK table in a DB and not load old data.

Also, for that matter, I also said,
"In future there have to be written a new module to remove so old records in the database (e.g. 6 months old data), vice verse it is obviously gonna cost an inefficiency in querying time." (_from description_)
+ that point also means that old records like that should be reset, cos people will not remember you after that long time WHERE performance could shine ⛱

@CharlesCCC, I think it has also touched your points, too 💭

-------

Cheers 😁
--------

sionking:
> * "Note that those JSONs are for viewing and has no any internal usage." (_from description_)

repeating this line won't make it acceptable.
You should consider other users in your PR and please listen to what other contributes have to say, if 2 complain on the same thing, you did something wrong.

uluQulu:
@sionking 

So you suggest that I would completely cut off that JSON file after its internal usage has been moved to the analogical DB table.
And not serve users that easily viewable JSON file which is generated by that DB table?

Or you mean, we should not use DB table for storing/retrieving that data and use that buggy JSON structure?

PR #2587, opened 2018-07-31
able to return True

JohnTitor:
I found that because of `raise`, the process has not reached `return True, 'Verification failure'`.
So, I deleted `raise`.

If you prefer `raise`, I will rewrite it like that.

### DIFF
--- a/instapy/comment_util.py
+++ b/instapy/comment_util.py
@@ -89,7 +89,6 @@ def verify_commenting(browser, max, min, logger):
                     "PostPage[0].graphql.shortcode_media.comments_disabled")            
             except:
                 logger.info("Failed to check comments' status for verification...\n")
-                raise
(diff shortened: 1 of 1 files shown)


sionking:
I no expecrt in python but if we will get fail all the time, this is something that need to be fixed. withour raise we will never know (only checking the logs..)

JohnTitor:
OK.
I rewrote it.

uluQulu:
Hey guys
`raise` will crash the whole program, meanwhile, I recommend,
```python
except Exception as exc:
    logger.info("Failed to check comments' status for verification!\n\t{}".format(str(exc).encode("utf-8"))) 
    return True, "Verification failure"
```

JohnTitor:
I tried to rewrite it as @uluQulu said.
How about this change?

PR #2599, opened 2018-08-02
added Sleep

Mehran:
Fix for 'NoneType' object is not iterable


timgrossmann:
@Mehran we have an own sleep method which adds randomised wait times "around" the given, median value. 

Please use this instead: `from .time_util import sleep`

PR #2604, opened 2018-08-02
Pull from tim:master

meinert:


timgrossmann:
@meinert Make sure to change your password after this 😉

PR #2578, opened 2018-07-30
Fixes for the #2563

uluQulu:
Hey guys, I've included a few issues from people after #2563 and the solutions for them.  

**1**. I didn't know that `git` doesn't like empty folders where it removed the **db** folder after **instapy.db** pre-loaded file was removed.  
+ Added a smart layer to make the **path** in the given or default database location **if not exists**  

>  Other options are adding some small info file so that it is not empty to be removed by `git` OR add that file to `.gitkeep`, etc. But these options will work only for default **db** folder, so I chose a solution which will also work for custom locations off `Settings` class, too.

_Affected users **were** WHO cloned/downloaded the repo after #2563. But e.g. users who just pulled #2563 as an update could not face that issue, cos `git` keeps that **db** folder for them._ 

**2**. There is an issue with fetching the data from a DB and making a good dictionary from it while dumping **record activity** data. I have forgotten about `fetchall` vs `fetchone`, so it's my mistake and is fixed.  
>Actually the new code for it was the old code I used while testing, but before pushed #2563, I did that replacement (_for the sake of readability_) and thought it will work 100% 😂 which lead to this issue.

_I have really tested each bit of that PR but some unexpected issues pops up all the time and I'm still here to fix if any._

Cheers 😁

PR #2607, opened 2018-08-02
Fix for the most nasty bug off #2563

uluQulu:
Having my **InstaPy** not affected any bit by that, it took lots of time thinking about the heart of this issue,
```erlang

Heeh! Error occured while getting a DB profile for 'None':
	b'NOT NULL constraint failed: profiles.name'

```

And **solved** it 🤝🏼 with the invaluable collaboration and help of @Mehran 💪🏼 WHERE shared the VPS console to troubleshoot the issue and funnily in the first attempt found the cause of it!  
>It's the problem issued in a few **Issues** and discussed in #2578 thread.

**The thing is** before _python_ `3.6`, dictionary order was arbitrary. It means that the dictionaries don't have an order for those versions.  
------------------------------------------------------------------------------------------------------------
#### There **was a line of code** I wrote which relied on a dictionary order and it was what _misbehaving_!
```erlang
profile = {"id":None, "name":None}
```
**And** as we know that **the order of the elements** of `profile` dictionary will be random on versions before _python_ `3.6`, it will cause troubles, **cos** of  this:
```python
id, name = profile.values()
```
_At this place, **sometimes**, `profile` dictionary's keys' order become `{"name", "id"}` instead of expected `{"id", "name"}`._
And the reason it worked sometimes and did **not work other times** is that _randomness_ of the order!
---------------------------------------------------------------------------------------------------------------
Well, we can use `OrderedDict` from `collections` module for _python_ before `3.6` **BUT** let's not touch this dictionary orders again, it is a so nasty data type! 😝

Thanks again, @Mehran to help me find out this 🙋🏼‍♂️  
  

Cheers 😁

### DIFF
--- a/instapy/database_engine.py
+++ b/instapy/database_engine.py
@@ -11,7 +11,7 @@ def get_db(make=False):
     logger = Settings.logger
     credentials = Settings.profile
     # get existing profile credentials
-    id, name = credentials.values()
+    id, name = credentials["id"], credentials["name"]
     # make sure the address points to a database file
     if not address.endswith(".db"):
         slash = "\\" if "\\" in address else "/"


timgrossmann:
@uluQulu @Mehran So many really interesting edge cases! 
This makes total sense once seen 😄 

Thank you very much guys for the deep analysis and thought through work!

PR #2242, opened 2018-06-14
Skip private users, skip no profile pic, minimum media count when follow

marcomokastyle:
session.set_skip_users(skip_private=True, skip_no_profile_pic=True, min_media=25)
- Skip private profile
- Skip users with less posts than a specific number
- Skip user without a profile pic. (not fully tested yet)

Fix problem in clarifi_util.py, changed the part of the code to retrieve the image url, using graphql.shortcode_media.display_url.

PR #2609, opened 2018-08-03
Faster way to check if user is private

timgrossmann:
Updated how we check if the user is private because it took 25 seconds with the find_element_by_xpath function.
It takes less than a second with this edit.

Work by @mehdichekori @sionking

### DIFF
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -10,9 +10,10 @@ from selenium.common.exceptions import WebDriverException
 from selenium.common.exceptions import NoSuchElementException
 
 from .time_util import sleep
-from .util import update_activity
 from .util import add_user_to_blacklist
 from .util import click_element
+from .util import is_private_profile
+from .util import update_activity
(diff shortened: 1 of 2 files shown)


timgrossmann:
@sionking Could you double check, please?

Also, what was the idea of having the following parameter?
We do not know whether we are following the user at the point of checking for private, was the idea to add that check beforehand?

Thank you

sionking:
I am not really checking for following before is_private most of the cases.

But you **can't** check the xpath for user that is private but you do follow.
the graphql however will give you is_private True even if you already following the user (so you see his post and there is no real visual evidence for been private)!

You better not check the xpath and leave the following=true all the time since the graphql error that we experienced before not really reproducing.

timgrossmann:
@sionking Ok, good to know, thank you very much.
Then we'll merge it like this and can do further adjustments in #2242 if needed.

Thank you

Iiviin:
Really excited for this feature!

PR #2612, opened 2018-08-03
Feature - mandatory like

YannikFirre:
Hello,

I added the possibility to add a mandatory word list.

Example : 
```
session.set_mandatory_like(['#food', '#instafood'])
session.like_by_locations(['213099780'], amount=100)
```
Like photo's in location "213099780" only if contains the hashtags "#food" and "#instafood".

### DIFF
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -382,14 +382,14 @@ def get_links_for_username(browser,
 
 
 
-def check_link(browser, post_link, dont_like, mandatory_like, ignore_if_contains, logger):
+def check_link(browser, post_link, dont_like, mandatory_words, ignore_if_contains, logger):
     """
     Check the given link if it is appropriate
 
     :param browser: The selenium webdriver instance
     :param link:
     :param dont_like: hashtags of inappropriate phrases
-    :param mandatory_like: hashtags of appropriate phrases
+    :param mandatory_words: words of appropriate phrases
     :param ignore_if_contains:
 
     :param logger: the logger instance
@@ -476,9 +476,9 @@ def check_link(browser, post_link, dont_like, mandatory_like, ignore_if_contains
     logger.info('Link: {}'.format(post_link.encode('utf-8')))
     logger.info('Description: {}'.format(image_text.encode('utf-8')))
 
-    if mandatory_like :
-        if all((word in image_text for word in mandatory_like)) == False :
-            return True, user_name, is_video, 'Not mandatory likes', "Not mandatory likes"
+    if mandatory_words :
+        if not all((word in image_text for word in mandatory_words)) :
+            return True, user_name, is_video, 'Mandatory words not fulfilled', "Not mandatory likes"
 
     if any((word in image_text for word in ignore_if_contains)):
         return False, user_name, is_video, 'None', "Pass"
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -382,13 +382,14 @@ def get_links_for_username(browser,
 
 
 
-def check_link(browser, post_link, dont_like, ignore_if_contains, logger):
+def check_link(browser, post_link, dont_like, mandatory_like, ignore_if_contains, logger):
     """
     Check the given link if it is appropriate
 
     :param browser: The selenium webdriver instance
     :param link:
     :param dont_like: hashtags of inappropriate phrases
+    :param mandatory_like: hashtags of appropriate phrases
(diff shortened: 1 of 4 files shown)


timgrossmann:
@InfinityWebMe Great idea and solid implementation!

Could you rename the method to something more like `.set_mandatory_words` and add a part about it in the README file describing what the feature does, please?

Other than that, thank you very much for this awesome feature!

instapy/like_util.py
```diff
@@ -475,6 +476,9 @@ def check_link(browser, post_link, dont_like, ignore_if_contains, logger):
     logger.info('Link: {}'.format(post_link.encode('utf-8')))
     logger.info('Description: {}'.format(image_text.encode('utf-8')))
 
+    if mandatory_like :
+        if all((word in image_text for word in mandatory_like)) == False :
```
timgrossmann: Could you change this to `if not all((word in image_text for word in mandatory_like)):`

instapy/like_util.py
```diff
@@ -475,6 +476,9 @@ def check_link(browser, post_link, dont_like, ignore_if_contains, logger):
...
+            return True, user_name, is_video, 'Not mandatory likes', "Not mandatory likes"
```
timgrossmann: And the reason to something like `Mandatory words not fulfilled` ?

YannikFirre:
Yes i will changed that tomorrow ;-)

timgrossmann:
@InfinityWebMe Thank you very much!

In the best case we can merge this with #2313 and add a parameter which says whether one or all the tags have to be found in the text.

Do you think you could do that/would you be interested in doing that?

Thank you very much for the hard work.
Best
Tim

mz462:
This is awesome feature, thanks @InfinityWebMe for making that possible.

YannikFirre:
@timgrossmann 
> and add a parameter which says whether one or all the tags have to be found in the text.

It's would not be more solid if we create a other for one the tags have to be found in the text ? 

Thank you for your review :-D

timgrossmann:
@InfinityWebMe That's a very good question. 
I unfortunately can only speak out of a "programmers" view. I'd like to have less methods with more parameters to make it more customisable.

However, talking to many people who are not much into coding, it seems that they get confused quite easily with the possibilities.
So I might agree with you here so that the method name also describes the behaviour as close as possible.

What name would you suggest? Are you interested in implementing (or adjusting the other PR) to resemble your vision for those features? Would love to see them implemented.

So only one last question before merging this:
What is the reason you're not adding the `smart_hashtag`-part in the setup?
```python
        if use_smart_hashtags is True and self.smart_hashtags is not []:
             print('Using smart hashtags')
             tags = self.smart_hashtags
         else:
             self.required_tag = tags or []
             
         if not isinstance(tags, list):
             self.logger.warning('Unable to use your set_required_tag '
                                 'configuration!')
             self.aborting = True
```

YannikFirre:
@timgrossmann

> Are you interested in implementing (or adjusting the other PR) to resemble your vision for those features? Would love to see them implemented.

Yes i can check :)

> So only one last question before merging this:
What is the reason you're not adding the smart_hashtag-part in the setup?

Because the smart hashtags are not necessarily all required.
If I understand the smart hastags, the script gets a list of hashtags and like the photos as like_by_tags.
If I add it they would all be required. So not useful.

EDIT : I check the code again and I do not understand why this code " tags = self.smart_hashtags". After the variable "Tags" is referenced nowhere in the function and the use of smart hashtags must not prohibit the mandatory.

PS : Sorry for my poor english :P

timgrossmann:
@InfinityWebMe Thank you very much for the explanation.
I'm really sorry, you're absolutely right, this was my fault, it does not really make sense to use the smart hashtag thing here.

I will double check this PR one last time and then merge it.
Thank you very much for your hard work on this.

Best
Tim

mz462:
Hi guys, thanks for this commit.
Just a thought, is it possible to extend to mandatory_comment?

YannikFirre:
@mz462  Would you like a list of mandatory hashtags only for comments ?

Example :
```
session.set_mandatory_comment(['#food', '#instafood'])
session.like_by_locations(['213099780'], amount=100)
```
Like all photo's in location "213099780" and comment only if contains the hashtags "#food" and "#instafood".

mz462:
Correct, it would be perfect. 

Currently I don't think there is a way to limit comments to targeted posts within user_interact.

PR #2620, opened 2018-08-04
general bug fixes and improvements

sionking:
Add 
notifications_dialog (at the first IG page.)
is_private_profile (check)
is_page_available (skip if fail)
self.clarifai_proxy (good to use multiply accounts)
interact_likers (I like to interact)
self.debug
self.ignore_users = set() ([erf)

and:
self.parameters
which is a defaultdict that way we can fill it with parameters (min_followers, interact_private ....) and send it to any
function without the need to increase the function inputs.
I use it in validate_username that way I don't need to change this function over and over again.

### DIFF
--- a/instapy/clarifai_util.py
+++ b/instapy/clarifai_util.py
@@ -6,6 +6,8 @@ from clarifai.rest import ClarifaiApp, Image as ClImage
 def check_image(browser, clarifai_api_key, img_tags, img_tags_skip_if_contain, logger, full_match=False, picture_url=None):
     """Uses the link to the image to check for invalid content in the image"""
     clarifai_api = ClarifaiApp(api_key=clarifai_api_key)
+    if proxy is not None:
+        clarifai_api.api.session.proxies = {'https': proxy}
     # set req image to given one or get it from current page
     if picture_url is None:
         img_link = get_imagelink(browser)
--- a/instapy/clarifai_util.py
+++ b/instapy/clarifai_util.py
@@ -3,7 +3,7 @@ the image for invalid content"""
 from clarifai.rest import ClarifaiApp, Image as ClImage
 
 
-def check_image(browser, clarifai_api_key, img_tags, img_tags_skip_if_contain, logger, full_match=False, picture_url=None):
+def check_image(browser, clarifai_api_key, img_tags, img_tags_skip_if_contain, logger, full_match=False, picture_url=None, proxy=None):
     """Uses the link to the image to check for invalid content in the image"""
     clarifai_api = ClarifaiApp(api_key=clarifai_api_key)
     if proxy is not None:
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -8,7 +8,7 @@ import os
 from platform import python_version
 from datetime import datetime
 import random
-
+import collections
 import selenium
 from pyvirtualdisplay import Display
 from selenium import webdriver
@@ -146,7 +146,7 @@ class InstaPy:
 
         self.dont_like = ['sex', 'nsfw']
         self.ignore_if_contains = []
-        self.ignore_users = []
+        self.ignore_users = set()
 
         self.user_interact_amount = 0
         self.user_interact_media = None
@@ -178,11 +178,12 @@ class InstaPy:
         self.min_comments = 0
 
         self.relationship_data = {username:{"all_following":[], "all_followers":[]}}
+        self.parameters = collections.defaultdict(dict)
 
         self.bypass_suspicious_attempt = bypass_suspicious_attempt
 
         self.simulation = {"enabled": True, "percentage": 100}
-
+        self.debug = debug
         self.aborting = False
 
         # Assign logger
@@ -455,7 +456,7 @@ class InstaPy:
         if self.aborting:
             return self
 
-        self.ignore_users = users or []
+        self.ignore_users = set(users) or {}
 
         return self
 
@@ -568,6 +569,9 @@ class InstaPy:
     def follow_commenters(self, usernames, amount=10, daysold=365, max_pic=50, sleep_delay=600, interact=False):
         """ Follows users' commenters """
 
+        if self.aborting:
+            return self
+
         message = "Starting to follow commenters.."
         highlight_print(self.username, message, "feature", "info", self.logger)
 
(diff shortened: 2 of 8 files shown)


instapy/instapy.py
```diff
@@ -158,6 +159,9 @@ def __init__(self,
         self.clarifai_img_tags = []
         self.clarifai_img_tags_skip = []
         self.clarifai_full_match = False
+        if clarifai_proxy is not None:
+            clarifai_proxy = 'https://' + clarifai_proxy
```
timgrossmann: Doesn't this lead to problems if users provide proxy with https in front?

instapy/instapy.py
```diff
@@ -8,7 +8,7 @@
 from platform import python_version
 from datetime import datetime
 import random
-
+import collections
```
timgrossmann: Do we use any other data structure than the set from this?
maybe doing a `from collections import set`would be better here

instapy/clarifai_util.py
```diff
@@ -3,9 +3,11 @@
 from clarifai.rest import ClarifaiApp, Image as ClImage
 
 
-def check_image(browser, clarifai_api_key, img_tags, img_tags_skip_if_contain, logger, full_match=False, picture_url=None):
+def check_image(browser, clarifai_api_key, img_tags, img_tags_skip_if_contain, logger, full_match=False, picture_url=None, proxy=None):
```
timgrossmann, line 6: Do we have documentation in the README for this parameter? 🤔

instapy/instapy.py
```diff
@@ -663,9 +673,47 @@ def follow_likers (self, usernames, photos_grab_amount=3, follow_likers_per_phot
...
+            return self
+
+        message = "Starting to interact likers.."
+        highlight_print(self.username, message, "feature", "info", self.logger)
+
+        if not isinstance(usernames, list):
+            usernames = [usernames]
+
+        if photos_grab_amount>12:
+            self.logger.info("Sorry, you can only grab likers from first 12 photos for given username now.\n")
```
timgrossmann, line 704: Can we add a TODO comment here which describes that we have to add the scrolling part here?

instapy/like_util.py
```diff
@@ -72,7 +72,7 @@ def get_links_for_location(browser,
         # Make it an array to use it in the following part
         media = [media]
 
-    browser.get('https://www.instagram.com/explore/locations/{}'.format(location))
+    web_adress_navigator(browser, 'https://www.instagram.com/explore/locations/' + location)
```
timgrossmann, line 78: What's the reason you're not using `.format()`here?

instapy/util.py
```diff
@@ -644,6 +638,36 @@ def highlight_print(username=None, message=None, priority=None, level=None, logg
...
+def is_private_profile(browser, logger, following=True):
+    is_private = browser.execute_script(
+        "return window._sharedData.entry_data."
+        "ProfilePage[0].graphql.user.is_private")
+    # double check with xpath that should work only when we not follwoing a user
+    if is_private is True and not following:
+        body_elem = browser.find_element_by_tag_name('body')
+        is_private = body_elem.find_element_by_xpath(
+            '//h2[@class="_kcrwx"]')
+        if is_private is not True:
```
timgrossmann, line 680: Shouldn't the indentation here he 1 less?
Considering that this part will never be reached if is_private is false

timgrossmann (commented):
Awesome additions, fixes and changes.

Love your work, could you go over the comments and clarify some parts for me, please?

Thank you very much @sionking

instapy/util.py
sionking: I really don't want to get into the double check... I thinkit can be removed.
But still didnot get your point...

instapy/clarifai_util.py
sionking: I will try to find a way to do it I am not good at doc

instapy/instapy.py
sionking: Will do

instapy/instapy.py
sionking: Which scrolling part?

instapy/like_util.py
sionking: Will change

instapy/instapy.py
timgrossmann: Here you log, if the grab_amount is larger than 12, that right now we are only able to grab the first 12. 
However in other methods we have a "scrolling" part which scrolls and loads more elements when the number is larger than 12

instapy/util.py
timgrossmann: Sorry for the missed clarity in the comment...

I mean if you check `if is_private is True and not following:`, then `is_private` has to be `True` in order to enter the block below. 
So the check for `if is_private is not True:` won't ever be `True` because `is_private` given the above condition was `True` has to be `False`.
So the block below the second if will never be reached.

sionking: let me check the logic again...

instapy/instapy.py
sionking: They should know the syntax.
What is your suggestions ?

timgrossmann: That's true, if we have good documentation than it should work.
Maybe doing something like

```python
if not clarifai_proxy.startswith('https'):
    clarifai_proxy = 'https://{}'.format(clarifai_proxy)
```

What do you think?

sionking: done

instapy/instapy.py
sionking: follow_likers don't have any scrolling part ...

instapy/util.py
sionking: I sse is_private already implemented

instapy/like_util.py
sionking: I don't think every string need to use format, if you just concatenate 2  strings it is better and more clear to use +

sionking:
@timgrossmann  please check and merge

timgrossmann:
@sionking Looks good to me now, part for the `proxy` in the README is still missing though, do you just provide the proxy URL there?

@uluQulu Could you maybe double check this PR so we can merge it, please?

sionking:
So I only took is_page_available and closing

Issue #2611, opened 2018-08-03
Pow! Error no such tables


Totallytorsten:
## Current Behavior

Program opens for about a second and then closes. Error is as follows:

> Pow! Error occurred while dumping follow restriction data to a local JSON: b'no such table: followrestriction'
> Pow! Error occurred while dumping follow restriction data to a local JSON: b'no such table: recordActivity'

The program then crashes because it cant find recordActivity. Downloaded a totally fresh version and getting the same error. Think it has to do with the changes to database_engine from earlier this morning.

uluQulu:
@Totallytorsten,
Are your files up-to-date?   (_see `git status`_)

If you got the latest files and the problem still persists, let me know. I will help.

-------------

Cheers 😁

eddyizm:
Getting the same error. git status shows everything already up to date.

uluQulu:
I have checked, @eddyizm.
You are right, @Totallytorsten, it is a bug born off #2605 🐝

Wait a few minutes, I will share a solution to it and explain the problem details 🤝🏼

uluQulu:
Hey @eddyizm, @Totallytorsten
The fix is ready - #2627 ☄

eddyizm:
Thanks @uluQulu . I actually ended up renaming the db and let it run again, thus generating a new db and everything was well. Should I go back to the old db and let the app handle the upgrade or can I leave it alone, not sure how much I'm giving up with all the old stats?

uluQulu:
Smart of you @eddyizm 🧐

Well, it does not matter, cos as of now, the data stored in the `statistics` (_now replaced by `recordActivity` table_) table is not in use by any module, it is just a useful data to see but in future it will be crucial and will be used by internal modules.
Anyways, even if you let the program handle the upgrade, it will make a new table `recordActivity` for that reason, and the `statistics` table will not be used.  

The reason I designed it so is I thought there are some people who uses the data inside `statistics` table and I kept it for them to take backups.  

------
Cheers 😁

callahanryang:
worked for me. Renamed old DB to instapy-1.db and the program created a new DB and no more issues logging in! Thanks! :-) +1

PR #2605, opened 2018-08-02
Fix Problem on Create Database

rafaelfcsouza:
Small refactor on Create Database and Get Database Logic on instapy/database_engine.py
Separated logic into small functions for better redability and removed comments (functions names have full name now)
Separated SQL into constants

Fixes #2588

PR #2627, opened 2018-08-05
Solve misbehaviours of #2605 🐝

uluQulu:
Popped up from #2611 I saw a bug and started to write a fix for it then saw a big gap in the logic from I designed to end result which missed lots of considerations regarding the code strength.  

Fine, I HAVE solved it once more but now I'm afraid that I wrote this dedicated DB engine (_maybe it was so dedicated_) 😝

#2611:
+ In case of there is already an **instapy.db** file in the old structure (_just `statistics` table_) it will neither make a new database nor make required new tables.
>It was just making the given tables only once there existed no any database file in the given location.

This is what I talked about smooth design, @timgrossmann 🙂

_If you have any questions about other changes (_non-visual_), please let me know!_

-----------

Cheers 😁

### DIFF
--- a/instapy/database_engine.py
+++ b/instapy/database_engine.py
@@ -3,11 +3,17 @@ import sqlite3
 
 from .settings import Settings
 
+
+
+
 SELECT_FROM_PROFILE_WHERE_NAME = "SELECT * FROM profiles WHERE name = :name"
+
 INSERT_INTO_PROFILE = "INSERT INTO profiles (name) VALUES (?)"
+
 SQL_CREATE_PROFILE_TABLE = """CREATE TABLE IF NOT EXISTS `profiles` (
                   `id` INTEGER PRIMARY KEY AUTOINCREMENT,
                   `name` TEXT NOT NULL);"""
+
 SQL_CREATE_RECORD_ACTIVITY_TABLE = """CREATE TABLE IF NOT EXISTS `recordActivity` (
                   `profile_id` INTEGER REFERENCES `profiles` (id),
                   `likes` SMALLINT UNSIGNED NOT NULL,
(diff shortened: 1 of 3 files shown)


ghoeffner:
Manually applied this patch and it worked.

timgrossmann:
@uluQulu Really sorry this is causing issues...
However I don't really understand why you need to pass the `make=True` initially and then never again 🤔 
Separating it into two different methods, `initialise` which basically is `make=True` and `get_db` which basically is `make=False`.
Could you elaborate on that, please? Looks like I'm just too stupid 😅 

Thank you very much

uluQulu:
@timgrossmann 
It is partly historical cos earlier while writing this module it was checking table integrity in every query. But to save some speed I made it check it only at the program start, so it is gonna make db and required tables only in the start (_and anytime database file does not exist_).  

Also in future there could be some calls to the database engine where it might require to force-make tables inside the program (_else than start, also when database file already exists_).  
I also tried to call it, `make=["auto", "forced"]`.  

Apart from that, what is `initialize_database` ?  
It sounds like you are firing up the engine and it will run until the end of the program 😁😁😁

What it does is just checking the viability and integrity of the database and its tables - it is not initializing anything.

Again, "`make=True`" argument will **force check** the situation and make a database or table if needed, whilst "`make=False`" argument will make a database and its tables **only once** the given path does not include a database file (_**instapy.db** by default_).
>Also, regardless of `make` value, checking the `ID` existence is gonna occur in every query.

The result is ultimate precision in maximum speed.  

_I have really designed that piece in tip-top shape, what a pity that it is not understood well_. 😐

--------

Cheers 😁

timgrossmann:
@uluQulu I'm sure it's more likely the thing that I'm somehow not experienced enough to understand it rather than you not designing it in tip-top shape.

In most, newer, projects I try to keep it very short in terms of functions and rather make more of them than add many parameters. 
Especially situations in which In only pass true once to the method and want to have a false all the other times seems like a method that, no matter the name, should be separated into different methods.

I absolutely understand your explanation of it and the reasoning. I think it's a reasonable design for now. 
It's simply more understandable, maybe just for me, if the methods are shorter and it has more references to other methods.
There even is the idea of "flow design" which only defines two basic components which are "execution "-units and "caller"-units. (http://geekswithblogs.net/theArchitectsNapkin/category/19719.aspx)
This is pretty interesting even though.

However with this amount of great devs, we most likely won't find a solution which is perfect for everyone.

Thank you very much @uluQulu 
Again, your work and the quick fixes for bugs are really essential to this project. The amount of features, code and fixes you contributed to this project until now is impressive.
Props to you

instapy/database_engine.py
```diff
@@ -72,14 +71,14 @@ def create_database(address, logger, name):
 
 
 
-def create_tables(cursor, table):
```
uluQulu: I also thought about it,
It is a good change ☄

uluQulu:
Many thanks, @timgrossmann.
"_flow design_" looks like a neat article, I really need to read some professional books to catch you.
As @converge said, "_it has always been fun_" with this project 🤩  

But I will have to say this.
Once I decide to change some lines in a source code (_written by others or me_) there is one thing that has a high priority for me:
+ Have I understood what this code is doing **correctly**?
+ + If I make a change will it **consider** the consequences?
+ + + **Is it** gonna lose some piece of useful (_not even crucial_) idea with my edits?

It seems that the external changes to database engine missed the original spirit and all the considerations I've outlined above was not taken- either it was not noticed, or was intentional.  
But the fact is code strength, smoothness, situation handling were all lost.

--------

Cheers 😁

PR #2633, opened 2018-08-05
Second Database Refactor

rafaelfcsouza:
Introduced ORM with SQLAlchemy
Sorry if the lib is not suited, but I found the docs really good, but we can change it easy.

Added some tests on database initizalization and setup (really poor).

ToDo: Migrate FollowRestriction and RecordActivity to use the same structure.


instapy/database/dao/__init__.py
```diff
@@ -0,0 +1,9 @@
+from sqlalchemy.orm import Session
+
+
+class DAO:
```
rafaelfcsouza: Is this the best place to put this class? I'm not sure. but a file only for it seems strange...

timgrossmann:
@rafaelfcsouza Oh wow, this is awesome!
I really have to read through this closely to get and understand what you're doing here.
How are you able to omit the whole db insertion part? 
This would be an incredible addition to InstaPy, thank you very much for the hard work.

Btw. do you think you're able to implement the things you put in ToDo? 
Or did you just put them there to make sure people notice?

Best
Tim

rafaelfcsouza:
Sure I can implement them, dont know when tho!
You can take a look at sqlalchemy.org is not some crazy magic hahaha, its just database abstraction.
pretty useful on not writing inserts all the time.

timgrossmann:
@rafaelfcsouza Thank you very much for the work up to this point.
Let's see if we should merge this soon or if we keep it and wait whether you find the time to add the remaining todos to the PR.

Thank you

CharlesCCC:
great work, 👍  
the change looks good to me, but there is conflict need be resolved @rafaelfcsouza  
I would merge this first and then when it is stable enough, we can then convert the TODO over.
I think this will be a big helper for future DB related  improvement/change.

joaoleite:
I was very interested in that! ORM is life!

Even what I would very much like is to have more tables with more metrics to work on top of that.

Is someone coding this PR?

timgrossmann:
@joaoleite Unfortunately the PR is very old.
It's still open to have it as  "reference" since at one point we definitely want to have that integrated into InstaPy 💪

joaoleite:
> It's still open to have it as "reference" since at one point we definitely want to have that integrated into InstaPy 💪

I will try next week to download the branch and try to rebase.

It will help us a lot to have an ORM on InstaPy!

😍

timgrossmann:
@joaoleite Would be awesome 💪 🙌 
If you need some insights into the DB, we have a small mysql workbench created by @converge here:
https://github.com/InstaPy/instapy-docs/tree/master/database

timgrossmann:
@joaoleite Considering the sophisticated changes we are making to the DB structure due to the introduction of the GUI, we'll close this PR for now since it is very outdated.

We would still love to have an ORM in place for the new setup.

PR #2651, opened 2018-08-08
Pick unfollowers error

meinert:
An exception was thrown each time I ran this.

    all_unfollowers, active_unfollowers = session.pick_unfollowers(username="xxxxxx", compare_by="earliest", compare_track="first", live_match=True, store_locally=True, print_out=True)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -107,3 +107,9 @@ logs/
 # pytest
 .pytest_cache/
 tests/logs
+unfollow.py
+meinerttt.py
+master.py
+follow.py
+
+myLogs/
--- a/.gitignore
+++ b/.gitignore
@@ -107,9 +107,3 @@ logs/
 # pytest
 .pytest_cache/
 tests/logs
-unfollow.py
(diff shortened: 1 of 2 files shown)


timgrossmann:
@meinert Thank you very much for this fix.

Let's wait for someone else to approve this and then merge it.
Thanks

richardwardza (approved):
This will at least give insight into why it may have failed.

uluQulu:
Hi guys
@meinert, 
All of those three object initializations are essential for the rest of the definition.
What I mean is if there is a problem and it is just skipped (_by `try`/`except` in your example_) is a bigger problem.
If there is a problem it must be solved.

I remember `load_followers_data`- _smart file chooser_ function was the most time consuming definition out of the _relationship data_ work and that part you have outlined is just doing some simple _dict_ stuff in list comprehension.

Can you share the error output you get out of that block?
I am sure it has a simple solution.

-------
Cheers 😁

PR #2659, opened 2018-08-10
In unfollow utils, removing the person from file if we cant unfollow him

HaimBendanan:
I had a situation where the bot was stuck on a person that can't be unfollowed, solved it by removing this person from the local followed file.

### DIFF
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -89,7 +89,8 @@ def get_following_status(browser, person, logger):
             '--> Unfollow error with {},'
             ' maybe no longer exists...'
                 .format(person.encode('utf-8')))
-
+        delete_line_from_file('{0}{1}_followedPool.csv'.format(logfolder, username), person +
+                                              ",\n", logger)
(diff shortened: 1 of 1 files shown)


CharlesCCC:
Good thoughts, the situation like this usually is caused by the followed by user changed their username. 
but one thing I would like to consider is, what if it is false/positive. this could potentially remove any user that should not be removed from the list (e.g. if the internet is acting up etc.)  @sionking I believe you had worked on something like this in the past.

sionking:
I will [post my unfollow file soon

sionking:
but removing every not succeed unfollow is not good !
Since internet connection you will remove all of your list.
Ther should be a verification.

timgrossmann:
@HaimBendanan @sionking Let's wait for @sionking s solution.
Also looking at the diff, it seems like there is no update in the file.

HaimBendanan:
I updated my branch and removed this change, forgot to close the PR.
Thanks for closing :)

PR #2660, opened 2018-08-10
Unfollow list returned

meinert:
Made the unfollow method return the list of users attempting to unfollow. Usefull to conflude if all possible users have been unfollowed or the job needs to be restarted. E.g. by comparing len(unfollow_list) with unfollowNumber

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -107,3 +107,8 @@ logs/
 # pytest
 .pytest_cache/
 tests/logs
+unfollow.py
+meinerttt.py
+master.py
+follow.py
+myLogs/ 
--- a/.gitignore
+++ b/.gitignore
@@ -112,3 +112,6 @@ meinerttt.py
 master.py
 follow.py
 myLogs/ 
+all_unfollowers.py
+nonfollower_clean.py
+test.py
--- a/.gitignore
+++ b/.gitignore
@@ -115,3 +115,4 @@ myLogs/
 all_unfollowers.py
 nonfollower_clean.py
 test.py
+
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -129,6 +129,7 @@ class InstaPy:
         self.commented = 0
         self.followed_by = 0
         self.unfollowNumber = 0
+        self.unfollow_list = []
         self.not_valid_users = 0
 
         self.follow_times = 1
@@ -2266,7 +2267,7 @@ class InstaPy:
                                                                  self.logfolder)
 
         try:
-            unfollowNumber = unfollow(self.browser,
+            unfollowNumber, unfollow_list = unfollow(self.browser,
                                       self.username,
                                       amount,
                                       customList,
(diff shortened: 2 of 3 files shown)


uluQulu:
hm @meinert 
I use this in my **quickstart** script
```python
spoilers = 0
zx = 0
# other initializations

while True and spoilers < 10 :
    try:
        if zx >= 1:   # I start unfollower in after some activity

            session.unfollowNumber = 0

            while session.unfollowNumber < 66 :
                session.unfollow_users(amount=random.randint(52, 66), nonFollowers=True, style="RANDOM")
                sleep(random.randint(100, 200))

            sleep(random.randint(400, 600))

        zx += 1

        """
        OTHER FEATURES HERE
        """

    except Exception as exc:
        spoilers += 1
        print("\n\n\tSpoilers : {}\n\t{}\n\n".format(spoilers, str(exc).encode("utf-8")))
```
Shortly, I think, it is kind of a personal need and should be prepared manually for the taste.

---------
Cheers 😁

meinert:
Thanks for your comment. With my added feature it is possible to check if all possible users to unfollow has been unfollowed or something went wrong and I need to restart the script.

Your solution - similar - but as I see it you keep invoking the unfollow_users method until your predefined number (66) is reached. But what if the the number of users avalible to unfollow is less than 66....your scipt will never end. I uses InstapyFollowed=(True, "nonfollowers") and unfollow_after options which often only has few users to unfollow.
Furthermore what if the number of users avalible to unfollow is larger than 66......your script will end before all avalible users have been unfollowed.

Still hope that this pull requst will be accepted

uluQulu:
You're welcome, @meinert
I think,
"_if all possible users to unfollow has been unfollowed.._"
- You want to unfollow everybody in some context? You should know that unfollowing more than some amount has a bad effect regarding the momentary, hourly and daily limits.
So, giving a good value and having it accomplished is the best experience and I used it in my while loop.

" _or something went wrong and I need to restart the script_"
- What can go wrong? Please explain.
E.g. if it breaks the unfollow loop with the message:
- `'username' might be blocked`, you should not do continue the loop, instead if you really wanna unfollow again, take a time break and try again to see if it works. (_still can be handled with the while loop in my example_).
- Other things can go wrong is internal error, such as XPATH change which is a clear situation.

That example is really my personal lines that's why I am sure that there are more than 66 non followers 😂
If you are unsure about it, you could do some other personal check, like simplest choice- adding `attempts` counter, etc or from core catching the return situation.
Again saying those are custom needs.

You have a nice idea, @meinert, I really like it cos I use sth similar to it.
And I think if you clear up,
+ What you do and can offer with your add ups?
+ Is it really needed AND is it targeting the GENERAL (_or MASSIVE_) audience?
+ Is it suited well? (_e.g. maybe it could be much better as a parameter?_)

If you got a time, please invest in clearing those points, as you know, you might have a brilliant addup but it must be expressed brilliantly, too.

That's all from me. 🙋🏼‍♂️
>BTW I would call the work you are doing- "UNFOLLOW the **Real Deal**" 👊🏼

--------
Cheers 😁

sionking:
The only option to verify unfollowing is to sync the following list for IG servers and compare.
I do it once every 2 days for each account.

PR #2661, opened 2018-08-11
fix the interact option in like_by_tags

raphaelcharriez:
Liking several pictures on a profile page gives better results, it feels less bot like

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1253,24 +1253,22 @@ class InstaPy:
                         if liked:
 
                             if interact:
-                                username = (self.browser.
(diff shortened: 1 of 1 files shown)


CharlesCCC:
hi @raphaelcharriez good opinion and it make sense, but I would rather leave the option to the user instead of force everybody to do it.

timgrossmann:
@raphaelcharriez @CharlesCCC I see it the same way as Charles. The `.set_user_interact` method is there to use. We should not just set it with some random numbers...

Still, thank you very much for the proposal.
Looking forward to further PRs from you

PR #2690, opened 2018-08-16
update "how to raspberry" docs

gherkins:
added instructions for how to install with raspbian stretch lite 4


timgrossmann:
@gherkins Awesome one!
Thank you very much, documentation is always key!

PR #2701, opened 2018-08-17
count words in bio and also check for links

ghost:
Check for links and count the words in bio


ghost:
i don't know what i am doing ...

Issue #2538, opened 2018-07-25
CRITICAL[...] Wrong login data!


Jenuan:
First of all, I was getting some message like "issue with cookie for user." Then, I searched here and found out that I had to set " self.switch_language = False in instapy.py "
I did it and now I'm getting "Cookie file not found, creating cookie..." and finally "Wrong login data!"
I tried to delete my cookie file but it didn't solve the problem.
Also, I checked my username and password and I could log in manually on instagram using chrome. 
Does anyone know how to solve it ? thx

Jenuan:
Just to give an update if some1 gets this error.
I fixed it by updating my instapy to the newest repository.

fgisslen:
I get this error to

Jenuan:
Well, i updated my instapy but when i tried running today it is still not working.
i get this error
...
Cookie file not found, creating cookie...
CRITICAL [2018-08-07 19:12:56] [insta_username]  Wrong login data!
...

Then, I watched my browser and saw that instapy is trying to login in the first page of instagram wich in fact is to open a new instagram account.
@fgisslen can u watch it and see if that is what happening with u?

@uluQulu have u seen this before? any clue about what I'm supposed to do ?

generalu:
This is a temporary fix, but this issue was bypassed by changing: 
login_util.py line 116 from "https://www.instagram.com"

to

"https://www.instagram.com/accounts/login"

fgisslen:
i'm using headless, then I can't check that, right? @Jenuan

Jenuan:
well, you can always turn off headless mode if you want i think you should just put headless=False @fgisslen 

@generalu i did change it but i got this error:

Message: no such element: Unable to locate element: {"method":"xpath","selector":"//article/div/div/p/a[text()='Log in']"}

priscila225:
I had the same problem!!! I finally fixed here when I changed the line 174 from login_util.py code to be if len(nav) >= 0 instead len(nav) == 2 --> note: I didn't investigate more this line but I bet this is totally diff than 2 and turns into a false return.

Jenuan:
@priscila225 could you post or send me how it is written your login_util.py ? because I did change like you said but it seems instapy can't find the place to log in my username and password.

fgisslen:
@Jenuan No I can't turn off headless because it runs on ubuntu server.

@priscila225 Do you know if this is a common problem?

fgisslen:
[...]

PR #2683, opened 2018-08-15
Fix: convert exception messages into a string before encoding AND some other small fixes

uluQulu:
A trivial mistake of mine (_mostly_) 🤤
Thanks to @Edhim to raise it at #2663 in a great feedback.

#### Now solved.

--------

Cheers 😁
----------

PR #2703, opened 2018-08-17
Made changes to login_util to prevent "Wrong login data! (#2538)" error for headless

dnishiyama:
Identified root cause of issue "Wrong login data! (#2538)" to be wrong Log In page url and update to the html tag for Log In page form submission


instapy/login_util.py
```diff
@@ -113,7 +113,8 @@ def login_user(browser,
     # include time.sleep(1) to prevent getting stuck on google.com
     time.sleep(1)
     
-    browser.get('https://www.instagram.com')
+    #browser.get('https://www.instagram.com')
+    browser.get('https://www.instagram.com/accounts/login'); #Go directly to login page per recommendation from @generalu
```
uluQulu: Hm, sounds smart idea but starting with that page directly is a bit noticeable and non-humanly.

instapy/login_util.py
```diff
@@ -137,7 +138,8 @@ def login_user(browser,
 
     # Check if the first div is 'Create an Account' or 'Log In'
     login_elem = browser.find_element_by_xpath(
-        "//article/div/div/p/a[text()='Log in']")
+        #"//article/div/div/p/a[text()='Log in']")
+        "//button[text()='Log in']") #Instagram changed the html layout of their login page
     if login_elem is not None:
```
uluQulu: The layout has not changed, you have misunderstood it.
`"//article/div/div/p/a[text()='Log in']")` is the location of **Log in** button at `.com/` address
`"//button[text()='Log in']")` is the location of **Log in** button at `.com/accounts/login` address

And since you navigate to `.com/accounts/login` rather than `.com/` **above**,
it surely does not find,
`"//article/div/div/p/a[text()='Log in']")` element at the loaded page
and next,
`if login_elem is not None:` returns False.

#### But all it needed was a little bit (_~1 more seconds_) to load the `username` element properly... 🤤

dnishiyama: Thanks uluQulu—I understand what you're saying. I'll try it out with time.sleep(2) like in your PR

uluQulu:
I just prepared this to make it more clearer,
ok
It is the state of current **login_util.py**.

I believe, 2nd commit- 9cd18c3a5f9fed6021dba8d2b089d0261b698e27 in #2683 definitely solved your issues, @dnishiyama.

-------

Cheers 😁
----------

instapy/login_util.py
uluQulu: The reason it worked for you with those changes is, since it directly navigates to `.com/accounts/login` address, it gets that **extra time** while searching for at least 2 XPATHs at that page 🤓 after navigation, so the `username` element could load properly during that time.
> There were an XPATH that took 25 seconds (**always**) to find I remember of 😋

PR #2711, opened 2018-08-18
fix proxy issue

Mehran:
Fix Proxy issue


sionking (changes requested):
if self.proxy_address and len(self.proxy_port) > 0:
should be 
 if self.proxy_address and int(self.proxy_port) > 0:

timgrossmann:
@sionking Could you explain why you're checking the port number to be greater than 0?

timgrossmann:
I think #2739 is the better approach here.

PR #2725, opened 2018-08-21
update for an error with the scrolling in the dialog

ghost:
in a recent changes of IG , the scroll doesn't work anymore because the dynamic fetch in the dialog is broken for some reason...
you can use this little modification as a workaround.

### DIFF
--- a/instapy/util.py
+++ b/instapy/util.py
@@ -420,6 +420,9 @@ def scroll_bottom(browser, element, range_int):
     # put a limit to the scrolling
     if range_int > 50:
         range_int = 50
+        
+    size = browser.get_window_size()
+    browser.set_window_size(size['width'],size['height'] - 1)
 
     for i in range(int(range_int / 2)):
         browser.execute_script(
@@ -427,6 +430,8 @@ def scroll_bottom(browser, element, range_int):
         # update server calls
         update_activity()
         sleep(1)
+        
+    browser.set_window_size(size['width'], size['height'] + 1)
 
     return
 


ghost:
this modification can be used even if IG fixes the bug

uluQulu:
Nice job, @ilmetu.
Can you make it inside `get_users_through_dialog` definition so that it does not apply to all of the scroll activities using `scroll_bottom`
Also get the size of window for once rather than all time cos it is a compute intensive activity which is also error prone.
>e.g. after getting size for once, `height = height+1 if sc_rolled % 2 == 0 else height-1`


@sionking, I think of scraping is good but let's keep it as a last resort and I'm sure that this problem will be solved and also this must be solved otherwise relying on just **graphql** data will decrease the flexibility.

### UPDATE:
@tayax
It seems to hit the loading **Likers** list, too.
So it is best to keep it in `scroll_bottom` definition and add a new parameter to trigger window resizing at request.


-------

Cheers 😁
----------

ghost:
Also, the problem remains from `likers_from_photo` in the file `commenters_util.py`

gitpatrickhub:
I've adding this code, but still encountering the error on Raspberry Pi 3.

sionking:
now scroll is back

timgrossmann:
@uluQulu @ilmetu @sionking Can we close this PR?

uluQulu:
Yes @timgrossmann.
It was a bug cos of dialog style being changed and now it is solved.
>@sionking said somewhere, "_don't worry, `followers` & `following` data is still open to public, it will be fixed_"

But the idea will be kept for the future in a situation like this.

----------

Cheers 😁
-------------

gitpatrickhub:
Still not working after latest pull on Raspberry Pi 3. An error doesn’t get output to the console, but skips the interaction with users. When I watch the browser it doesn’t continue to scroll, although scrolling is possible.

PR #2727, opened 2018-08-21
Log the userid when following other users

CharlesCCC:
I have thought of implement this for awhile now,  Especially after got more than 100+ user that stuck in my following list that not able to unfollow (most likely they have changed their username)

Finally I was able to implement it after I got back from a long business trip. 

here are the changes:

Inside the `USERNAME_followedPool.csv` instead of only logging the username as it was (today) e.g.
```
2018-08-21 16:04 ~ javaabcdefg
```

It now become something like this 
```
2018-08-21 16:04 ~ javaabcdefg ~ 1234567890
```

**During unfollow procedure, if the instapy not able to load the following user profile by the `name` it will try use the `userid` that was logged to get to the profile page.** Which means, whatever the username change into, we should always able to find the user and to unfollow it. (unless the user is been deleted/bloced/banned, i'm not sure how does Instagram gonna treat on that situation, assume it will be automatially removed from your following list) 

please point out if anything I might have missed and happy contributing.

### DIFF
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -316,8 +316,7 @@ def unfollow(browser,
                                 ' now unfollowing: {}'
                                 .format(str(unfollowNum), amount, person.encode('utf-8')))
 
-                            delete_line_from_file('{0}{1}_followedPool.csv'.format(logfolder, username), person +
-                                              ",\n", logger)
+                            delete_line_from_file('{0}{1}_followedPool.csv'.format(logfolder, username), person, logger)
 
                             print('')
                             sleep(15)
@@ -349,7 +348,7 @@ def unfollow(browser,
                             .format(str(unfollowNum), person.encode('utf-8')))
 
                         delete_line_from_file('{0}{1}_followedPool.csv'.format(logfolder, username),
-                                              person + ",\n", logger)
+                                              person, logger)
 
                         print('')
                         sleep(2)
@@ -358,7 +357,7 @@ def unfollow(browser,
                     # if he is a white list user (set at init and not during run time)
                     if person in white_list:
                         delete_line_from_file('{0}{1}_followedPool.csv'.format(logfolder, username),
-                                              person + ",\n", logger)
+                                              person, logger)
                         list_type = 'whitelist'
                     else:
                         list_type = 'dont_include'
@@ -474,8 +473,7 @@ def unfollow(browser,
                         '--> Ongoing Unfollow {}/{}, now unfollowing: {}'
                         .format(str(unfollowNum), amount, person.encode('utf-8')))
 
-                    delete_line_from_file('{0}{1}_followedPool.csv'.format(logfolder, username), person +
-                                      ",\n", logger)
+                    delete_line_from_file('{0}{1}_followedPool.csv'.format(logfolder, username), person, logger)
 
                     print('')
                     sleep(15)
@@ -557,8 +555,7 @@ def unfollow_user(browser, username, person, relationship_data, logger, logfolde
         click_element(browser, unfollow_button) # unfollow_button.send_keys("\n")
         logger.warning("--> Unfollowed '{}' due to Inappropriate Content".format(person))
 
-        delete_line_from_file('{0}{1}_followedPool.csv'.format(logfolder, username), person +
-                          ",\n", logger)
+        delete_line_from_file('{0}{1}_followedPool.csv'.format(logfolder, username), person, logger)
 
         if person in relationship_data[username]["all_following"]:
             relationship_data[username]["all_following"].remove(person)
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -91,7 +91,7 @@ def get_following_status(browser, person, logger):
     except:
         logger.error(
             '--> Unfollow error with {},'
-            ' maybe no longer exists...'
+            ' maybe no longer exists....'
                 .format(person.encode('utf-8')))
 
     return following, follow_button
@@ -236,6 +236,7 @@ def unfollow(browser,
         try:
             sleep_counter = 0
             sleep_after = random.randint(8, 12)
+            index = 0
 
             for person in unfollow_list:
                 if unfollowNum >= amount:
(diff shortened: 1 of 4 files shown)


sionking:
what is this site:
https://www.instagram.com/web/friendships/{}/follow/ 
?

instapy/unfollow_util.py
```diff
@@ -250,10 +256,18 @@ def unfollow(browser,
...
                     try:
+                        browser.get('https://www.instagram.com/' + person)
+                        following, follow_button = get_following_status(browser, person, logger)
+                    except:
+                        logger.error(
+                            '--> Unfollow error with {},'
+                            ' maybe username has changed...'
+                                .format(person.encode('utf-8')))
+                        pass
+                    try:
```
uluQulu: Don't you think trying to get the profile page by userid should only occur **if** normal round fails?
I mean currently try/except stays in the middle of the road xD
Maybe I misunderstood it, @CharlesCCC ?

instapy/unfollow_util.py
```diff
@@ -44,16 +44,20 @@ def set_automated_followed_pool(username, unfollow_after, logger, logfolder, poo
...
                     except ValueError:
                         fword = row[0]
                         automatedFollowedPool["all"].append(fword)
                         automatedFollowedPool["eligible"].append(fword)
                 else:
                     try:
                         fword = row[0].split(' ~ ')[1]
+                        fuid = row[0].split(' ~ ')[2]
                     except IndexError:
                         fword = row[0]
```
uluQulu: here `IndexError` says that: _the line is not in the requested format to be splitted_.
It was meant to support old data before the `unfollow_after` was introduced.
I mean, why `fuid` does not exist in that except block, @CharlesCCC?

uluQulu:
First I thought it is an extra work but after reading "_most likely they have changed their username_" it sounded cool 😎
Nice job, @CharlesCCC
I rarely unfollow from the local data rather I use **nonFollowers** parameter most of time and if exists in future I would unfollow ghosts or inactive users (_at given period_) and e.g. for non followers, that case is not a problem. But your change will solve all of the issues with **customList**, **InstapyFollowed** and other scenarios where the username was taken from an old record.

And yes I agree that changing usernames is like a tradition for every user 😂
So this is also a very efficient addition, amazing!

-----------

Cheers 😁
------

sionking:
Most of the time user not exist since he block your account! btw...

instapy/unfollow_util.py
CharlesCCC: yea, good call. I will change that after I got off from work.

instapy/unfollow_util.py
CharlesCCC: the reason being if it was old data, it will not have any userid logged. so i didn't even bother to capture that. but i will add a little bit more robust logic to prevent indexerror tho.

CharlesCCC:
@sionking that `URL` should return you to the profile page. {userid}

CharlesCCC:
@sionking i don't know if somebody blocked you, you will see they doesn't exists ?

bianpratama:
@CharlesCCC, I totally agree with this addition 🎉 

But I don't know why InstaPy uses only `username` in the first place. Is there any potential risk to get detected if we use the `userid`?

> In [instabot](https://github.com/instagrambot/instabot), they use some functions which are related to this matter: `convert_to_user_id`, `get_user_id_from_username`, `get_username_from_user_id`

CharlesCCC:
@bianpratama Nothing I aware of. 
 
and I don't know why InstaPy used only `username`. it might due to the original design, it is the most convenient and straightforward way to deal with users.

timgrossmann:
@CharlesCCC @sionking @uluQulu @bianpratama Do we all agree to merge this?

uluQulu:
I have seen again and @CharlesCCC has done some neat changes in there.
@timgrossmann, I think it is ready to merge.

-------

Cheers 😁
-------

bianpratama:
@CharlesCCC, I've just started to use it, but it gives me this error:

```shell
File "/Users/username/Sites/github/InstaPy/instapy/instapy.py", line 1275, in like_by_tags
    self.user_interact_media)
  File "/Users/username/Sites/github/InstaPy/instapy/instapy.py", line 1435, in like_by_users
    self.logfolder)
  File "/Users/username/Sites/github/InstaPy/instapy/unfollow_util.py", line 516, in follow_user
    userid = browser.execute_script("return window._sharedData.entry_data.PostPage[0].graphql.shortcode_media.owner.id")
  File "/usr/local/lib/python3.7/site-packages/selenium/webdriver/remote/webdriver.py", line 465, in execute_script
    'args': converted_args})['value']
  File "/usr/local/lib/python3.7/site-packages/selenium/webdriver/remote/webdriver.py", line 236, in execute
  File "/usr/local/lib/python3.7/site-packages/selenium/webdriver/remote/errorhandler.py", line 192, in check_response
selenium.common.exceptions.WebDriverException: Message: unknown error: Cannot read property '0' of undefined
  (Session info: headless chrome=68.0.3440.106)
  (Driver info: chromedriver=2.40.565386 (45a059dc425e08165f9a10324bd1380cc13ca363),platform=Mac OS X 10.13.6 x86_64)
```

For now, it works for me by changing the `window._sharedData.entry_data.PostPage[0].graphql.shortcode_media.owner.id` in `follow_user` with `window._sharedData.entry_data.ProfilePage[0].graphql.user.id`. 

I'm wondering:
- Whether`follow_user` is initiated in the profile page or post page.
- In my case, the `follow_user` gets called from `like_by_users`. Will it be any different if it is being called from another function? What I understand from the code, `follow_user` only get the `follow_xpath` from whichever page it is being called.
- Whether it is being affected by the environment, the OS and browser.
- Should we check the userid between `ProfilePage[0]` and `PostPage[0]`?

CharlesCCC:
@bianpratama thank you for the quick feedback, while I was testing this yesterday, I didn't run into this error(I was running with minimum customization setup, like, follow). 

Would you please share your configure so i can test it out and debug it. 

in general, it doesn't matter where to get the `user.id` from, I just would like to know how/why each account's `_sharedData.entry_data` structure is somehow different ?

bianpratama:
@CharlesCCC, I just tried to manually called the graphql via console, and it turns out:
- I open a profile page, call the `ProfilePage[0]`, and it shows the `userid`.
- But when I click one of the post, and call the `ProfilePage[0]` again, it still shows the `userid`. And the `PostPage[0]` turns undefined.
- I need to refresh the post page for the `PostPage[0]` to show the `userid`.

I tried to use Chrome's React developer tools:
- I open a profile page, the props has `user` object (which has the `id`)
- I click one of the post, the props still has `user` object (because it's only opened as modal)
- When I refresh the page, or if I open the post page directly, the props has `owner` object (which also has `id`)

I think that what I mention above is somehow related to this matter. The `_sharedData.entry_data` will depend on the previous state of the browser action, CMIIW.

My InstaPy setup is heavily customised with additional Class and loops. But I don't touch the core source code. You can take a look at [this gist](https://gist.github.com/bianpratama/3fc9aef1c817ebb29467deb6ce5e719d) for the main loop.

CharlesCCC:
@bianpratama Thank you for the detailed debug info/update. 
Based off the testing I did, I can see the following with what you have found: 
```
I click one of the post, the props still has user object (because it's only opened as modal)
```
This is happing when manually open a post through Chrome, but if you observer when `InstaPy` run,  it will actually open the page (vs. Modal popup) when the post is opened.  
so that confirmed the following:
```
When I refresh the page, or if I open the post page directly, the props has owner object (which also has id)
```

one way I can see to fix the issue is to add a `try` to try both.

bianpratama:
@CharlesCCC, for now, I use this:
```python
try:
    userid = browser.execute_script(
        "return window._sharedData.entry_data.PostPage[0].graphql.shortcode_media.owner.id")
except WebDriverException:
    try:
        browser.execute_script("location.reload()")
        userid = browser.execute_script(
            "return window._sharedData.entry_data.PostPage[0].graphql.shortcode_media.owner.id")
    except WebDriverException:
        userid = browser.execute_script(
            "return window._sharedData.entry_data.ProfilePage[0].graphql.user.id")
```

CharlesCCC:
@bianpratama yep, that is the approach I was thinking. i have PRed the change #2743 
thank you 👍

mehdi691:
awesome work thank you :)

PR #2732, opened 2018-08-22
person was not able to get Follow! maybe blocked

sionking:
Check by reload that person is not follow blocked.
I have seen follow block by follow button not changing to "following" but stay "Follow.
but for any case I also reload the page each 1-5 times..

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,674 @@
-The MIT License (MIT)
+GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
 
-Copyright (c) 2016 Tim Großmann
+ Copyright (C) 2007 Free Software Foundation, Inc. {http://fsf.org/}
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
 
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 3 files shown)


timgrossmann:
@sionking Looks good, this will probably make things more clear if we really encounter a follow block.

Thank you very much for this!

P.S. Let's wait for another approval before merging this 😉

sionking:
lets wait for the crying babies get blocked :()

sionking:
I need to return also "failed to retrieve following status" from check_following function, since it might be it that we failed to read the status and we think we got blocked.

timgrossmann:
@sionking Haha, there might already be some in the issues 😅 

I think the second comment you made absolutely makes sense. We don't want the users to think they have been blocked, too often.

Thank you for your great additions

README.md
```diff
@@ -71,6 +71,7 @@ Table of Contents
   * [Pick Nonfollowers of a user](#pick-nonfollowers-of-a-user)
   * [Pick Fans of a user](#pick-fans-of-a-user)
   * [Pick Mutual Following of a user](#pick-mutual-following-of-a-user)
+* [Third Party InstaPy GUI for Windows](#third-party-instapy-gui-for-windows)
```
timgrossmann, line 74: I think you quickly need to do a rebase on master before we merge this.

sionking: I am new to git baaa

timgrossmann: Should be solved by a rebase on InstaPy master:
https://gist.github.com/ravibhure/a7e0918ff4937c9ea1c456698dcd58aa

PR #2739, opened 2018-08-23
Fix: proxy_port default value is None instead of 0

Nuqlear:
Although there is already [PR to fix such issue](https://github.com/timgrossmann/InstaPy/pull/2711) (and I am sorry for duplicate, initially I just fixed it for myself), it's incomplete and I think that changing port argument type from int to string is not a correct way to solve the problem (it may require changes for existing projects).

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -82,7 +82,7 @@ class InstaPy:
                  headless_browser=False,
                  proxy_address=None,
                  proxy_chrome_extension=None,
-                 proxy_port=0,
+                 proxy_port=None,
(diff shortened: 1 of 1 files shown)


timgrossmann:
@Nuqlear Seems like this one is a better approach. 

Is this fully functional?
My tests with this worked.

@sionking I saw you upvoted the PR, did you try it out?

Nuqlear:
Sorry for late reply.
It seems to be working fine, however I dont know for sure because my proxies are banned by instagram. 😅

PR #2743, opened 2018-08-23
fix a minor bug could cause cant get the userid from post page

CharlesCCC:
Fix a bug that reported by @bianpratama from #2727 .


timgrossmann:
@CharlesCCC @bianpratama Awesome, thank you for the quick fix!

Issue #2757, opened 2018-08-25
"//div[text()='Followers' or text()='Following']/following-sibling::div")


fgisslen:
Traceback (most recent call last):
  File "/home/fredrik/InstaPy/5b4f8b23d69086001412c068.py", line 20, in <module>
    session.interact_user_followers(['alisanikushuna','vogue_ukraine','elle_ukraine','alina_frendiy'], amount=30, randomize=True)
  File "/home/fredrik/InstaPy/instapy/instapy.py", line 1820, in interact_user_followers
    self.logfolder)
  File "/home/fredrik/InstaPy/instapy/unfollow_util.py", line 919, in get_given_user_followers
    channel, logger, logfolder)
  File "/home/fredrik/InstaPy/instapy/unfollow_util.py", line 658, in get_users_through_dialog
    "//div[text()='Followers' or text()='Following']/following-sibling::div")
  File "/usr/local/lib/python3.6/dist-packages/selenium/webdriver/remote/webdriver.py", line 393, in find_element_by_xpath
    return self.find_element(by=By.XPATH, value=xpath)
  File "/usr/local/lib/python3.6/dist-packages/selenium/webdriver/remote/webdriver.py", line 966, in find_element
    'value': value})['value']
  File "/usr/local/lib/python3.6/dist-packages/selenium/webdriver/remote/webdriver.py", line 320, in execute
    self.error_handler.check_response(response)
  File "/usr/local/lib/python3.6/dist-packages/selenium/webdriver/remote/errorhandler.py", line 242, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.NoSuchElementException: Message: no such element: Unable to locate element: {"method":"xpath","selector":"//div[text()='Followers' or text()='Following']/following-sibling::div"}
  (Session info: headless chrome=68.0.3440.106)
  (Driver info: chromedriver=2.41.578700 (2f1ed5f9343c13f73144538f15c00b370eda6706),platform=Linux 4.15.0-33-generic x86_64)

@uluQulu some more errors ;)
I will PR the other two late today.

jeffarese:
See https://github.com/timgrossmann/InstaPy/issues/2754#issuecomment-415950495

forgottensword:
@jeffarese The above replacement that you've said, is not working. I tried both. It still shows the error to find the element.

forgottensword:
[...]

PR #2758, opened 2018-08-25
fix error #2757

Mehran:
Fix error //div[text()='Followers' or text()='Following']/following-sibling::div")


timgrossmann:
@Mehran Thank you very much for this fix!

Fixes #2757

PR #2765, opened 2018-08-25
upgrade is_private_profile; get_relationship_counts switch to check f…

sionking:
…irst graphql; change logger to be more clear

### DIFF
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -75,7 +75,10 @@ def set_automated_followed_pool(username, unfollow_after, logger, logfolder, poo
         followedPoolFile.close()
 
     except BaseException as e:
-        logger.error("set_automated_followed_pool error {}".format(str(e)))
+        msg = "set_automated_followed_pool error {}".format(str(e))
+        logger.error(msg)
+        # send error in telegram
+        #send_telegram_msg(msg, logfolder, type=2)
 
     return automatedFollowedPool
 
@@ -99,8 +102,8 @@ def get_following_status(browser, person, logger):
                     following = "Requested"
     except:
         logger.error(
-            '--> Unfollow error with {},'
-            ' maybe no longer exists....'
+            '--> error get_following_status with {},'
+            ' maybe no longer exists...'
                 .format(person.encode('utf-8')))
         raise Exception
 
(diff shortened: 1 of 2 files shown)


instapy/util.py
```diff
@@ -18,10 +18,18 @@
 
 
 def is_private_profile(browser, logger, following=True):
-    is_private = None
-    is_private = browser.execute_script(
-        "return window._sharedData.entry_data."
-        "ProfilePage[0].graphql.user.is_private")
+    try:
+        is_private = browser.execute_script(
```
timgrossmann, line 22: @sionking I'm not absolutely sure about this, but shouldn't the declaration for `is_private` be outside of the try-catch block?
This looks like the scoping of the variable is now set to only the scope leading to it being `None` on the check below

sionking: I did not get you.
Please write the whole function in the comment.

timgrossmann: @sionking Sorry for the misunderstanding.

I meant this method here:
```python
def is_private_profile(browser, logger, following=True):
    try:
        is_private = browser.execute_script(
            "return window._sharedData.entry_data."
            "ProfilePage[0].graphql.user.is_private")
    except WebDriverException:
        try:
            is_private = browser.execute_script(
                "return window._sharedData.entry_data."
                "ProfilePage[0].graphql.user.is_private")
        except WebDriverException:
            return None

    # double check with xpath that should work only when we not follwoing a user
    if is_private is True and not following:
        logger.info("Is private account you're not following.")
        body_elem = browser.find_element_by_tag_name('body')
        is_private = body_elem.find_element_by_xpath(
            '//h2[@class="_kcrwx"]')

    return is_private
```

the variable is_private is only assigned (and created) inside of the try-catch block. 
However, it is used after the try-catch block, which (at least in most languages) seems like it should be None.
I'll look what python does if there is no variable creation inside the scope of the method. It probably will create a global scoped var which will then hold the given value.
If yes, this is about fine. Would be nicer to have a local variable (just setting it to None in the first line of the method or putting it inside the try-catch block)

Also, I see that you're querying the same graphql endpoint a second time if an error occurs, what is the reason for that (only for me to understand 😉 )

Thank you very much

sionking: I am missing the reload of the page before the second time.
I think is_private=Null at the start is not necessary but I will add.

instapy/unfollow_util.py
```diff
@@ -818,6 +821,14 @@ def follow_through_dialog(browser,
 
     return person_followed
 
+def unfollow_dialog(browser):
+    try:
+        unfollow_button = browser.find_element_by_xpath(
+            "//button[text()='Unfollow']")
+        if unfollow_button.is_displayed():
```
uluQulu, line 828: Nice addition, I think to add this to my PR 😛

sionking: Did you ?

uluQulu: Yeah @sionking 😂
Long time ago

PR #2816, opened 2018-09-01
Follow block branch

sionking:


timgrossmann:
@sionking Looks way better now, thank you very much. 
Let's merge this now so we get an improved following feature 😉

PR #2828, opened 2018-09-02
Proposition - Add date in general log filename

YannikFirre:
Hello,

here a proposition to avoid having a big log file. 

What do you think ?

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -205,7 +205,7 @@ class InstaPy:
             # initialize and setup logging system for the InstaPy object
             logger = logging.getLogger(__name__)
             logger.setLevel(logging.DEBUG)
-            file_handler = logging.FileHandler('{}general.log'.format(self.logfolder))
+            file_handler = logging.FileHandler('{0}{1}-general.log'.format(self.logfolder, datetime.now().strftime("%Y-%m-%d")))
             file_handler.setLevel(logging.DEBUG)
             extra = {"username": self.username}
             logger_formatter = logging.Formatter('%(levelname)s [%(asctime)s] [%(username)s]  %(message)s', datefmt='%Y-%m-%d %H:%M:%S')


timgrossmann:
@InfinityWebMe Thank you for this, considering the really huge general.log files, I think this is a good idea.
Maybe in the future I'd still go for an approach where log files are "overwritten" one a specific number of log files has been reached. (Deletes the oldest one and creates a new one). Just to keep you from having hundreds of log files... Even though it shouldn't be a problem space-wise.

What do you guys think?
@sionking @uluQulu @CharlesCCC

uluQulu:
Hi @timgrossmann @InfinityWebMe 
I think it has been raised before- having daily log files for the same [and extra] reason

My opinions are,
>https://stackoverflow.com/a/9583136   // thanks to [blueshift](https://stackoverflow.com/users/14200/blueshift) at SO

**1**-) When log file reaches a certain age or maximum size, start a new one
+ _age_: monthly;
+ _size_: `50` **MB**;
+ _naming_: **fileName_fileOpenDate.log**;

**2**-) Zip or otherwise compress the old one
+ _compression_: have **logs.zip** to store old logs;

**3**-) throw away the nth oldest compressed log
+ _oldness_: 1 year;

_**2**nd option will help to have a clearer log folder_ [IMPORTANT]
>I really think removing old files might be 🤔 matter to think cos I remember once there were an error and I looked for old logs for similar errors in order to get more information about it.  
But e.g. 1 year is too big time, so **3**rd option can also be considered.

As of your proposal, @InfinityWebMe, I think having a daily log recording will open too many log files and it will make it very difficult for e.g. troubleshooters or people looking to find some information from old logs WHERE in that case there will be _e.g._ `30*3` **=** `90` log files in just `3` months 🤓

--------

Cheers 😁
--------

YannikFirre:
@timgrossmann 

@uluQulu  I see what you mean and I agree with the opinions that you quote. 

But in the general log file, you see the search for error, me, I search the result of the script at the end (numbers of likes for example). We could perhaps put the statistics in another file to have access more easily in a first time ? 

What do you think ? 

Thanks for your time :D

uluQulu:
@InfinityWebMe 
Can you express your needs clearly?
+ Do you look at log file to get the sessional stats?
+ Do you use it to get data analysis on certain date stats?

I think with the `live_report()` function coming in my PR, it will **always** print the sessional stats in console output before ending the session.

But you should know that behaving log file as a stat recorder is a bit weird 😁

Also you can get the daily stats from **recordActivity.json** file and with my PR it will record in hourly basis.

It's a pleasure to discuss 🤝🏼 and enrich our knowledge, @InfinityWebMe 🎉

------

Cheers 😁
------

timgrossmann:
@uluQulu Exactly my thoughts about the rotating log files.
Very good summary, thank you very much.

uluQulu:
Not at all @timgrossmann 🤝🏼

@timgrossmann @InfinityWebMe 
I know that it is bit of big care hungry work and I am sure you will achieve in some day.
And meanwhile I would offer having logs recorded in monthly basis which will be a good temporary solution.

-------

Cheers 😁
-----------

CharlesCCC:
@uluQulu I think the monthly new log file make sense. usually the file wasn't get too much within the month. and I agreed that old log should be kept as if need for debug purpose. 

@InfinityWebMe thank you for your thought and contribution 👍

YannikFirre:
@uluQulu 

I launch the script on a server without visible console and for now i go check the stats in the log.

By stats, i mean this informations at the end of the script :
> INFO [2018-09-04 14:06:24] []  Liked: 57
> INFO [2018-09-04 14:06:24] []  Already Liked: 12
> INFO [2018-09-04 14:06:24] []  Commented: 0
> INFO [2018-09-04 14:06:24] []  Followed: 0
> INFO [2018-09-04 14:06:24] []  Inappropriate: 0
> INFO [2018-09-04 14:06:24] []  Not valid users: 81

And Having them in a separate log allows me to find them more easily.

sionking:
Archive general.log file bigger than 100mb.
dont remove the date.

uluQulu:
I see, @InfinityWebMe 
It is a very interesting situation 😂 and I really think there should be some work to solve your problem cos I think using log files like that is a real pain 🤦🏼‍♂️

@sionking, my PC hardly opens ~`45` **MB** of log files using **NPP** and other editors freezes for hours once trying to open it 😄☝🏼

------

Cheers 😁
--------

Issue #2817, opened 2018-09-01
BUG ALERT: (line.find(lineToDelete) < 0) will remove any username contain other user


sionking:
line = '2018-08-24 13:11 ~ blondstudiotlv,'
lineToDelete ='blo'
line.find(lineToDelete) < 0  == TRUE !!!

sionking:
Also previous implementation will remove users ends with:
if not line.endswith(lineToDelete):

sionking:
I hate regex, we need different approach

sionking:
Those are bugs for 1 - 2 weeks application :(

uluQulu:
@sionking 
`if line.split(" ~ ")[1] == lineToDelete:`
for current style only ☝🏼

Considering old data having different structure,
```python
## taken from `set_automated_followed_pool()` file:
"""
Data entry styles [historically]:
    user,   # oldest
    datetime ~ user,   # after `unfollow_after` was introduced
    datetime ~ user ~ user_id,   # after `user_id` was added
"""


## it must be checked vs all those conditions:

entries = line.split(" ~ ")
sz = len(entries)

if sz == 1:
    user = entries[0][:-2]

elif sz == 2:
    user = entries[1][:-2]

elif sz == 3:
    user = entries[1]

# OR

user = entries[0][:-2] if sz==1 else entries[1][:-2] if sz==2 else entries[1]
```

>using CSV reader will be better there

Using **regex** will be slightly faster (_unnoticable speed_) and more simpler but using simple methods (_like in above_) will make things easier to understand and in future easier to maintain in case of data style changes.

------

Cheers 😁
--------

sionking:
Most ugly database :()
We really need to think of using something more modern.
I really like this:
https://github.com/RaRe-Technologies/sqlitedict

The serialize the data and store it in sqlite.

CharlesCCC:
Actually just stepped into this issue couple days ago with my own account, it messed up my followed_pool file and my following just skyrocketed.  thank you @sionking to point it out and thank you @uluQulu for the quick fix suggestion. I will PR the fix tonight if there isn't one exists yet.  (another International business trip coming up soon, Getting super busy again :(    )

sionking:
@uluQulu I think we need your solution the one that @CharlesCCC implemented have the same bug
https://github.com/timgrossmann/InstaPy/pull/2863
But will fail when one user ends the same

PR #2863, opened 2018-09-06
Fix a potential bug that remove more than one line from the followedpool csv

CharlesCCC:
The issue reported by @sionking from #2817


sionking:
How this solving the issue ?
You are still using .endswith(lineToDelete) which can cause the issue (with less probability)

CharlesCCC:
@sionking you are right, I haven't fully think it through, kind just put it back to original way of doing things. will have come back with another/better approach. (open for suggestion)

timgrossmann:
#2893

Issue #2591, opened 2018-08-01
Raspberry view problem


skaeslin:
## Expected Behavior
It should work as usual but I get this error after my login. Does anyone know what I can do? Changing the height and width in in instapy.py did not work...

## Current Behavior
pi@raspberrypi:~/Projects/instapy2 $ sudo python3.5 firefoxExample.py

OOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOO
INFO [2018-08-01 17:06:29] [USERNAME]  Session started!
oooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo

Cookie file not found, creating cookie...
Traceback (most recent call last):
  File "firefoxExample.py", line 19, in <module>
    .set_switch_language(False)\
  File "/home/pi/Projects/instapy2/instapy/instapy.py", line 347, in login
    self.bypass_suspicious_attempt):
  File "/home/pi/Projects/instapy2/instapy/login_util.py", line 142, in login_user
    ActionChains(browser).move_to_element(login_elem).click().perform()
  File "/usr/local/lib/python3.5/dist-packages/selenium/webdriver/common/action_chains.py", line 80, in perform
    self.w3c_actions.perform()
  File "/usr/local/lib/python3.5/dist-packages/selenium/webdriver/common/actions/action_builder.py", line 76, in perform
    self.driver.execute(Command.W3C_ACTIONS, enc)
  File "/usr/local/lib/python3.5/dist-packages/selenium/webdriver/remote/webdriver.py", line 320, in execute
    self.error_handler.check_response(response)
  File "/usr/local/lib/python3.5/dist-packages/selenium/webdriver/remote/errorhandler.py", line 242, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.MoveTargetOutOfBoundsException: Message: (366.5, 780) is out of bounds of viewport width (640) and height (345)
## InstaPy configuration
on raspberry pî 3
with firefox

Issue #803, opened 2017-10-23
Headless Raspberry Pi w/ Firefox login out of bounds.


elmerilaakkonen:
## Expected Behavior
To log in normally.

## Current Behavior
InstaPy can't log in because the login-button is out of view.

> Traceback (most recent call last):
>   File "xxx.py", line 11, in <module>
>     session.login()
>   File "/home/xxx/xxx/instapy/instapy.py", line 184, in login
>     self.switch_language):
>   File "/home/xxx/xxx/instapy/login_util.py", line 24, in login_user
>     ActionChains(browser).move_to_element(login_elem).click().perform()
>   File "/usr/local/lib/python3.5/dist-packages/selenium/webdriver/common/action_chains.py", line 80, in perform
>     self.w3c_actions.perform()
>   File "/usr/local/lib/python3.5/dist-packages/selenium/webdriver/common/actions/action_builder.py", line 76, in perform
>     self.driver.execute(Command.W3C_ACTIONS, enc)
>   File "/usr/local/lib/python3.5/dist-packages/selenium/webdriver/remote/webdriver.py", line 308, in execute
>     self.error_handler.check_response(response)
>   File "/usr/local/lib/python3.5/dist-packages/selenium/webdriver/remote/errorhandler.py", line 194, in check_response
>     raise exception_class(message, screen, stacktrace)
> selenium.common.exceptions.MoveTargetOutOfBoundsException: Message: (417, 619) is out of bounds of viewport width (720) and height (469)
> 

## Possible Solution (optional)
Fixed this quickly by changing the launch resolution on instapy.py line 54:
`self.display = Display(visible=0, size=(800, 600))`
`self.display = Display(visible=0, size=(1900, 1200))`

## InstaPy configuration

Raspberry Pi 3, Raspbian Stretch Lite (2017-09-07 release)
Mozilla Firefox 52.4.0 (firefox-esr)
Selenium 3.6.0
Pyvirtualdisplay 0.2.1
Xvfb

Running with Python 3, nogui=True and use_firefox=True.

My solution isn't probably the best one, but it worked for me.
Maybe for some reason Display() doesn't launch correctly sized viewport?
Another solution could be to launch the browser maximized?

alexsmedile:
The suggested solution doesn't resolve my problem.
```
Traceback (most recent call last):
  File "quickstart.py", line 8, in <module>
    session.login()
  File "/home/pi/Projects/InstaPy/instapy/instapy.py", line 264, in login
[...]

Issue #596, opened 2017-08-28
Error using firefox mode


peguilherme77:
This error appeared at all times:

Traceback (most recent call last):
  File "quickstart.py", line 4, in <module>
    session.login()
  File "C:\Users\pedro\Desktop\Github\InstaPy\instapy\instapy.py", line 163, in login
    if not login_user(self.browser, self.username, self.password, self.switch_language):
  File "C:\Users\pedro\Desktop\Github\InstaPy\instapy\login_util.py", line 21, in login_user
    action = ActionChains(browser).move_to_element(login_elem).click().perform()
  File "C:\Users\pedro\AppData\Local\Programs\Python\Python36\lib\site-packages\selenium\webdriver\common\action_chains.py", line 77, in perform
    self.w3c_actions.perform()
  File "C:\Users\pedro\AppData\Local\Programs\Python\Python36\lib\site-packages\selenium\webdriver\common\actions\action_builder.py", line 76, in perform
    self.driver.execute(Command.W3C_ACTIONS, enc)
  File "C:\Users\pedro\AppData\Local\Programs\Python\Python36\lib\site-packages\selenium\webdriver\remote\webdriver.py", line 297, in execute
    self.error_handler.check_response(response)
  File "C:\Users\pedro\AppData\Local\Programs\Python\Python36\lib\site-packages\selenium\webdriver\remote\errorhandler.py", line 194, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.MoveTargetOutOfBoundsException: Message: (932.5, 666) is out of bounds of viewport width (1366) and height (659)

Firefox opens the window on the Instagram page, but it does not get there.

converge:
hey @peguilherme77 , did you managed to fix it ?

PR #2865, opened 2018-09-07
Fix MoveTargetOutOfBoundsException on Legacy Firefox

nyx-rattapoom:
I'm use InstaPy in raspberry pi and cannot move to newer version of firefox.
I found some bug about MoveTargetOutOfBound if use method move_to_element in ActionChain class
in it like bug in selenium.
bug suggest to fix with manual move to element with alternative way and to something with it. 
I use
Raspberry pi 2
Selenium: 3.4.3
Geckodriver: 0.17.0
Firefox-esr: 52.9.0esr-1~deb9u1

Relate issue: #596 #803 #2591

Solution:
I wrote patch to move to target element before call real method.
I don't call patch in part of InstaPy code to prevent people who use old selenium or geckodrive but I also add example feel free to use.

### DIFF
--- a/examples/firefox_patcher.py
+++ b/examples/firefox_patcher.py
@@ -1,0 +1,31 @@
+
+from instapy import InstaPy
+from instapy.monkey_patcher import patch_all
+
+#
+# use_firefox=True : Switches the browser from Chrome to Firefox/Geckodriver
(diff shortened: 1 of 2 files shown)


timgrossmann:
@nyx-rattapoom Awesome, thank you very much!

Issue #2882, opened 2018-09-09
shadow ban using unfollow


jtrieuDev:
Hi everyone, I am trying to use the unfollow methods

but every time I am launching the script, i got this:

"There is a serious issue: 'shadow ban'!	~leaving Unfollow-Users activity
INFO [2018-09-10 01:09:48] [username]  --> Total people unfollowed : 1
"

Here is my script:
    session.login()
    session.set_relationship_bounds(enabled=False,
				 potency_ratio=-1.21,
				  delimit_by_numbers=False,
				   max_followers=4590,
				    max_following=5555,
				     min_followers=45,
				      min_following=77)
    session.set_simulation(enabled=False)
    session.unfollow_users(amount=400, InstapyFollowed=(True, "all"), style="FIFO", unfollow_after=None, sleep_delay=501)

davidvee7:
I'm also running into this issue

kukkakeppi:
Same here.. No idea what to do.

timgrossmann:
@jtrieuDev @davidvee7 @kukkakeppi Could you guys pull and try again, please?
@uluQulu should have fixed it 😉

jtrieuDev:
it is still happening for me.. 
sometimes it goes until more than 1x unfollows but sometimes still stop after 2-3 unfollowers

uluQulu:
Hey guys
@timgrossmann I just noticed that bug while running a test for this problem,
>There is a serious issue: 'shadow ban'! ~leaving Unfollow-Users activity

I mean my latest PR solves another bug but not the one above.
It's a bit mystery, needs more feedback.

It happens once it clicks to unfollow a user and then it checks again and user is still not unfollowed then points out a possible shadow ban.

@jtrieuDev Can you run this test?
Open the session in graphical mode and watch what happens:
+ Does it click on "_Following_" or "_Requested_" button?
+ Does the "_Unfollow_" dialog pop up?  & Does it click on it?
+ Does that button now is "_Follow_" or "_Follow Back_"?

It's the same procedure introduced by @sionking, I have not changed anything in that part, just error message has changed 😉

_Anybody having that exact error message could run that test_.

-----

Cheers 😁
--------

jtrieuDev:
Hi @uluQulu thank you for your answer,
I just did the test and

Yes, it does click on "Following" or "Requested" button
Yes, the "Unfollow" dialog does pop up, and it clicks on it
Yes, that button is either "Follow" for "Follow back"

[...]

PR #2886, opened 2018-09-10
Bug Fix: make sure `unfollow_list` is always a list (rather than a `dict_keys` object)

uluQulu:
Problem is when shuffling COS `random.shuffle` cannot work with `dict_keys` objects
For example in my python 3,
```python
zx = {1:2, 3:4}
sl = zx.keys()   # sl is dict_keys([1, 3])
ld = list(zx.keys())   # ld is [1,3]
```

Hint: There is no problem while iterating over `dict_keys` objetcs

_It happened specially after I refactored `set_automated_followed_pool`_

Still wonders how I did not notice bug this before 🤔 until running a test for #2882

------

Cheers 😁
-----

### DIFF
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -204,7 +204,7 @@ def unfollow(browser,
 
         elif InstapyFollowed == True:
             logger.info("Unfollowing the users followed by InstaPy\n")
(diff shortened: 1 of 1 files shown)


timgrossmann:
@uluQulu Thanks for the fix 😉

PR #2893, opened 2018-09-11
revert of #2863

timgrossmann:


sionking:
This is even worst then the one @CharlesCCC had.

CharlesCCC:
uhh, yea, i don't think this going to work @timgrossmann  :(

timgrossmann:
@sionking @CharlesCCC Well, that's exactly what we had before #2863.

Following the comment of @CharlesCCC 
```
kind just put it back to original way of doing things.
```
I reverted it back to how it was before until we have another solution.

PR #2908, opened 2018-09-12
Xpath to focus in the comment field

keryc:
replace: //a[@role='button']/span[text()='Comment']/..
to: //button/span[@aria-label='Comment']

### DIFF
--- a/instapy/comment_util.py
+++ b/instapy/comment_util.py
@@ -29,7 +29,7 @@ def open_comment_section(browser, logger):
         '--> Warning: Comment Button Not Found:'
         ' May cause issues with browser windows of smaller widths')
     comment_elem = browser.find_elements_by_xpath(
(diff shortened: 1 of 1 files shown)


timgrossmann:
@kdzch Thank you for this change, is this something that happens in "every" release of Instagram or is this, again, some A/B-Test rollout thing?

I just tested it and I get the same problem when trying to get the comments.

keryc:
The previous xpath failed, gave an error when trying to write in a field that was not selected.

keryc:
At this moment I am running the robots and they are running correctly.

keryc:
Well the tests were done with firefox

timgrossmann:
@kdzch Double checked it in Safari, FireFox and Chrome.
Looks good, thank you!

PR #2925, opened 2018-09-15
Add decision making in interaction

sionking:
When interaction with a user start we shall:
* do at least one action follow/like/like+comment if validation is pass
* dont follow if one image in posts fail in check_image (optional* default is True)
* roll decision again in second post and beyond.
* fix: don't try to follow (again) when amount of posts is more than 1.
* save time by not checking the data bases over and over (static decisions)

Please verify logic.

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1906,14 +1906,14 @@ class InstaPy:
             # decision making
             # static conditions
             not_dont_include = username not in self.dont_include
-            follow_restrected = follow_restriction("read", username,
+            follow_restricted = follow_restriction("read", username,
                                                     self.follow_times, self.logger)
             counter = 0
             while True:
                 following = (random.randint(0, 100) <= self.follow_percentage and
                              self.do_follow and
                              not_dont_include and
-                             not follow_restrected)
+                             not follow_restricted)
                 commenting = (random.randint(0, 100) <= self.comment_percentage and
                               self.do_comment and
                               not_dont_include)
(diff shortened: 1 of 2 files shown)


sionking:
can someone try?

timgrossmann:
@sionking Just tested it and it seems to work fine, when looking through the code I noticed a tiny typo, fixed that before merging.

Thanks for the improvements, @sionking 
Really appreciate that!

sionking:
But more important, is this something tht we all agree we need ?

timgrossmann:
@sionking Good question.

For me: 

> - do at least one action follow/like/like+comment if validation is pass

That's the only thing I actually can't say anything about... For me, this is kind of weird, however it seems like some users forget to actually put some settings into the quickstart when using the interact function. Encountering a problem they might not know how to fix.
Also, could you point me to the part in the code where you do one of the actions if the validation passes? Somehow I might understand this wrong.
Is this simply the part which skips if all the settings are not set?

> - dont follow if one image in posts fail in check_image (optional* default is True)

Absolutely believe that this is what users want (including me)

> - roll decision again in second post and beyond.

Should've been like that from the beginning. Makes way more sense

> - fix: don't try to follow (again) when amount of posts is more than 1.

Totally valid, just wastes time

> - save time by not checking the data bases over and over (static decisions)

Also totally valid

sionking:
> Also, could you point me to the part in the code where you do one of the actions if the validation passes? Somehow I might understand this wrong.
> Is this simply the part which skips if all the settings are not set?

I don't understand what you meant sorry.

timgrossmann:
@sionking 
I just saw this:
> do at least one action follow/like/like+comment if validation is pass

And for me it reads like even if the user has nothing enabled with `.set_do_...` you simply do one interaction just because the user is pass

However I can't find this behaviour in your code so I guess I misunderstood this.

PR #2927, opened 2018-09-15
fix proxy

esdotzed:


timgrossmann:
@esdotzed Good catch, thank you!

fgisslen:
Haha finally, I have changed that manually on each fork of instaPy, I should begin to make pull requests....

timgrossmann:
@fgisslen Definitely 🙈 😉 

Especially for such small fixes that won't need a lot of reviewing and are only used by few people

PR #2931, opened 2018-09-16
Update comment_util.py, comment button not working with newest insa. Keys.return works. exported from selenium.webdriver.common.keys import Keys, didnt do that in past

kamilk91:


timgrossmann:
@kamilk91 Thank you very much for this and congrats to your first PR to InstaPy.
Could you help me understand this?

How is it possible to replace an element with a "KEY" from Selenium?
After the comment_elem assignment there is a `.click`executed on that element. 

@sionking @CharlesCCC @uluQulu Does this work for you?
If I do this manually in the browsers it looks like nothing happens when I trigger `RETURN`

kamilk91:
@timgrossmann hmmm maybe i messed something, but problem was simply - IG hide button, which can be replaced by return. Another way to execute this kind of stuff is simply use `\n` on the end of line. 
Its working for me

CharlesCCC:
@kamilk91 it seems like, the comment should still there. but when you don't see the `submit` comment button, you can send the `return` key instead ??? If that the case, I don't think we really should remove the original approach, but add additional logic to see if the button is missing or not, then press the `return`

PR #2938, opened 2018-09-17
update instapy to disable image load in chrome webdriver from #1364

timgrossmann:
Kudos to @ganeshkrishnan1

### DIFF
--- a/README.md
+++ b/README.md
@@ -1518,7 +1518,11 @@ If you want to save some bandwidth, you can simply disable the image/video loadi
 
 To do this simply pass the `disable_image_load=True` parameter in the InstaPy constructor like so:
 ```python
-
+session = InstaPy(username=insta_username,
+                  password=insta_password,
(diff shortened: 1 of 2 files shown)


sionking:
I think you flag yourself as a bot when you do that.
Maybe not now, but one day IG will check this.

timgrossmann:
@sionking True that.
In most cases the bandwidth isn't even a problem.

Let's add this as default `False` so people have an easy way to add it if they want to.
Still, partially loading images or restricting the image loading to a lower quality won't work that easy.

uluQulu (approved):
It is amazing, @timgrossmann 🎊🎊

CharlesCCC:
change looks good. default to `False`  make perfect sense.

CharlesCCC (approved):
Let's merge this 👍

sionking:
this is good when using luminti and other expensive proxy providers.

Issue #2939, opened 2018-09-17
Skip Private/Business/No Profile Pic, Min/Max posts


timgrossmann:
> https://www.bountysource.com/issues/63470533-skip-private-business-no-profile-pic-min-max-posts

## Expected Behavior
There are basically two parts to this request.
1. Extending the relationship bounds with the min/max amount of posts.
2. Adding the `set_skip_users` feature which will allow InstaPy-users to skip:
    - Private Profiles
    - Profiles without profile pictures
    - Business profiles
        - Since there is a fixed amount of business categories, also skipping users of specific categories/only using users with a specific category

There already are older feature requests for this stuff.
#980 (Skip Business - $25 Bounty)
#1006 (Skip private - $33 Bounty)

> I will add another $42 to this Issue. So the whole feature will have a bounty of $100.
Whoever implements this, please think about sharing it with @marcomokastyle since he already did a lot of the work.

## Possible Solution (optional)
There already is an old, outdated PR which had some of the stuff implemented, the author unfortunately is not able to complete it.
#2242
In the comment section there also is information about how ti find whether it's a user profile or not.

##### Min/Max media
```python
set_relationship_bounds(enabled=True,
                                    min_media=5,
                                    max_media=100
				    max_following=5555)
```

##### Profile skipping
```python

# skip all private profiles, profiles without profile pictures and all business profiles
set_skip_users(skip_private=True, 
                          skip_no_profile_pic=True, 
                          skip_business=True)

# only skip all business profiles of the given categories
set_skip_users(skip_business=True,
                          skip_business_categories=['Creators & Celebrities', '...'])


# skip all but the given categories
set_skip_users(skip_business=True,
                          dont_skip_business_categories=['Home Services'], '...')
```

C&C very welcome, this is a community project so we want to have everyone agree on what we want for InstaPy.
Thanks guys for making this tool awesome.

> Any comments @sionking @CharlesCCC @uluQulu ?

Best
Tim

photogsy:
I've actually just implemented something similar to suit my own needs - happy to contribute if you like.
My solution was a bit of a "quick & dirty" approach, but I think it could easily be adapted to be more flexible.

While rooting about in the code it struck me that there is a LOT of repetition that really should be refactored into more flexible functions...

For example, wherever we attempt to get some information by using `graphql `and then falling back to  `find_element_by_xpath` - this should be encapsulated in a function.

Anyway, my solution was to add a function into utils.py as follows:

```
[...]

PR #2957, opened 2018-09-19
validate private users and business users and some extra stuff

sionking:
according to:
https://github.com/timgrossmann/InstaPy/issues/2939

**Private**:
under interaction it will act according to follow_private_percentage
other will based on True/False logic only set in skip_private
**business**:
Setting a number in self.parameters['interact_business'] 0-100
default: 80, will  skip 2/10 business users
use:
```
    def set_custom_settings(self, interact_business=80, skip_private=True):
        """Defines custom settings """
        if self.aborting:
            return self

        self.parameters['interact_business'] = interact_business  # default: will skip 2/10 business profiles
        self.skip_private_users = skip_private
```
while also added the abilaty to follow private by percentage:
```
def set_do_follow(self, enabled=False, percentage=0, follow_private_percentage=0, times=1):
```
which will only work in interaction.

**min/max posts:**
Currently only validated min posts.
use:
```
set_relationship_bounds(enabled=True,
                                    min_posts=5,
                                    max_posts=100
				    max_following=5555)
```

**EXTRA:**
improved ignore by using sets:
self.ignore_users = set()

like_by_must_like:
for like_by_tags and like_by_locations
I added an optional parameter for those who asks not to do any likes, so basicly those functions now act by interact rules.

followFeed:
optional to follow Feed users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1519,7 +1519,7 @@ To do this simply pass the `disable_image_load=True` parameter in the InstaPy co
 session = InstaPy(username=insta_username,
                   password=insta_password,
                   headless_browser=False,
-		  disable_image_loading=True,
+		  disable_image_load=True,
                   multi_logs=True)
 ```
 
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1936,11 +1936,11 @@ class InstaPy:
                 if commenting and not liking and amount == 1:
                     continue
                 if following or commenting or liking:
-                    self.logger('username actions: following={} commenting={} liking={}'.format(following, commenting, liking))
+                    self.logger.info('username actions: following={} commenting={} liking={}'.format(following, commenting, liking))
                     break
                 # if for some reason we have no actions on this user
                 if counter > 5:
-                    self.logger('username={} could not get interacted'.format(username))
+                    self.logger.info('username={} could not get interacted'.format(username))
                     break
 
             try:
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -190,7 +190,7 @@ class InstaPy:
         self.max_comments = 35
         self.min_comments = 0
         
-        self.parameters = collections.defaultdict(dict)
+        self.parameters = defaultdict(dict)
         self.parameters['interact_business'] = 80  # will skip 2/10 business users
         self.skip_private_users = False
         
@@ -3151,7 +3151,6 @@ class InstaPy:
                                                                 user_name,
                                                                 None,
                                                                 self.blacklist,
-                                                                False,  # is_private,
                                                                 self.logger,
                                                                 self.logfolder)
                                         if follow_state == True:
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -405,7 +405,7 @@ class InstaPy:
         if self.aborting:
             return self
 
-        self.parameters['interact_business'] = inputs['interact_business']  # default: will skip 2/10 business profiles
+        self.parameters['interact_business'] = interact_business  # default: will skip 2/10 business profiles
 
         self.skip_private_users = skip_private
 
@@ -1758,7 +1758,6 @@ class InstaPy:
                                         username,
                                         None,
                                         self.blacklist,
-                                        is_private,
                                         self.logger,
                                         self.logfolder)
                 if follow_state == True:
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1995,11 +1995,11 @@ class InstaPy:
                if commenting and not liking and amount == 1:
                    continue
                if following or commenting or liking:
-                    self.logger('username actions: following={} commenting={} liking={}'.format(following, commenting, liking))
+                   self.logger('username actions: following={} commenting={} liking={}'.format(following, commenting, liking))
                    break
                # if for some reason we have no actions on this user
                if counter > 5:
-                    self.logger('username={} could not get interacted'.format(username))
+                   self.logger('username={} could not get interacted'.format(username))
                    break
 
             try:
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1985,23 +1985,23 @@ class InstaPy:
                                  self.do_follow and
                                  not_dont_include and
                                  not follow_restricted)
-                commenting = (random.randint(0, 100) <= self.comment_percentage and
-                              self.do_comment and
-                              not_dont_include)
-                liking = (random.randint(0, 100) <= self.like_percentage)
+               commenting = (random.randint(0, 100) <= self.comment_percentage and
+                             self.do_comment and
+                             not_dont_include)
+               liking = (random.randint(0, 100) <= self.like_percentage)
 
-                counter += 1
+               counter += 1
 
-                # if we have only one image to like and comment and we only want to comment without like
-                if commenting and not liking and amount == 1:
-                    continue
-                if following or commenting or liking:
-                    self.logger.info('username actions: following={} commenting={} liking={}'.format(following, commenting, liking))
-                    break
-                # if for some reason we have no actions on this user
-                if counter > 5:
-                    self.logger.info('username={} could not get interacted'.format(username))
-                    break
+               # if we have only one image to like and comment and we only want to comment without like
+               if commenting and not liking and amount == 1:
+                   continue
+               if following or commenting or liking:
+                    self.logger('username actions: following={} commenting={} liking={}'.format(following, commenting, liking))
+                   break
+               # if for some reason we have no actions on this user
+               if counter > 5:
+                    self.logger('username={} could not get interacted'.format(username))
+                   break
 
             try:
                 links = get_links_for_username(self.browser,
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -2945,8 +2945,8 @@ class InstaPy:
                                 amount=50,
                                  randomize=False,
                                   unfollow=False,
-                                   interact=False):
-                                    followFeed=False
+                                   interact=False,
+                                    followFeed=False,
                                      validate_feed_users=False):
         """Like the users feed"""
 
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -60,6 +60,7 @@ from .relationship_tools import get_nonfollowers
 from .relationship_tools import get_fans
 from .relationship_tools import get_mutual_following
 from .database_engine import get_database
+from collections import defaultdict
 
 # import exceptions
 from selenium.common.exceptions import NoSuchElementException, WebDriverException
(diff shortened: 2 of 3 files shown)


sionking:
@timgrossmann @uluQulu @CharlesCCC 
Can you verify and not let this PR stand for too long ?

timgrossmann:
@sionking Will do later today, thanks for the PR.
Could you resolve the conflict that was created due to merging your logger fi, please?

sionking:
Which conflict? It WA just 2 line changing..  soory I am still learning this git

timgrossmann:
@sionking I think it's just a `git rebase master` on this branch.

sionking:
Done rebase

timgrossmann:
@sionking Thank you very much for this PR!

I'm sorry but I'm a bit confused about it in some parts.

1. When you wrote about using `self.parameters['someval']` is somehow thought that you meant passing the whole `self` variable into a function and then 'destructuring' it in the function to get all the needed parameter. I misunderstood the part about the `.parameters`, I'm not sure about this, this looks kind of wrong... I'd prefer to stick with the old variant without the parameters 'inner' dictionary.

2. I'm absolutely not sure about this: `set_custom_settings`. From what I experienced with some people that got support from me, the problem is that it's too non descriptive. I'm quite sure it will be problematic for a lot of people to understand this, that's why I defined it as `skip_users` which is clear right away.
Also, the business_profile part is also kind of unexpected... simply skipping a specific percentage of business accounts is not intuitive in my view.
I'm absolutely not sure about all of this, the definition in the Issue is only my point of view, maybe some users will also have problems with that but I think the  `skip_users` is a better approach than `set_custom_settings` which is too general and gives the user too much power (seems like it can/will in the future).

Sorry for the listed points, but I think I simply was expecting the predefined adjustments and features added to InstaPy. I hope you can understand this.

Edit: Thinking about it again and after looking at the python style guide, I think I made a huge mistake in telling you that the `self.parameter` thing wouldn't be too bad... Even though it works in JS doesn't mean it's something that should be used in Python. 
I'm really sorry about this and the work you put into this PR...

sionking:
@timgrossmann 
So what is your solution for **not** passing tons of different vars into validate_users should we simply pass the instapy object? I can't stand when the validate_user function takes to much place in the code.
self.parameter I think is a great idea, and very flexible, why not using it in python ?

sionking:
quickstart.py when managing many users you can't base configuration for each user in the code.
I don't mind change set_custom_settings to what you expected.

sionking:
@uluQulu what is your opinions on decreasing the number of parameters we pass to this function.
BTW I pass 10 more params with the self.parameter dict.

uluQulu:
@sionking 
I have really limited time to checkout your work in details and I already had a quick glance on it and I totally agree with @timgrossmann in most cases.

For that part, although I'm not very aware of your style, I think having parameters passed explicitly is more clearer.
I really agree with your point but maybe it can be handled differently- yet in a clear way.

The only time I have requested this kind of style is with `explicit_wait()`:
```python
explicit_wait(browser, track, ec_params, logger, timeout=35, notify=True)
```
`ec_params` parameter itself takes a list of parameters, why?
Cos different expected conditions need different kind of specific parameters.
_e.g._,
+ "**visibility of element located**" condition needs 2 specific parameters: `[element_address, find_method]` 
+ "**title contains**" condition needs only one specific parameter: `[expect_in_title]`
+ "**page fully loaded**" condition does not need any specific parameter: `[]`

Else than this kind of situation, passing parameters separately is the best way.

-----

Cheers 😊
----

timgrossmann:
@sionking Sorry for not responding earlier... 

Well that's a good question on how to make this "better".


I think there are basically three ways we could handle that which all would probably be a bit nicer than the current approach:

1. Only pass the full self object to the methods and then retrieving the elements in the method. (This will loose all the internal "code documentation" on what the methods need or don't need)
We could probably also pass in `**self` and then only pull out the arguments we need in the method signature. (If that works, would definitely be a, at least a bit, cleaner solution)

2. Pass every parameter (will only increase over time making the code really hard to read at one point/also very lengthy)

3. Create a parameters dict for every functionality (not `self.paramters` but something like `self.like_params`, `self.follow_params` etc.)
However this will probably lead to a lot of duplication...

I'll read up on handling large amounts of parameters and how to best handle this.

timgrossmann:
@sionking I'm sorry but I don't really understand what you mean with this:
> quickstart.py when managing many users you can't base configuration for each user in the code.

Could you please elaborate on that?

sionking:
quickstart :

> @sionking I'm sorry but I don't really understand what you mean with this:
> 
> > quickstart.py when managing many users you can't base configuration for each user in the code.
> 
> Could you please elaborate on that?

Think that you have configuration for each user, than each will need a different quickstart.py, changing the code to change the inputs is wrong approach ! the input should not be a python file, and if it is than let is be just a tunnel to the real code.

JUST a reminder on how validate_user looks like:
```
                validation, details, return_values_dist = validate_username(self.browser,
                                                        username,
                                                        self.username,
                                                        self.ignore_users,
                                                        self.blacklist,
                                                        self.potency_ratio,
                                                        self.delimit_by_numbers,
                                                        self.max_followers,
                                                        self.max_following,
                                                        self.min_followers,
                                                        self.min_following,
                                                        skip_private,
                                                        self.parameters,
                                                         ...
                                                        self.logger)
```
* 17 times !

we can have:

```
                validation, details, return_values_dist = validate_username(self.browser,
                                                        username,
                                                        self.username,
                                                        self.ignore_users,
                                                        self.blacklist,
                                                        self.parameters,
                                                        self.logger)
```
And will not increase in the future. Still think this solution is great!
Also we can have parameters['relational_tools']['max_followers'] and so on ...

timgrossmann:
@sionking Thank you, now I see what you were saying.

I absolutely do understand the concerns with the length of the methods.
It somehow looks like this method is doing too much stuff in one big method and therefore has a ton of args.
I still don't believe in "blurring" the signature of methods... I know and understand what you're pointing at. But making it mandatory to read into the code body of a method to understand what arguments get passed in is not good and very unreadable.

Regarding the functionality and interface, I'd still love to have the skip_users method.
Focusing on people that use InstaPy solely for them and not for dozens of accounts, this interface makes it way more readable and understandable for them.

Hope you can understand that, @sionking 
Thanks for everything you're doing for our project

Best
Tim

sionking:
> Regarding the functionality and interface, I'd still love to have the skip_users method.

OK!

we can have name changed to validation_parameters, BTW in this open source project having the abilaty to modify the code with minor changing (I have 10 more validation parameters in this dict and it will look the same as instapy.py original) is a must.

timgrossmann:
@sionking Thank you very much for understanding this.

I absolutely see your points for deciding for the parameters approach.
I'm sorry but we can't just move the signature of a method from
```python
validation, details, return_values_dist = validate_username(self.browser,
                                                        username,
                                                        self.username,
                                                        self.ignore_users,
                                                        self.blacklist,
                                                        self.potency_ratio,
                                                        self.delimit_by_numbers,
                                                        self.max_followers,
                                                        self.max_following,
                                                        self.min_followers,
                                                        self.min_following,
                                                        skip_private,
                                                         ...
                                                        self.logger)
```

to something like 

```python
validation, details, return_values_dist = validate_username(self.browser,
                                                        self.parameters,
                                                        self.logger)
```

This just removes the whole context.

What I'd be absolutely ok with is to have something like this:

```python
def test(a, b):
  print(a, b)
  
  
args = {'a': 1, 'b': 2}
test(**args) #=> prints (1, 2)
```

Which will keep the signature of the method, describing what parameters are expected, also keeping the possibility to assign default values.

It will automatically assign the values to the vars given the right naming, the key has to be variable name.

```python
def test(b, a):
  print(a, b)
  
  
args = {'a': 1, 'b': 2}
test(**args) #=> prints (1, 2)
```

With this approach, we can also make sure that all the mandatory arguments are enforced.

```python
def test(b, a, c):
  print(a, b)
  
  
args = {'a': 1, 'b': 2, 'd':3}
test(**args) # => throws unexpected argument d 
```

and 

```python
def test(b, a, c):
  print(a, b)
  
  
args = {'a': 1, 'b': 2}
test(**args) # => expects 3 arguments, 2 given 
```

What do you guys think about that?

sionking:
@timgrossmann 
So it is basically what I made just need to change the validate_username input ?

timgrossmann:
@sionking I think so, yes.
However it keeps the signature of the method and does not simply squash it together.

Could you adjust your PR to this?
Also, creating the dict of the parameters for this specific version might even be done right before the method is called instead of the InstaPy constructor.

sionking:
> Also, creating the dict of the parameters for this specific version might even be done right before the method is called instead of the InstaPy constructor.

NO NO, Do you really want to create the dict before each call 👎 ?
It need to be created once in the constructor or in a specific method (and we shall have no duplication).
Any other method would be able to use it as well. 👍  (or only part of the parameters)

timgrossmann:
@sionking I just meant that if you really want to only have a `self.parameters` you need to do it in the methods themselves since you can't just have a huge `self.paramters`with everything in it and then use this and pass it to methods.

sionking:
> @sionking I just meant that if you really want to only have a `self.parameters` you need to do it in the methods themselves since you can't just have a huge `self.paramters`with everything in it and then use this and pass it to methods.

So I don't understand what we want to do ... 
"you need to do it in the methods" ? the parsing ...
"have a huge `self.paramters`" ? why not ? you can use that ever you need...

timgrossmann:
@sionking If you look at the example I gave above, you can't just have a single giant parameters dict.
It will throws an error if you pass it to a method using the double asterisk notation. It needs to have the exact same dictionary members as the signature of the method states.

sionking:
> exact same dictionary members as the signature of the method states.

I think we shall throw exception only in case there is a missing arguments not when we have extra.

sionking:
@timgrossmann 
The correct solution will be:
```
def print_values(**kwargs):
    for key, value in kwargs.items():
        print("The value of {} is {}".format(key, value))

print_values(my_name="Sammy", your_name="Casey")
```

timgrossmann:
@sionking From the definition of the keyword argument unpacking syntax (**) this is not possible. It will move pass all the elements (key, val) from the dict into the method.

I think it's also not really clean to simply pass a huge pack of parameters into every method, feels kind of weird.

sionking:
So my best option is to make some parameters input as we do now, ans some in **KWORD 

```
                validation, details, return_values_dist = validate_username(self.browser,
                                                        username,
                                                        self.username,
                                                        self.ignore_users,
                                                        self.blacklist,
                                                        self.logger,
                                                        self.validate_username_parameters)
                                                        
```

and self.validate_username_parameters will include **all** and **only** parameters for this function.
We also need not to keep duplicates so we shall move some currently have parameters to this function.


?

timgrossmann:
@sionking That definitely sounds reasonable

sionking:
So this branch is not dead I will PR some of the relevant changes.

Issue #2953, opened 2018-09-19
Unfollow issue (not logged in, not connected and really slow)


tompicca:
For several days now I had no luck with unfollow session. I've used always the last commit but something always gone wrong.
I need to know if there's a problem on my side or something needs to be reviewed with the last update.

Below the log.

The script tried to unfollow 2 users before report the serious issue: `not logged in`.
Sometimes I also get the error `not connected`.

Also, to check the first user the script takes more than 5 minutes, I believe it is a biblical time compared to the old versions.

I have the impression that InstaPy worked better and faster some time ago, even with some bug. Anyone with the same issue? Any solution?

```
INFO [2018-09-19 11:04:14] [user]  Starting to unfollow users..
"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
INFO [2018-09-19 11:04:14] [user]  Unfollowing from the list of pre-defined usernames

INFO [2018-09-19 11:04:14] [user]  Total 2726 users available to unfollow  ~not found in 'followedPool.csv': 0  |  didn't pass `unfollow_after`: 0

INFO [2018-09-19 11:04:14] [user]  Ongoing Unfollow [1/100]: now unfollowing 'b'dorsey_8365__darron''...
INFO [2018-09-19 11:05:33] [user]  Timed out with failure while explicitly waiting until visibility of element located!

INFO [2018-09-19 11:06:50] [user]  Timed out with failure while explicitly waiting until visibility of element located!

ERROR [2018-09-19 11:06:50] [user]  --> Unable to detect the following status of 'b'dorsey_8365__darron''!
INFO [2018-09-19 11:06:50] [user]  Pinging 'instagram.com' to check the connectivity...
PING instagram.com (54.209.17.35): 56 data bytes
64 bytes from 54.209.17.35: icmp_seq=0 ttl=234 time=137.672 ms

--- instagram.com ping statistics ---
1 packets transmitted, 1 packets received, 0.0% packet loss
round-trip min/avg/max/stddev = 137.672/137.672/137.672/0.000 ms
INFO [2018-09-19 11:06:50] [user]  Checking if 'user' is logged in...
WARNING [2018-09-19 11:06:53] [user]  Maybe 'b'dorsey_8365__darron'' has changed the username!        ~verifying through the user ID
INFO [2018-09-19 11:08:13] [user]  Timed out with failure while explicitly waiting until visibility of element located!

[...]

PR #2978, opened 2018-09-21
Bunch of improvement and solution for the latest bugs out of misbehaviours known as 'not connected' & 'log-out' issues

uluQulu:
Hello

Having caught a terribly seasonal grip I will be short in description and for that reason I made the commits highly separate for you to understand.

The only thing needs to be mentioned is,
While getting the username from the user ID using the IG Follow endpoint- 
`https://www.instagram.com/web/friendships/{user_id}/follow/`
it was sometimes logging out and redirecting to the
`https://www.instagram.com/accounts/login/?next=/web/friendships/{user_id}/follow/`
webpage after logging out the user.

I have talked about debugging it at #2953 in details - special thanks to @spikeyhair for great feedback.

After having diagnosed the issue, I have offered different approaches to do it- getting the username from the user ID.
Now, it has 3 steps:
#### method using graphql 'Account media' endpoint
+ It loads a posts of a user from the give user ID then gets the shortcode of a post and navigates to it and the gets the username from there as usual.
+ It's using the same graphql URL- "https://www.instagram.com/graphql/query/?query_hash={}" we use for relationship tools and is super stable.
+ It has drawbacks- can load posts of users only if it is non-private [or private and is being followed] and works only the user has a post [photo or video]
#### method using private API
`https://i.instagram.com/api/v1/users/{user_id}/info/`
+ If you disagree, I can remove that method easily but I think it is cool link and has no problem and also is optional and we do not rely on it.
+ Good news is it gets the username of any user- private or non-private

#### method using graphql 'Follow' endpoint
The same method with the bug- `https://www.instagram.com/accounts/login/?next=/web/friendships/{user_id}/follow/`
+ If you disagree, we can remove it. But note that this 3rd step is gonna be requested only in rare cases cos almost everytime prior 2 steps will get the username.

----

Cheers 🙋🏼‍♂️
-----


sionking:
Thanks, but none of the solution is normal user use case (the post one is, but ignore private users).
The only solution in my opinion is to make a search in the search bar for the username and take the first result.
It supposed to be correct and also can be used for other things as well, also make it non boty.

pagokrat0r:
With the new commits this is what happens now on "Sorry, this page isn't available." cases.

```
INFO [2018-09-22 00:42:56] [************]  Ongoing Unfollow [11/800]: now unfollowing 'b'everything_everyone_smile''...
INFO [2018-09-22 00:44:16] [************]  Timed out with failure while explicitly waiting until visibility of element located!

INFO [2018-09-22 00:45:33] [************]  Timed out with failure while explicitly waiting until visibility of element located!

ERROR [2018-09-22 00:45:33] [************]  --> Unable to detect the following status of 'b'everything_everyone_smile''!
INFO [2018-09-22 00:45:33] [************]  Pinging 'instagram.com' to check the connectivity...
CRITICAL [2018-09-22 00:45:37] [************]  There is no connection to the 'instagram.com' server!
WARNING [2018-09-22 00:45:37] [************]  There is a serious issue: 'not connected'!	~leaving Unfollow-Users activity
INFO [2018-09-22 00:45:37] [************]  --> Total people unfollowed : 10
```

uluQulu:
@sionking 
>The only solution in my opinion is to make a search in the search bar for the username and take the first result.

Your suggestion is poor to be a solution so far and I have stated it 2 times before and is prone to cause misbehaviours out of inaccuracy.
example
One more fact, that algorithm of that search box is very fuzzy and even fails to find the user with the non-changed username WHOM a profile is still following.  I don't even talk about the changed username.

>...also can be used for other things as well...

Using search box for other things can be considered.

>Thanks, but none of the solution is normal user use case (the post one is, but ignore private users).

We use graphql for loading followers/following data and it navigates to more than ~100 pages of that sort in one run to load 50x100=5000 usernames.
It has not been a problem and I think loading only one page of similar sort will not either make a problem- so the statement of "_none of the solution is normal user use case_" is not correct.

Well, even though both 2nd and 3rd options are to dismissed and only the 1st method is used, then it will at least find ~80% of users who is non-private [or is private and is being followed] and has at least one post.
Considering that earlier it was not finding those users at all- who changed the username, 80% itself is a big amount.

@pagokrat0r 
You have really not applied the content of this PR properly.
If you did, you wouldn't get those timeout messages because I HAVE disabled that message for that part.
See https://github.com/timgrossmann/InstaPy/pull/2978/commits/ac6e2b488247544fc1e029e204eca736854f5b46.

----

Cheers 🙂
----

uluQulu:
#### Notable changes:
New track in `explicit_wait()` function: wait until "page fully loaded"
```python
condition = (lambda browser: browser.execute_script("return document.readyState")
                                        in ["complete" or "loaded"])
```
I wrote it for the `is_page_available()` definition operate properly cos since it is checking the page's title for some keywords, there can be a misbehavior as,
Page has not yet loaded and it returns the title of the older page which is an unwanted behavior.
Now with that new explicit wait method, it will first wait for the page fully load and the check the title of page in 100% accuracy.
How fast does it take page fully load?
In my tests, it takes, 
```assembly
page load test:

page loaded in 0.05199766159057617 seconds
page loaded in 0.01803112030029297 seconds
page loaded in 0.018996715545654297 seconds
page loaded in 0.016997098922729492 seconds
page loaded in 0.020003557205200195 seconds
page loaded in 0.9089870452880859 seconds
page loaded in 0.018996715545654297 seconds
page loaded in 0.050997018814086914 seconds
```
_In future that explicit wait method can be applied to other blocks which also need the page load fully_.

-----

Also other improvements in `is_page_available()` definition is adding 2 more blocks of getting the page title:
`browser.execute_script("return document.title")`
`browser.execute_script("return document.getElementsByTagName('title')[0].text")`

Talking a bit about its usage,
Earlier it was checking the emergency situations- check connectivity with pinging + check authorization if logged in.
But now, it first checks for the page being valid by `is_page_available()` function.
And most of times it is that situation of page being invalid- ["Page Not Found", "Content Unavailable"] in page title AND in case of page is still available, then it will look for emergency situations.

-----

Cheers 🧐
----

sionking:
> Your suggestion is poor to be a solution so far and I have stated it 2 times before and is prone to cause misbehaviours out of inaccuracy.
> example

This situation will fail 0.001% while you suggest a 80% accuracy (not thinking on users that only follow private 0%). I also never experienced fail check for username change name in search box (that is the point of this search right ?)

Haven't you learn not you use things that are not normal use (the logged out issue)?
I only use **fake bots with proxy** to operate **non-human actions** so I really care only about the users here and their account being flagged, I won't anyway.

uluQulu:
>not thinking on users that only follow private 0%

There are only very rare people [possibly] who target only [or mostly] private profiles and that is crazy 😄

>I also never experienced fail check for username change name in search box

You have not explained it. I had a quite old profile and there is a user and I am following that user at the moment. But when I search in the search box it does not pop up in the first place, not even first place, it does not even appear there. I go to that user's profile manually and it is looking quite normal and has not blocked my profile. Now take this example on the fuzziness of the search algorithm.

>,,,that is the point of this search right ?

The point of the search feature is to find people [etc.]. Cos [casual] user does not do navigate [sometimes does] to other profile by `.com/username`, they just run a search.
But it's point has never been finding the user who has changed the username.
There are too many scenarios. If you want to hear I will keep telling 😛
_Also you have not explained my example scenario with user A and user B_.

>Haven't you learn not you use things that are not normal use (the logged out issue)?

I think you have not read the main description of the PR.
There I have explained why I have included that as 3rd step:
"_If you disagree, we can remove it. But note that this 3rd step is gonna be requested only in rare cases cos almost everytime prior 2 steps will get the username_"

>I only use fake bots with proxy to operate non-human actions so I really care only about the users here and their account being flagged, I won't anyway.

Every people isn't eligible to get a [GOOD & SECURE & RELIABLE] proxy service as you might do.
Also, I do not think InstaPy is a fake bot, it is a quite nice automation service.
IF you think one more time you will see that your users' accounts can already be flagged in your case. I will give one example:
+ direct unrelated navigations by `GET` method

_It's up to IG to allow people do it and noone knows how far it will go like that and if you do not exceed the limits, everything should be fine_.

>This situation will fail 0.001% while you suggest a 80% accuracy

Which situation?

-----

Cheers 😀
-----

sionking:
> Which situation?

image
This situation is as rare as ...


Fake bot == I mean I have a server with fake bots that do all the data processing and if I am going to navigate to a page that brings the user_id I will use a FAKE BOT and not real account.


> * direct unrelated navigations by `GET` method

What is wrong with that ? referral page is still the page you was last, so if you go back and forward to, for example, profile page you take all the followers from, it seems correct in my point of view.

uluQulu:
@sionking 
>This situation is as rare as ...

It is not rare, this tool is generating a massive data and it will encounter lots of misbehaviours in that case.
Yet, search box does not properly work for my profiles to find normal people.
It has so many leaks,
User A was following me, now does not or I do not follow. So search will most likely not find it, etc.

Actually once again I have run it in **full test mode** where it skips every other user if username was not changed and find the new usernames for the ones who changed.
##### It found 3 users and the result is the search box could not find 2 users with old usernames.

#### Now I don't know on what you relied on saying: "_I also never experienced fail check for username change name in search box_" even in the most normal mode- username changed and old username is on the fly.  
And those new usernames [with the same user ID] are fully functional and is accessible that I visited.
It was an expensive test and I tested it on my main profile and it visited the profile page of 116 unique users to just find 3 usernames changed and run it without a sleep delay (_hope did not burn server requests_ 🐬)

>[Fake bot]

Note that, thousands of InstaPy users does not use extra proxy to get data as you do.
I think if one day that capability is added, then all the graphql queries might be moved to there.

>What is wrong with that ? referral page is still the page you was last, so if you go back and forward to, for example, profile page you take all the followers from, it seems correct in my point of view.

I do not want to stay on this topic, but think about this:
You are at the profile page of user Z WHERE there is no any either reference or notion of user X.
And then you navigate to the profile page of user X with the `GET` navigation method == easily noticeable. And you do it many many times..

----

Cheers 😎
------

sionking:
> I do not want to stay on this topic, but think about this:
> You are at the profile page of user Z WHERE there is no any either reference or notion of user X.
> And then you navigate to the profile page of user X with the `GET` navigation method == easily noticeable. And you do it many many times..

I don't do it ... I always go to the referral page before going to the user X.

You can't know what IG flag you since they know you are a bot.
1. no mouse movements
2. **security** brick here is way worse than you think (proxy is detectable !!! if you don't hide it in chrome)

instapy/unfollow_util.py
```diff
@@ -96,21 +98,26 @@ def set_automated_followed_pool(username, unfollow_after, logger, logfolder):
 
 
 
-def get_following_status(browser, person, logger):
+def get_following_status(browser, track, person, logger):
     """ Verify if you are following the user in the loaded page """
+    if track == "profile":
```
sionking, line 103: Why do we need it ?

instapy/util.py
```diff
@@ -1120,3 +1146,102 @@ def get_current_url(browser):
...
+    """ Check if the page is available and valid """
+    # wait for the current page fully load
+    explicit_wait(browser, "PFL", [], logger, 10)
+
+    try:
+        page_title = browser.title
+
+    except WebDriverException:
+        try:
+            page_title = browser.execute_script("return document.title")
```
sionking, line 1223: Selenium has a built in method for that

instapy/util.py
```diff
@@ -1120,3 +1146,102 @@ def get_current_url(browser):
...
+        except WebDriverException:
+            try:
+                page_title = browser.execute_script(
+                    "return document.getElementsByTagName('title')[0].text")
+
+            except WebDriverException:
+                logger.info("Unable to find the title of the page :(")
+                return True
+
+    expected_keywords = ["Page Not Found", "Content Unavailable"]
```
sionking, line 1234: please add:
"Facebook | Error"
Page url as title (ERR in connection)

sionking:
I added some notes

sionking:
Good JOB BTW !

instapy/unfollow_util.py
uluQulu: While getting the username from the user ID by loading a post, it stays at the post page and returns back the new username (_if any_).
And after that step `get_following_status()` might run again.
For that reason I added that navigation.
But actually getting the following status from post page is also possible but getting it from the profile page is much better.

I have also thought of every other occasion and there is no any problems.

instapy/util.py
uluQulu: Yes, the first block is a selenium way- `browser.title`.
But in case of a failure from that way, there are 2 JS alternatives to be used cos for this part of code, getting the title of page is highly important.

pagokrat0r:
@uluQulu After applying the PR properly, i'm getting this error
```

INFO [2018-09-24 02:16:05] [********]  Ongoing Unfollow [1/800]: now unfollowing 'b'travelreadyco''...
ERROR [2018-09-24 02:16:07] [********]  Unfollow loop error:  get_following_status() missing 1 required positional argument: 'logger'

INFO [2018-09-24 02:16:07] [********]  --> Total people unfollowed : 0

```

spikeyhair:
After running for 12 hours, hit this error:

  File "D:\Dropbox\Instapy\instapy\util.py", line 521, in click_element
    element.click()
  File "C:\Program Files (x86)\Python\lib\site-packages\selenium\webdriver\remot
e\webelement.py", line 80, in click
    self._execute(Command.CLICK_ELEMENT)
  File "C:\Program Files (x86)\Python\lib\site-packages\selenium\webdriver\remot
e\webelement.py", line 628, in _execute
    return self._parent.execute(command, params)
  File "C:\Program Files (x86)\Python\lib\site-packages\selenium\webdriver\remot
e\webdriver.py", line 312, in execute
  File "C:\Program Files (x86)\Python\lib\site-packages\selenium\webdriver\remot
e\errorhandler.py", line 237, in check_response
selenium.common.exceptions.StaleElementReferenceException: Message: stale elemen
t reference: element is not attached to the page document
  (Session info: chrome=69.0.3497.100)
  (Driver info: chromedriver=2.40.565498 (ea082db3280dd6843ebfb08a625e3eb905c4f5
ab),platform=Windows NT 6.1.7601 SP1 x86_64)


During handling of the above exception, another exception occurred:

  File "start_instapy.py", line 249, in <module>
    session.follow_commenters(follow_username[users-1], amount=follow_commenters
_amount, daysold=random.randint(follow_commenters_daysold_min,follow_commenters_
daysold_max), max_pic=random.randint(follow_commenters_pic_min,follow_commenters
_pic_max), sleep_delay=random.randint(sleep_min,sleep_max), interact=True)
  File "D:\Dropbox\Instapy\instapy\instapy.py", line 642, in follow_commenters
    commenters = extract_information(self.browser, username, daysold, max_pic)
  File "D:\Dropbox\Instapy\instapy\commenters_util.py", line 217, in extract_inf
ormation
    user_commented_list, pic_date_time = extract_post_info(browser)
  File "D:\Dropbox\Instapy\instapy\commenters_util.py", line 54, in extract_post
_info
    click_element(browser, load_more_comments_element)
  File "D:\Dropbox\Instapy\instapy\util.py", line 532, in click_element
    browser.execute_script("document.getElementsByClassName('" +  element.get_at
tribute("class") + "')[0].scrollIntoView({ inline: 'center' });")
  File "C:\Program Files (x86)\Python\lib\site-packages\selenium\webdriver\remot
e\webelement.py", line 143, in get_attribute
    resp = self._execute(Command.GET_ELEMENT_ATTRIBUTE, {'name': name})
  File "C:\Program Files (x86)\Python\lib\site-packages\selenium\webdriver\remot
e\webelement.py", line 628, in _execute
    return self._parent.execute(command, params)
  File "C:\Program Files (x86)\Python\lib\site-packages\selenium\webdriver\remot
e\webdriver.py", line 312, in execute
  File "C:\Program Files (x86)\Python\lib\site-packages\selenium\webdriver\remot
e\errorhandler.py", line 237, in check_response
selenium.common.exceptions.StaleElementReferenceException: Message: stale elemen
t reference: element is not attached to the page document
  (Session info: chrome=69.0.3497.100)
  (Driver info: chromedriver=2.40.565498 (ea082db3280dd6843ebfb08a625e3eb905c4f5
ab),platform=Windows NT 6.1.7601 SP1 x86_64)

uluQulu:
#### Hi all

Thanks, @sionking 👋🏼

-------

@pagokrat0r 
You haven't applied it properly.
Make sure all of the `get_following_status()` function calls have 4 parameters passed in **unfollow_util.py** file. (_I added one more parameter- `track` with this PR_)

_If you find hard to apply it, then download a zip of my fork and put into e.g. `InstaPy_test` folder and bring your **quickstart** script to there and run & test & observe it properly_.

----

@spikeyhair 
I don't even ask if you have applied this PR properly cos it **seems to be** a different issue to be raised in another thread.
Just mention my username to debug it together ⚽

----

Cheers 😂
----

timgrossmann:
Did someone apply this PR and test it?

sionking:
Looks OK (code review not tested).

spikeyhair:
Sorry, I might not be familiar with the process. After committed into master means will be available to the pack under https://github.com/timgrossmann/InstaPy.git ?

uluQulu:
Yes @spikeyhair. Now you can pull 💪🏼 it

sionking:
i WILL TEST IT NOW

sionking:
@uluQulu 
THERE IS something in the logic.
in unfollow_user function you should check first for is_page_available and only than get_following_status.

uluQulu:
@sionking 
>in unfollow_user function you should check first for is_page_available and only than get_following_status.
  
I have thought about every possible occasion.
**1**. `get_following_status()` is super flexible to operate in invalid pages WHERE will return `None`  
**2**. I specifically chose that style cos of, having clearer design in case of user has changed a username:
`if not valid_page or sirens_wailing == False:`  <- in both conditions username might be changed  
**3**. The only fallback is the time delay from `explicit_wait()` to wait until expected element in an invalid page (it will fully wait).
But since that scenario is ~5% it is not a problem.

----

Cheers 🕵🏼‍♂️
----

sionking:
@uluQulu I am not fully understand Y not also checking this before get_following_status to save time.

you can have (ugly)
```
        # find out following status
        if not is_page_available:
            following = None
            follow_button = None
        else:
            following, follow_button = get_following_status(browser, person, logger)
```
is_page_available is very quick while get_following_status is 66 seconds!

uluQulu:
@sionking 
Right, I remember now.
Earlier once `following is None` then it directly goes and checks for emergency situation [connectivity & authorization]. Read that part of code, you'll understand.
That's also another reason of my approach.

Actually that kind of mind confusion came with trying to merge your function- `is_page_available()`.
It's so good, no question.

But it made that part of code look a lot unstructured.
And having considered the urgency of the fix to be offered to community, I just chose that suggested style for then.

I think it is time to rebase all of the ACTION VERIFICATION block into a separate function and make things look much simpler.
With that rebase I think we can do that change you mentioned- running `is_page_available()` before `get_following_status()`. I will look again and I think it is viable.

----

Cheers 😁
----

sionking:
BUG : sirens_wailing might be checked `sirens_wailing == False` before assigend
```
        if following is None:
            # check if the page is available
            valid_page = is_page_available(browser, logger)
            if valid_page:
                # check out if the loop has to be broken immidiately
                sirens_wailing, emergency_state = emergency_exit(browser, username, logger)
                if sirens_wailing == True:
                    return False, emergency_state

            if not valid_page or sirens_wailing == False:
```

sionking:
I tested this solution and also with is_page_available and it works ok.
I must say I have my own unfollow util, that kind of double check everything, but I will better align with this one... and let you know.

JoeFertig:
Tested everything and it works perfectly♥️
Thanks for the awesome work @uluQulu!!

PR #2985, opened 2018-09-22
comments mandatory words

sionking:
Validate post have at least one valid word from comments_mandatory_words list.
Do comment on it if does.

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -187,6 +187,7 @@ class InstaPy:
         self.commenting_approved = True
         self.max_comments = 35
         self.min_comments = 0
+        self.comments_mandentory_words = []
 
         self.relationship_data = {username:{"all_following":[], "all_followers":[]}}
 
@@ -978,12 +979,13 @@ class InstaPy:
     def set_delimit_commenting(self,
                                 enabled=False,
                                  max=None,
-                                  min=None):
+                                  min=None,
+                                    comments_mandentory_words=[]):
 
         self.delimit_commenting = True if enabled==True else False
         self.max_comments = max
         self.min_comments = min
-
+        self.comments_mandentory_words = [] # comment only when the image description contain at least one of those words
 
 
     def set_simulation(self, enabled=True, percentage=100):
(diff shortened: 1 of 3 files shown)


timgrossmann:
@sionking Where did that feature come from? 😅 
There is no documentation about `set_delimit_commenting` is that something that is only used internally? 
Looks like some feature that enables users to skip posts with more/less than *n* comments...

Thanks for the mandatory_word addition!

instapy/instapy.py
```diff
@@ -978,12 +979,13 @@ def set_delimit_liking(self,
...
                                  max=None,
-                                  min=None):
+                                  min=None,
+                                    comments_mandatory_words=[]):
 
         self.delimit_commenting = True if enabled==True else False
         self.max_comments = max
         self.min_comments = min
-
+        self.comments_mandatory_words = [] # comment only when the image description contain at least one of those words
```
timgrossmann: Shouldn't you assign `self.comments_mandatory_words = comments_mandatory_words`, though? 🤔

sionking: might be merge issue

timgrossmann:
@sionking Is this a yes for the question about the documentation?
Could you add some documentation to the README about that feature including the mandator words addition, please? 🙈

sionking:
> @sionking Is this a yes for the question about the documentation?
> Could you add some documentation to the README about that feature including the mandator words addition, please? 🙈

Hi ! I really don't know how to work with the documents here, what should I do ?

instapy/instapy.py
sionking: fixed

timgrossmann:
@sionking Basically the README is the only place to work for documentation, the syntax is Markdown.

If you could quickly write down what the feature does and what your addition does, I can integrate it into the documentation in a separate PR.

sionking:
This is a very simple feature before commenting we would compare description of post (and the first comment) to comments_mandatory_words list.
if any word from the list is in the post than we pass the verification.
It is good when you want to comment only on specific tags (which are in the post desc) and so ...

timgrossmann:
@sionking Just found the documentation for the comment delimiting feature 👍 I'll add the mandatory words description 

Thank you

Issue #2592, opened 2018-08-01
instapy.instapy.InstaPyError: ensure chromedriver is installed


tibor:
I get the same error as in #2179, #2618.

## Current Behavior
```
ERROR [2018-08-01 21:11:58] [username]  Message: unknown error: Chrome failed to start: exited abnormally
  (unknown error: DevToolsActivePort file doesn't exist)
  (The process started from chrome location /usr/bin/google-chrome is no longer running, so ChromeDriver is assuming that Chrome has crashed.)
  (Driver info: chromedriver=2.41.578700 (2f1ed5f9343c13f73144538f15c00b370eda6706),platform=Linux 2.6.32-042stab120.6 x86_64)
Traceback (most recent call last):
  File "/home/user/InstaPy/instapy/instapy.py", line 302, in set_selenium_local_session
    chrome_options=chrome_options)
  File "/usr/lib/python2.7/site-packages/selenium/webdriver/chrome/webdriver.py", line 69, in __init__
    desired_capabilities=desired_capabilities)
  File "/usr/lib/python2.7/site-packages/selenium/webdriver/remote/webdriver.py", line 90, in __init__
    self.start_session(desired_capabilities, browser_profile)
  File "/usr/lib/python2.7/site-packages/selenium/webdriver/remote/webdriver.py", line 177, in start_session
    response = self.execute(Command.NEW_SESSION, capabilities)
  File "/usr/lib/python2.7/site-packages/selenium/webdriver/remote/webdriver.py", line 236, in execute
    self.error_handler.check_response(response)
  File "/usr/lib/python2.7/site-packages/selenium/webdriver/remote/errorhandler.py", line 192, in check_response
    raise exception_class(message, screen, stacktrace)
WebDriverException: Message: unknown error: Chrome failed to start: exited abnormally
  (unknown error: DevToolsActivePort file doesn't exist)
  (The process started from chrome location /usr/bin/google-chrome is no longer running, so ChromeDriver is assuming that Chrome has crashed.)
  (Driver info: chromedriver=2.41.578700 (2f1ed5f9343c13f73144538f15c00b370eda6706),platform=Linux 2.6.32-042stab120.6 x86_64)

Traceback (most recent call last):
  File "username_de.py", line 11, in <module>
    InstaPy(username='username',password='Password', nogui=True)\
  File "/home/user/InstaPy/instapy/instapy.py", line 195, in __init__
    self.set_selenium_local_session()
[...]

PR #2991, opened 2018-09-23
Set headless_browser default value to True

tibor:
Fixes issue `instapy.instapy.InstaPyError: ensure chromedriver is installed`(#2592 and similar)

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -85,7 +85,7 @@ class InstaPy:
                  browser_profile_path=None,
                  page_delay=25,
                  show_logs=True,
-                 headless_browser=False,
+                 headless_browser=True,
                  proxy_address=None,
                  proxy_chrome_extension=None,
                  proxy_port=None,


sionking:
Y?

ghost:
yeah this doesnt make sense to do.

timgrossmann:
@tibor @sionking @halalfood @uluQulu  This is a really tough one I'm thinking about for some time already, that error is not new to me...

There are basically two situations.
1. People want to use it on ubuntu/a server and will encounter this, creating dozens of issues because it does not work and they do not read the documentation. But new users are able to see the bot work when trying it out.

2. We change it to the, currently proposed, version which also works on any other system. We'd loose the "haptic" for new users who simply want to try it out on their local machine and might get confused when they only see the terminal output.

I think, since the ubuntu/server thing is more advanced, people should "simply" know that they have to adjust that (read the documentation) when they want to use it on a server.

sionking:
But no real fix for chrome and proxy with auth

PR #3001, opened 2018-09-24
Docu update with chromedriver/db relocation

timgrossmann:
Also changing some points in the quickstart.py

Any ideas on how we could remove the exception part out of the quickstart file?

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ Table of Contents
   * [Simulation](#simulation)
   * [Disable Image Loading](#disable-image-loading)
   * [Using Multiple Chromedrivers](#using-multiple-chromedrivers)
+  * [Changing DB or Chromedriver locations](#changing-db-or-chromedriver-locations)
 
 ## Getting started
 
@@ -1530,6 +1531,14 @@ If you need multiple os versions of chromedriver just rename it like:
 chromedriver_linux
 chromedriver_osx
 chromedriver_windows
+```
+
+### Changing DB or Chromedriver locations
+If you want to change the location/path of either the DB or the chromedriver, simply head into the `instapy/settings.py` file and change the following lines. 
+Set these in instapy/settings.py if you're locating the library in the /usr/lib/pythonX.X/ directory.
+```
+#   Settings.database_location = '/path/to/instapy.db'
+#   Settings.chromedriver_location = '/path/to/chromedriver'
 ```
 
 ---


uluQulu:
Hi @timgrossmann 

I also thought of it before and I would suggest to use a decorator there,
added to e.g. **util.py** file:
```python
from tempfile import gettempdir   # other imports already exist in 'util.py'

@contextmanager
def smart_run(session):
    try:
        session.login()
        yield

    except Exception as exc:
        # if changes to IG layout, upload the file to help us locate the change
        if isinstance(exc, NoSuchElementException):
            file_path = os.path.join(gettempdir(), '{}.html'.format(time.strftime('%Y%m%d-%H%M%S')))
            with open(file_path, 'wb') as fp:
                fp.write(session.browser.page_source.encode('utf8'))
            print('{0}\nIf raising an issue, please also upload the file located at:\n{1}\n{0}'.format(
                '*' * 70, file_path))
        # full stacktrace when raising Github issue
        raise

    finally:
        session.end()
```

#### And now, **quickstart.py** file could look like,
```python
# imports
from instapy import InstaPy
from  instapy.util import smart_run

# login credentials
insta_username = 'abc.az'
insta_password = '123_L'

# instantiation of the InstaPy class with [custom] arguments
session = InstaPy()

# as session is ready, let's go!
with smart_run(session):
    """ Activity flow here """    
    # SETTINGS
    session.set_relationship_bounds(enabled=True, potency_raito=1.1)
    
    # ACTIONS
    session.like_by_tags(['natgeo'], amount=1)
```

---

Cheers 😀
-----

timgrossmann:
@uluQulu I've never read about the `@contextmanager` decorator before but I already love it.
Still learning so much out of this project every day!

Thanks for that, love that approach.
If we need to add some more logic in the future, we can just add it and really nicely hide it from the user like that.

Do you think you could PR this? 🙈

> Shouldn't it be `yield session` though? 🤔 


Would also be awesome to have something like this:

```python
# imports
from instapy import InstaPy
from  instapy.util import get_session

# login credentials
insta_username = 'abc.az'
insta_password = '123_L'

# instantiation of the InstaPy class with [custom] arguments
instapy = InstaPy(username=insta_username,
                               password=insta_password)

# as session is ready, let's go!
with get_session(instapy) as session:
    """ Activity flow here """    
    # SETTINGS
    session.set_relationship_bounds(enabled=True, potency_raito=1.1)
    
    # ACTIONS
    session.like_by_tags(['natgeo'], amount=1)
```

sionking:
Why making things complex ? less and less contributes will understand the code.

timgrossmann:
@sionking I think it should get less complex for non technical people to use this tool in the future.
If at one point completely non-technical people can download InstaPy and basically just "double click" something to start a GUI, do their configuration in there and run InstaPy, it would be awesome.

Up to that point we should make the setup as easy as possible and slowly remove/hide the complexity from the users.
Something like the exception block scares non-technical people away. Having a clean quickstart.py which is super simple to understand and you basically only have to copy in all the commands is the "hardest" thing we should have until there is a real GUI that is based on the api.json file which then will get updated with each new feature.

You're definitely right about keeping the code as simple as possible, too.
But I think seeing some more advanced things like decorators and then searching google for what this does will teach a lot of people, if they are interested, about patterns like this.

uluQulu:
Thanks, @timgrossmann 
Context managers are so useful that's why it is added to the `contextlib` module [as a generator function- `contextmanager`] to ease the process ,

You can implement any context manager using classes [in long way],
**1**-) you must define an `__enter__` method that return an object
**2**-) and must define an `__exit__` method 
It is a minimal setup, e.g. `__init__` also can be included [optionally].
It's literally like "_Hey all the time do **this** before doing [some job] and just before leaving that job do **that**_"
And can be written like this,
```python
class smart_run:
    def __enter__(self):
        print("Hey! I have just logged in :)")
        return self

    def __exit__(self, *args):
        print("session has ended!")

with smart_run():
    try:    
        print("I am doing come like activity here!")
    except Exception as exc:
        raise
```
As you can see, the `contextmanager` generator replaces only the `try`/`finally` block of this handwritten alternative above.
And the `except` block I have included in the `smart_run()` function is catching **BOTH** exceptions of the activity flow AND inside the `smart_run()` function.

@timgrossmann, the session object is static and doesn't need to be generated,
But if you are to use it as,
`with smart_run(instapy) as session`
then `yield session` is a requirement.



Also from your suggestion,
**1**-)
"instapy = InstaPy()"
new user thinks: "Wow! lots of **instapy**s 😟"

It could be e.g.
"my_instapy = InstaPy()" 😛

**2**-)
`get_session` sounds like it is getting a session, but `session=InstaPy()` line gets the session.
_"Session started!" messaged is being printed in the `__init__` method of InstaPy_. [just after `session=InstaPy()` line]

Actually the `smart_run()` just **logs in** and _coordinately_ runs the session.


@sionking
It's not complex, rather easier.
Remember very frequently used `open()` function:
+ Using `with open()` [by context manager] which auto closes the file in case of you forget,
```python
with open('file_location', 'r') as my_file:   # 'my_file' is gonna be closed as code execution leaves the `with` block
    lines = my_file.readlines()
```
+ But if you just used `open()` [without `with` = no context manager] and forgot to close the file, then it could lead to many issues.
```python
my_file = open('file_location', 'r')
lines = my_file.readlines()
#my_file.close()   # forgot to close file :(   OR if not forgot, it is one more extra line
```
But your point is also understandable, however, it requires some knowledge about _generators_, `yield` and _decorators_.

@timgrossmann, to generalize,
in both variations,
`with smart_run(session)`  [having `yield` or `yield None`]
and
`with smart_run(my_instapy) as session` [having `yield session`]
`session` is the same object.

I mean since there is no any change in those variations, your suggestion might add a little bit confusion for people (_programmers_) to understand.

@timgrossmann,
Sure I can PR it.
Do you think I would add it on my live PR?

-----

Cheers 😁
-----

timgrossmann:
@uluQulu Thank you very much, I thinks it's better to create a separate one for this 😉 
I'll change this PR to only make changes to the README, and we'll merge yours with the context manager once it's PRed 👍

PR #3037, opened 2018-09-28
Updated recommendations of versions of chromium and chromedriver

meinert:


timgrossmann:
@meinert Awesome, thank you very much for the documentation update!

PR #3078, opened 2018-10-04
Allow skipping private users when grabbing following list

photogsy:
I've found it really useful to be able to grab a list of the people I'm following that are _not_ private. I can then use this list as a starting point which I can filter further, then use for unfollowing.

Very small changes but a really useful (IMHO) feature.
I've been using this for a week or so with expected behaviour.

### DIFF
--- a/instapy/relationship_tools.py
+++ b/instapy/relationship_tools.py
@@ -226,7 +226,8 @@ def get_following(browser,
                       live_match,
                        store_locally,
                         logger,
-                         logfolder):
+                         logfolder,
+                          skip_private):
     """ Get entire list of following using graphql queries. """
     if username not in relationship_data:
         relationship_data.update({username: {"all_following":[], "all_followers":[]}})
@@ -258,7 +259,7 @@ def get_following(browser,
 
     graphql_endpoint = 'https://www.instagram.com/graphql/query/'
     graphql_following = (
-        graphql_endpoint + '?query_hash=58712303d941c6855d4e888c5f0cd22f')
+        graphql_endpoint + '?query_hash=c56ee0ae1f89cdbd1c89e2bc6b8f3d18')
 
     all_following = []
 
@@ -321,7 +322,8 @@ def get_following(browser,
                 data['user']['edge_follow']['page_info'])
             edges = data['user']['edge_follow']['edges']
             for user in edges:
-                all_following.append(user['node']['username'])
+                if (skip_private == False or user['node']['is_private'] == False):
+                    all_following.append(user['node']['username'])
 
             grabbed = len(set(all_following))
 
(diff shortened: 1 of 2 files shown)


sionking:
I like private users :)

timgrossmann:
@photogsy Could you elaborate a little bit more on why this is useful?
What is the reason you don't want to unfollow private users?

Thank you very much for PRing your feature!

photogsy:
Basically its useful to be able to grab a user's following list but ignore private users since you don't want to worry about how many requests to follow you are sending to private accounts.

The problem before was that if you ask InstaPy to grab 500 of a users following list but only want to interact with public accounts, then you end up with far fewer useful users in your list than 500 since you are then going to skip the private ones later.

Hope that makes sense!

I am going to PR a similar change for grabbing a user's followers.

timgrossmann:
@photogsy What is the reason this is called `ignore_private` at one place and `skip_private` at the other?

Also we would need some documentation before merging this. Do you think you could add a quick section about it in the README?

timgrossmann:
@photogsy Any updates on this?

timgrossmann:
@photogsy Any updates on this? 😊

photogsy:
Sorry I haven't had time.
There is no reason for the parameter naming difference, my Python knowledge is limited so I wasn't sure about scope, so to avoid possible issues I named them differently.

timgrossmann:
@photogsy No worries, thank you for the response.

I think it's a good idea to allow something like this, we just need to make sure that the logic is right.
Also, the naming should be unified so it should be called `skip_private` everywhere. 

Thank you very much for proposing this

instapy/relationship_tools.py
```diff
@@ -258,7 +259,7 @@ def get_following(browser,
 
     graphql_endpoint = 'https://www.instagram.com/graphql/query/'
     graphql_following = (
-        graphql_endpoint + '?query_hash=58712303d941c6855d4e888c5f0cd22f')
+        graphql_endpoint + '?query_hash=c56ee0ae1f89cdbd1c89e2bc6b8f3d18')
```
timgrossmann, line 262: Double check this GraphQL query hash

instapy/relationship_tools.py
```diff
@@ -321,7 +322,8 @@ def get_following(browser,
                 data['user']['edge_follow']['page_info'])
             edges = data['user']['edge_follow']['edges']
             for user in edges:
-                all_following.append(user['node']['username'])
+                if (skip_private == False or user['node']['is_private'] == False):
```
timgrossmann, line 325: boolean checking in python should be done with `is True/False` instead of `== True/False`

timgrossmann (changes requested):

timgrossmann:
@photogsy Sorry to see that we could not get this features merged.
If at one point you can update this to the latest state of InstaPy and apply the adjustments, we can definitely merge this. I'm sure a lot of users will find this useful.

Thank you very much

PR #3088, opened 2018-10-06
Feature/track liked accounts

packetfocus:
Tested, most everything works. 

only the hourly status .py file has been converted to use ENV and is included.
the like_tag? is only updated with like by tag and interact with username. Will fail on others because they all need the liked_method added.

Updated the blacklist to work, you pass it MaxLikes in the config. 

The campaign.txt log file is now created and tracks the campaign better, showing what was done and the username.

Added randomness into all py files that call likes etc.


packetfocus:
Didn't mean to create PR.

Issue #2824, opened 2018-09-01
 blacklist dictWrite error 'encoding' is an invalid keyword argument for this function


redsector72:
I get this error every action.
I dont' really need the blacklist, but it is related to the campaign.

sionking:
maybe related to latest update

sionking:
in add_user_to_blacklist
change to this:
        with open('{}blacklist.csv'.format(logfolder), 'a+') as blacklist:
from this:
        with open('{}blacklist.csv'.format(logfolder), 'a+', encoding="utf-8") as blacklist:

I don't know why you getting this error due.

redsector72:
HI @sionking is already like this (an update?)
But I still get the same message.
I have even deleted the blacklist.csv file and seems like it is not getting generate/written anymore.
I'm on a mac.

the full log line is
ERROR [2018-09-02 08:20:42] [Username]  blacklist dictWrite error 'encoding' is an invalid keyword argument for this function

redsector72:
No news?

jeremycjang:
@sionking  Encountering this issue as well and trying to read up more on the ```Dictwrite``` function. At first I thought it might be because the utf-8 string uses double ticks instead of single, but that didn't seem to fix it

https://docs.python.org/3.1/library/csv.html
>import csv
with open('some.csv', newline='', encoding='utf-8') as f:
    reader = csv.reader(f)
    for row in reader:
        print(row)

jeremycjang:
@redsector72  In the interim, I've confirmed that you can fix it by removing ```encoding="utf-8"``` so your ``` add_user_to_blacklist'``` code should look like this:

```
 open('{}blacklist.csv'.format(logfolder), 'a+'
```
Hope that helps!

timgrossmann:
#3107

PR #3107, opened 2018-10-09
Blacklist is still not fixed

redsector72:
Referring to issue #2824, this fix works on my system (a mac).
I suppose is ok also for everyone else.
If not, try to find the best fix as, as is, will not write the file.


timgrossmann:
@redsector72 I'm not sure if simply removing the encoding really fixes this issue...
Could you explain why you removed that?

timgrossmann:
@redsector72 Are you using InstaPy with python2? It looks like this is a python2 problem.

timgrossmann:
@redsector72 Seems like this really does fix it.
Thank you very much!

PR #3123, opened 2018-10-11
Updates for Working with Clarifai API Image 

dfundingsland:
I made several updates to both in `instapy.py` and `clarifai_util.py` to resolve some issues caused when using the Clarifai API.

In `instapy.py`, I made the following updates:

To reduce the amount of duplicate code used (and chance for errors to arise), I added the instance method `query_clarifai` to call the `check_image` method. Updated `clarifai_check_img_for` so that if no value was provided for a `tags_skip` it would default to an empty list (i.e. `[]`).

In `clarifai_util.py` the following updates were made:

Removed import of `Image` from `clarifai.rest` and use `predict_by_url` instead of `predict`. 

Added a check to see if the `nsfw` tag is used. If detected Clarifai’s nsfw model will be used for prediction. At present, only two models, nsfw and general, are being used. This is a feature which could be expanded on in the future to include additional models that Clarifai provides. 

Changed the xpath selector to `//img[@class = "FFVAD" or @class="_8jZFn”]` to ensure that the function also checks for keyframe images for video posts. 

Added `get_clarifai_tags` function to filter the results from returned from Clarifia. At present, the confidence level is set to log concepts with 50% or higher confidence. If the results are not filtered, any request to nsfw model, for example, will return both nsfw and sfw concepts. The end result being that every image queried would be flagged as nsfw.

Current implementation not perfect as the nsfw test can still be fooled if the keyframe image for a posted video is considered sfw. Also, 50% confidence threshold may still allow some suggestive content to get through.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -107,3 +107,7 @@ logs/
 # pytest
 .pytest_cache/
 tests/logs
+
+# Pipfile
+Pipfile
+Pipfile.lock
--- a/README.md
+++ b/README.md
@@ -1387,9 +1387,9 @@ session.end()
 ### Enabling Imagechecking
 
 ```python
-# default enabled=False , enables the checking with the clarifai api (image
+# default enabled=False , enables the checking with the Clarifai API (image
 # tagging) if secret and proj_id are not set, it will get the environment
-# variables 'CLARIFAI_API_KEY'
+# variables 'CLARIFAI_API_KEY'.
 
 session.set_use_clarifai(enabled=True, api_key='xxx')
 ```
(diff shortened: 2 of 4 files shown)


sionking:
finally things getting nicer here :)

timgrossmann:
@dfundingsland Totally agree with @sionking, this is awesome!
Thank you very much.

Could we get some more approvals for this, please?
Working on my side, @sionking have you tested this?

Edit: Sorry, misclicked...

dfundingsland:
@timgrossmann and @sionking Thanks for kind words!

In addition to this, I'm looking into adding support for additional Clarifai models as well as support for checking an image against more than one model. Will share on separate issue when I have something working for that as well.

timgrossmann:
@dfundingsland Sounds incredible, really looking forward to this!
Thank you for being part of InstaPy

sionking:
i think we should use 0.92% and not 0.5

dfundingsland:
> i think we should use 0.92% and not 0.5

The way it currently works, setting the threshold to 0.92% would set the bar for what is considered `nsfw` much higher and allow many more suggestive images to get through. If you want to err on the side of caution, something lower like 0.15% would flag more images as 'nsfw'.

dfundingsland:
@sionking and @timgrossmann I've added a parameter to allow users to set probability (have pushed out the latest commit). I still have the default set to 50%, but that's an easy change. Let me know what you think.

timgrossmann:
@dfundingsland Thank you very much!

sionking:
hi what will //img[@class = "FFVAD" or @class="_8jZFn”]
return for video ? isn't it better to ignore videos (as the main content is not in the first frame)

dfundingsland:
@sionking Yes, it would be better to analyze the video rather than just the keyframe. Checking the key image is somewhat of an improvement though. Prior to adding `@class="_8jZFn"`, if a post was a video post and not an image post, the user would get a `no such element` error and neither the image nor the video would be checked. 

I'm happy to look into implementing a video check (unless someone beats me to it).

PR #3134, opened 2018-10-12
restyling comments

sionking:
only cosmetics work.
But more important this will modulate the comments Technic so each user can create his uniqe coding without modifying the all code.

Enjoy.


timgrossmann:
@sionking Had this running for a little while now, works fine.

thank you very much for this!

PR #3140, opened 2018-10-14
FIX - issue with multiple instantiation of instapy in same script

YannikFirre:
Hello,

Little fix who fix the logger when multiple instantiation of Instapy in the same file. 

Before, the logger keeped the configuration of the first account logged.


timgrossmann:
@InfinityWebMe Good catch!

Thank you

PR #3142, opened 2018-10-15
Added comment by tags functionality

SkyTreasure:


timgrossmann:
@SkyTreasure Could you talk about why we need this feature and how you use it, please?

sionking:
please not another function to maintain !!!
We shall use the same function like_by tag which need to rename to interact_by_tag and get the like\comment\follow option.

sionking:
one more copy-pate and we are doomed

timgrossmann:
@sionking Absolutely on your side. Just curious on why it's necessary to have this feature.

SkyTreasure:
@timgrossmann I had a use case where i needed comment on pics with specific hashtags on satisfying certain conditions, so i had made that change and was using it. Thought if someone else might also needed that functionality. If its not useful for you guys then you can ignore this.

timgrossmann:
@SkyTreasure Good to know and hear about your usage.
Thank you very much for creating a PR about it. It's really important that people are doing PRs with their changes.

What's the reason you need to avoid liking the picture?

SkyTreasure:
@timgrossmann Sorry, avoid liking the picture? I didn't get it. I have added code to like the pic and wait for some random seconds and comment later.

timgrossmann:
@SkyTreasure Thank you very much for clarifying!
I'm just referring to the fact that you could use `like_by_tag` with `set_do_comment` set to 100% to do exactly that. 
The only thing you can't do is **not** like and still comment.

For me it sounds like this would lead to the same result:
```python
session.set_do_comment(True, percentage=100)
.
.
.

session.like_by_tags([...], amount=10)
```

Could you please approve whether this is the thing you wanted to achieve with this addition?

SkyTreasure:
@timgrossmann oh yea this should work, in my case since i had certain conditions where i used to do some NLP over the caption and then decide whether to comment or not, I didn't want to change the set_do_comment(). 

Anyway, your above-mentioned code works fine. Thank you.

timgrossmann:
@SkyTreasure Perfect, this is really good to know, thank you very much for clarifying.

Still, thank you very much for creating a PR for this feature, maybe someone will stumble across it and use it as you did.

Have a good one
Best
Tim

PR #3145, opened 2018-10-15
applying @sionkings proposal about getting post count

timgrossmann:
Thanks @anna-osipova for pointing this out
@sionking For the more generally applicable fix

### DIFF
--- a/instapy/util.py
+++ b/instapy/util.py
@@ -686,10 +686,23 @@ def username_url_to_username(username_url):
 
 def get_number_of_posts(browser):
     """Get the number of posts from the profile screen"""
-    num_of_posts_txt = browser.find_element_by_xpath("//section/main/div/header/section/ul/li[1]/span/span").text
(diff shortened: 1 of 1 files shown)


sionking:
verified

timgrossmann:
@sionking I just saw that the 
```python
num_of_posts_txt = num_of_posts_txt.replace(" ", "")
num_of_posts_txt = num_of_posts_txt.replace(",", "")
num_of_posts = int(num_of_posts_txt)
```

Is only inside the second except block, shouldn't this be on the root level like the return statement?

sionking:
@timgrossmann  no the graphql return int.

timgrossmann:
@sionking Thank you for clarifying, the graphql in fact returns an int.

Also, I just realised that the indentation of the conversion is one less than the try/catch block which means that it will be executed after either the try or catch block have been completed.

Thank you @sionking

PR #3147, opened 2018-10-16
Clarifai Models and Workflows

dfundingsland:
Added support for the majority of Clarifai’s public models. Given that the API response for `Focus`, `Face Detection`, `Face Embedding`, and `General Embedding` is different from other models, I did not include them at this time. 

I have also added support for using Clarifai **workflows** (a single API call for checking an image against up to 5 models). With the changes, a user can choose which specific model (or models) they wish to check an image against. If a model isn’t chosen, the general model is used. It’s worth noting that you can either chose to check an image with a workflow or with one or more models.  

In the interest of simplicity,  I made the decision that `full_match=False` will always be used when checking tags against `img_tags_skip_if_contain`. That is, if any tag in `img_tags_skip_if_contain` is matched, InstaPy will not comment. I feel that this the safest course of action, but happy to discuss. `full_match` for other tags will still work as before.

If anything is unclear from the code or if any issues emerge from your tests, do let me know. I still haven't addressed the issue of checking video, but I agree with @sionking that this would be an improvement to just checking the keyframe image of a video.

For your reference, I've included some notes on potential use cases below (mostly just taken from the updated README file).

### Simple Queries
**Note**: If not specified by setting the `models=[‘model_name’] in `session.set_use_clarifai`, `models` will be set to `general` by default.

If you wanted to comment ‘Beautiful!’ on images with the concept `landscape` but exclude those that also contain the concept `water`, you would use the following settings:
```python
session.set_use_clarifai(enabled=True, api_key=‘xxx’)
session.clarifai_check_img_for([‘landscape’], [‘water’], comment=True, comments=[‘Beautiful!’])
```

If you wanted to comment ‘Beautiful!’ only on images which contain both `landscape` and `no person` as concepts but exclude those that also contain the concept `water` or `tree`, you would use the following settings:
```python
session.set_use_clarifai(enabled=True, api_key=‘xxx’, full_match=True)
session.clarifai_check_img_for([‘landscape’, ‘no person’], [‘water’, ‘tree’], comment=True, comments=[‘Beautiful!’])
```

As before, if you did not want to comment on images which contain the concept of `landscape`, you would use the following settings:
```python
session.set_use_clarifai(enabled=True, api_key=‘xxx’)
session.clarifai_check_img_for([‘landscape’])
```

To check on images which contain the concept `nsfw` by querying the Clarifai NSFW model, you would use the following settings:
```python
session.set_use_clarifai(enabled=True, api_key=‘xxx’, models=[‘nsfw’])
session.clarifai_check_img_for([‘nsfw’])
```

#### Querying Multiple Models with Workflow (Single API Call)
You can query multiple Clarifai models with a single API call by setting up a custom workflow.  Using a `workflow` is the recommended way to query multiple models. Alternatively, it is possible to query multiple models separately (see Querying Multiple Models (Multiple API Calls) below).

To setup a workflow, see the [Workflow Documentation](https://www.clarifai.com/developer/guide/workflow#workflow).

**NOTE** :As mentioned above, the `Focus`, `Face Detection`, `Face Embedding`, and `General Embedding` models are not current supported.

Once you have a workflow setup, you can use InstaPy to check images with the Clarifai Image API by setting the `workflow` parameter in `session.set_use_clarifai` to the name of your custom workflow.

Let's say you want to comment 'Great shot!' on images of men or women with the hashtag `#selfie`, but you want to make sure not to comment on images which might contain inappropriate content. To get general concepts, e.g. `woman`, you would setup your workflow using `General` and to check the image for the concepts `nsfw` and `explicit` you would also want to add NSFW and Moderation models to your workflow.

For example:
```python
session.set_use_clarifai(enabled=True, api_key='xxx', workflow=['your-workflow'])
session.clarifai_check_img_for(['woman', 'man'], ['nsfw', 'explicit', 'suggestive'], comment=True, comments=['Great shot!'])
```
If Clarifai's response includes the concepts of either `woman` or `man` but also includes at least `nsfw`, `explicit`, or `suggestive`, InstaPy will not comment. On the other hand, if Clarifai's response includes the concepts of either `woman` or `man` but does not include any of the concepts `nsfw`, `explicit`, or `suggestive`, InstaPy will add the comment `Great shot!`

#### Querying Multiple Models (Multiple API Calls)
In the event that you do not want to set up a workflow (or query more than 5 models), you can also query multiple models by making one API call for each of the models.

**WARNING**: If you are using a free account with Clarifiai, be mindful that the using compound API queries could greatly increase your chances of exceeding your allotment of free 5000 operations per month. The number of Clarifai billable operations per image check equals the number of models selected. For example, if you check 100 images against `models=['general', 'nsfw', 'moderation']`, the total number of billable operations will be 300.

Following the example above, to get general concepts, e.g. `woman`, you would use the model `general` and to check the image for the concepts `nsfw` and `explicit` you would also want to check the image against the NSFW and Moderation models. 

For example:
```python
session.set_use_clarifai(enabled=True, api_key='xxx', models=['general', 'nsfw', 'moderation'])
session.clarifai_check_img_for(['woman', 'man'], ['nsfw', 'explicit', 'suggestive'], comment=True, comments=['Great shot!'])
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1502,9 +1502,11 @@ session.clarifai_check_img_for(['food', 'lunch'], comment=True, comments=['Tasty
 ```
 
 ### Querying Multiple Models with Workflow (Single API Call)
-You can query multiple Clarifai models with a single API call by setting up a custom workflow. To setup a workflow, see the [Workflow Documentation](https://www.clarifai.com/developer/guide/workflow#workflow). **NOTE** :As mentioned above, the `Focus`, `Face Detection`, `Face Embedding`, and `General Embedding` models are not current supported.
+You can query multiple Clarifai models with a single API call by setting up a custom workflow.  Using a `workflow` is the recommended way to query multiple models. Alternatively, it is possible to query multiple models separately (see Querying Multiple Models (Multiple API Calls) below).
 
-**NOTE**: Using a `workflow` is the recommended way to query multiple models. Alternatively, it is possible to query multiple models separately (see Querying Multiple Models (Multiple API Calls) below).
+To setup a workflow, see the [Workflow Documentation](https://www.clarifai.com/developer/guide/workflow#workflow).
+
+**NOTE** :As mentioned above, the `Focus`, `Face Detection`, `Face Embedding`, and `General Embedding` models are not current supported.
 
 Once you have a workflow setup, you can use InstaPy to check images with the Clarifai Image API by setting the `workflow` parameter in `session.set_use_clarifai` to the name of your custom workflow.
 
--- a/README.md
+++ b/README.md
@@ -1401,6 +1401,8 @@ If you wish to check against a specific model or multiple models (see Support fo
 
 To get a better understanding of the models and their associated concepts, see the Clarifai [Model Gallery](https://clarifai.com/models) and [Developer Guide](https://clarifai.com/developer/guide/)
 
+**NOTE ON MODEL SUPPORT**: At this time, the support for the`Focus`, `Face Detection`, `Face Embedding`, and `General Embedding` has not been added.
+
 ```python
 # Check image using the NSFW model
 session.set_use_clarifai(enabled=True, api_key='xxx', models=['nsfw'])
@@ -1420,8 +1422,17 @@ session.set_use_clarifai(enabled=True, api_key=‘xxx’, models=[‘demographic
 # Check image using the Food model
 session.set_use_clarifai(enabled=True, api_key='xxx', models=['food'])
 
+# Check image using the Landscape Quality model
+session.set_use_clarifai(enabled=True, api_key='xxx', models=['landscape quality'])
+
+# Check image using the Logo model
+session.set_use_clarifai(enabled=True, api_key='xxx', models=['logo'])
+
 # Check image using the Moderation model
 session.set_use_clarifai(enabled=True, api_key='xxx', models=['moderation'])
+
+# Check image using the Portrait Quality model
+session.set_use_clarifai(enabled=True, api_key='xxx', models=['portrait quality'])
 
 # Check image using the Textures and Patterns model
 session.set_use_clarifai(enabled=True, api_key='xxx', models=['textures'])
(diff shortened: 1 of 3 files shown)


timgrossmann:
@dfundingsland Awesome addition!
I'll block some time tomorrow to at least look through the code, it might take a while until it is tested but I'm sure this is something that will be interesting for all the people using clarifai.

This might also be a way better base for the face detection in profile image feature some users already requested.

Thank you
Best
Tim

sionking:
please rebase to latest.
can you explain what those models gives that general model don't ?

dfundingsland:
@sionking Rebased. Should be fine now. Please let me know.

As for differences, there are many. Essentially, the non-general models return concepts that which are much more specific. While the general model might return the concept `dress`, the apparel model will return concepts about the specific type of dress (e.g. `cocktail dress`, `midi skirt`, etc.). For users to get the most out of using different models, they will have to do their homework a bit by testing images for potential use cases.

A few more detailed examples below.

**Explicit or NSFW Images**
* **General Model**: Will return concepts like `nude`, `sexy`, or `erotic`
* **NSFW Model**: Will return concept `nsfw` (or `sfw` for safe images)
* **Moderational Model**: Will return concepts like `explicit` or `suggestive` (or `safe` for safe images)

**Landscape Image of Mountain Valley**
**General Model**: Will return concepts like `mountain`, `travel`, `no person`, `landscape`
**Travel Model**: Will return concepts like `view`, `snow & ski sports`

**Catalog Image of Model in Dress**
**General Model**: Will return concepts like `fashion`, `woman`, `model`, `glamour`
**Apparel Model**: Will return concepts like `cocktail dress`, `midi skirt`, `sarong`

I would be happy to provide more examples, but the above should give you an idea of some of the differences. Admittedly, using multiple models is geared more toward power users, but I can also imagine that many users might at least want to check images against both the general model and either the nsfw or moderation models.

timgrossmann:
@dfundingsland Looks and sounds good!

Especially love the documentation, really extensive and descriptive, thank you very much!

uluQulu:
Wow @dfundingsland 👏🏼👏🏼👏🏼🎉
This and #3123 work is amazing, I just read the documentation and is amazed by the addition on Clarifai usage.
Now I really want to reconfigure my Clarifai setting to use those interesting features 😋

Also,
@timgrossmann @sionking @CharlesCCC 
we urgently need a **changelog** file with to contain a short information on the changes after updates.
kind of "_What's New?_" information.
It should basically contain the outer changes to the user interface.
We need it cos, e.g. I have not used InstaPy for a month or more and couldn't keep track of the changes happened so far and I wanted to use new features from that time.
So trying to find those changes made a bit trouble, reading the modifications on the `READM.md` to find the outer changes by running `gitk README.md` command 😒

But if there were a **changelog**, then user could easily read the changes happened so far and benefit those additions 🎊🎊

The structure should be simpler. E.g.,
**changelog.txt**:
```erlang
21/10/2018
Fix to "Failed to load desired amount of users"
Add Bypass suspicious login attempt option for mobile phone  

15/10/2018
Upgrade on Clarifai: Models & Workflow
```

Elsewhere it is a big pain 😟
User can also see the list of commits which is listed in historical order which is good but commits are so bulkish AND also can view the merged PR list which is listed in a weird order..
That's why a **changelog** containing the short information which affects the user & usage after updates [commits] is really important.

Any ideas?

---
Cheers 🤩
---

timgrossmann:
@uluQulu Totally agree with you, the changelog is very much needed.
Could you start it, please? 😉 
Thank you

GOliveira10:
How do you make it where you only leave Clarifai checked comments? The way I have it running now it checks Clarifai, but for posts where it does doesn't pass it still leaves the goofy default comments.

dfundingsland:
@GOliveira10 Good question. At present, I don't believe it's possible to limit comments only to those that match a Clarifai check. @timgrossmann can provide further context on what that is or is not the case.

Depending on the type of content you're targeting, the easiest way would be to specify the type of content you want Clarifai to skip. Taking an example from the docs:

`session.clarifai_check_img_for(['woman', 'man'], ['nsfw', 'explicit', 'suggestive'], comment=True, comments=['Great shot!'])`
In this example, InstaPy would leave a comment if the image contains the concept man or woman, but not if it also contains the concepts nsfw, explicit, or suggestive.

It's not a perfect solution and it's one that requires you to set up some fairly specific rules. Nevertheless, a few of the default comments may still get made.

GOliveira10:
Looking at the code, I think it will work if I just explicitly make the list of comments [] and set comment to 100%. The commenting workflow concatenates the clarifai comments to the regular comments and skips but doesn’t break if comment length is < 1. I’ll report back.

PR #3159, opened 2018-10-18
Fix to "Failed to load desired amount of users"

anna-sunberg:
When follower list is scrolled directly to the bottom, suggestions are showed and follower list is not being loaded. Scroll to the end of list first to initiate the first load which hides the suggestion list.

### DIFF
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -611,7 +611,11 @@ def follow_user(browser, track, login, user_name, button, blacklist, logger, log
 
     return True, "success"
 
-
+def scroll_to_bottom_of_followers_list(browser, element):
+    browser.execute_script(
+            "arguments[0].children[1].scrollIntoView()", element)
+    sleep(1)
+    return
 
 def get_users_through_dialog(browser,
                           login,
@@ -651,6 +655,9 @@ def get_users_through_dialog(browser,
     simulated_list = []
     simulator_counter = 0
 
+    # scroll to end of follower list to initiate first load which hides the suggestions
+    scroll_to_bottom_of_followers_list(browser, dialog)
+    
     # scroll down if the generated list of user to follow is not enough to
     # follow amount set
     while (total_list < amount) and not abort:


cormo1990:
Thanks so much! works good @anna-osipova

vgarba1:
Doesn't work for me. Still failed to load desired number of users.
Are there any specific setting that need to used for this to solve the issue?

cormo1990:
@vgarba1 what’s the traceback? Do you fix the right file in the right position as suggested?

vgarba1:
> @vgarba1 what’s the traceback? Do you fix the right file in the right position as suggested?

Yep opened it in VSCode and put the function on line 614 and the invocation on 657 in unfollow_util.py.
There is no traceback (I not sure what traceback is but Im assuming its was I refer to as a call stack or error stack). It runs perfectly fine, its just it still fails to grab desired number of users. 😞 

Now I don't know python to well, Im a Js/Php kinda guy but is this line correct?  browser.execute_script(
         "arguments[0].children[1].scrollIntoView()", element)
I might be wayyyy off here but should arguments[0].children[1].scrollIntoView() be a string ? Again Im probably wayyy off here Im just comparing it to JS/PHP where whats inside that string is simple a string, where I believe you probably want it to execute.

vgarba1:
Anyway other than that, Im running it with Headless=False and Simulate=True. Im not sure if that would have any effect n the scrolling.

gitpatrickhub:
@anna-osipova works great! thank you!

@vgarba1 Not sure if this would make a difference, but try setting set_simulation to False

vgarba1:
> @anna-osipova works great! thank you!
> 
> @vgarba1 Not sure if this would make a difference, but try setting set_simulation to False

Ill give it a shot!

Olcaey:
still continues to take recommended lists accounts to follow afterwards for me...

Tachenz:
@anna-osipova  For me it gives this error:
untitled

Luisjanz:
Can someone explain what i have to do now? I use the py on raspberry..

gitpatrickhub:
@Luisjanz you have to download the branch and point to it using git;

```
git fetch origin pull/3159/head:pr-3159
git checkout pr-3159
```

Luisjanz:
Can you add me maybe on steam(JohnTheRipper) and you can explain me how to fix it?

Luisjanz:
how can i point it?

timgrossmann:
@anna-osipova Nice and easy fix for the suggestion problem!
Thank you very much

dokuroskull:
it worked for followers at the moment but still not for following

nadavdrukker:
I'm having trouble with this patch. If I request many users it works fine, if I request around 10 it returns none: "Grabbed 0 ...."

cormo1990:
@nadavdrukker Try with the latest release, it works for me. Let us know

nadavdrukker:
@cormo1990 I'm still getting this problem with the latest release. Sometimes returns desired amount and sometimes 0. Certainly for small number of requested follows returns 0. I'll try to find a minimal consistent unworking example.

nadavdrukker:
A bit more diagnostics. If I ask for a low number (up to 20) and those are available on the first screen (after a partial scroll to eliminate the suggeted users), it grabs 0. If I ask for more, or not enough of the users on the first screen are available, it starts scrolling further, loading more users. In that case it does grab them.

uluQulu:
Thanks for the shortest fix, @anna-osipova 🔥
@nadavdrukker, have a read this https://github.com/timgrossmann/InstaPy/issues/3173#issuecomment-433670354 to get the solution.

myehrajat:
i indent as you said and add function but suddenly it close it may is not compatible with ne version indent @anna-osipova

bergeny:
Thanks @anna-osipova ,Now It works like a charm

Issue #2827, opened 2018-09-02
Question - optimisation likes requests


YannikFirre:
Hello,

I have an optimization proposal about the like feature.

If we run the script often and / or on weak hashtags, we quickly have images already liked.

If I'm not mistaken the script takes the list of the most recent posts, so we could add an option that would give the possibility to stop the script for the current hastags if they fall on a post already liked.

We would save time and requests.

What do you think ?

Nocturnal-2:
I agree!

sionking:
Simple imlementation, database the liks

YannikFirre:
Or just stop the script when the returned post is already liked ^^ No need database ;-)

sionking:
I have implemented both I need to PR

YannikFirre:
Anyone have a opinion on this ? interesting feature?

PR #3164, opened 2018-10-19
ADDED - break if image already liked

YannikFirre:
Hello,

Break the current loop if one image is already liked. #2827 

What do you think ?

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -491,12 +491,15 @@ class InstaPy:
 
         return self
     
-    def set_break_after_already_liked(self, active=True) :
+    def set_break_after_already_liked(self, after=5) :
         """Break the loop if one image is already liked"""
         if self.aborting:
             return self
 
-        self.break_after_already_liked = active
+        if after <= 0 :
+            after = False
+
+        self.break_after_already_liked = after
 
         return self
     
@@ -1096,6 +1099,7 @@ class InstaPy:
 
         liked_img = 0
         already_liked = 0
+        already_liked_continuous = 0
         inap_img = 0
         commented = 0
         followed = 0
@@ -1167,6 +1171,7 @@ class InstaPy:
                                                      self.logfolder)
 
                         if like_state == True:
+                            already_liked_continuous = 0
                             liked_img += 1
                             # reset jump counter after a successful like
                             self.jumps["consequent"]["likes"] = 0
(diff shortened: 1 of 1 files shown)


sionking:
Will it break the tag/location loop ? not a good idea, maybe you should count at least 5 .

YannikFirre:
Why after x count ? if one image is already liked, the following ones too. 

No ?

Nocturnal-2:
> if one image is already liked, the following ones too

@InfinityWebMe not necessarily. While liking images using closely related tags, there are chances that some images are posted with both these tags. For example, consider the tags 'pursuitofportraits' and 'portrait_perfection'. Assuming that both these tags are in _like_by_tags_ list and assuming we like by  'pursuitofportraits' first and 'portrait_perfection' next, we might encounter an image which has both these tags. So if we break on first _already liked image_ while traversing through 'portrait_perfection' (since it also had hashtag 'pursuitofportraits'), we will miss other images which did not use 'pursuitofportraits' but only used 'portrait_perfection'.

So like @sionking has mentioned, it will be better to break only if there is an _x_ number of _continuous already liked images_.

YannikFirre:
Good catch, i will change this PR :)

redsector72:
I think the best way to address this is the blacklist. 
If a user is in blacklist, it is skipped.
No need to break cycle.

YannikFirre:
@redsector72 i don't understand what you mean ? How the blacklist can affect the images already liked ?

YannikFirre:
@sionking @Nocturnal-2  Like this ?

redsector72:
@InfinityWebMe you can use the black list function. When you first like, the user is blacklisted and will never get a second like. there is no need to break, as the users are skipped already.

InfinityWebMe:
@redsector72 okey but you need to remove them for the next tag/action. 

Use the blacklist function is not enought.

YannikFirre:
@redsector72 Not if you use multiple tags.

YannikFirre:
@timgrossmann  Any review on this ? :)

converge:
hey mate, it's a really nice idea!

I tested it, but it doesnt seam to work for me, I set:
```python
session.set_break_after_already_liked(after=2)
session.like_by_tags(['balao', 'bike'], amount=5)
```
but it keeps liking already liked posts

To merge it we will also need an update in our README file explaining the feature.

Congratulations! Some adjust and we're ready to merge it.

timgrossmann:
@InfinityWebMe Any updates on this? 😊

timgrossmann:
@InfinityWebMe Any updates on this? 😊

timgrossmann:
@InfinityWebMe Unfortunate that we had to close this... 
I'm really sorry to see this.

Maybe if we can update everything to the current state, we can get this feature merged 😊

PR #3174, opened 2018-10-20
Add Bypass suspicious login attempt option for mobile phone 

imkimchi:
I run a lot of accounts with instapy, and sometimes user didn't add their email to instagram, or they'd prefer phone number over email so I thought It'd be really great to give people an choice for bypass suspicious login attempt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -883,6 +883,11 @@ email, and you will be prompted to enter the security code sent to your email.
 It will login to your account, now you can set bypass_suspicious_attempt to False
 ```bypass_suspicious_attempt=False``` and InstaPy will quickly login using cookies.
 
+If you want to bypass suspicious login attempt with your phone number, set `bypass_with_mobile` to `True`
+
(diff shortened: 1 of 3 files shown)


timgrossmann:
@imkimchi Sounds really useful.
Code looks good, too.

I hope we can find some people who can trigger this to test it with their accounts.

Thank you very much for the addition!

sionking:
I only login manually

imkimchi:
@sionking but I'm pretty sure that's not the case for most of the people. Especially when you have to bypass suspicious login attempt in CLI environment like Ubuntu.

imkimchi:
@timgrossmann I already tested it with multiple accounts, and currently using it for 20+ accounts! If you think we would need more test for this, feel free to do it 🙂

timgrossmann:
@imkimchi Sounds good.
Thank you very much, this will definitely help.

PR #3178, opened 2018-10-20
Duplicate comments fix

butaminas:
I've noticed that this issue isn't fixed even though several people mentioned it in the issues (as well as myself a while ago) and it I think it is a serious issue so I decided to share my fix.

Basically, I have added a user_id parameter to be logged at start and passed every time to **comment_image** method. In the method itself it goes through each comment and checks if any of the comments author id is your user id and if so, it skips the post.

Perhaps this code could be optimized, I don't know, this was actually my first interaction with python but it works like a charm for me. Any improvements are welcome.

Hope it helps.

### DIFF
--- a/instapy/comment_util.py
+++ b/instapy/comment_util.py
@@ -47,7 +47,7 @@ def open_comment_section(browser, logger):
 
 
 
-def comment_image(browser, username, comments, blacklist, logger, logfolder):
+def comment_image(browser, username, comments, blacklist, logger, logfolder, user_id):
     """Checks if it should comment on the image"""
     # check action availability
     if quota_supervisor('comments') == 'jump':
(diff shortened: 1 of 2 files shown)


timgrossmann:
@butaminas You're absolutely right that we have to take care of that, I'll try to find some time to dig into your code and understand your approach a bit better.

Thank you very much for this fix and congrats for your first PR to InstaPy.

sionking:
it seems to be a working solution but I think it a bit overkill :)

butaminas:
> 
> 
> it seems to be a working solution but I think it a bit overkill :)

You could just get the user id every time before going through the comments but I've decided to store the ID at the beginning just so there wouldn't be additional increase of requests to Instagram API.

If it's not a big deal, then you could just remove the **user_id** from **instapy.py** and just insert the
```
user_id = browser.execute_script(
            "return window._sharedData.config."
            "viewer.id")
```
code in the **comment_image** method.

varun13169:
@butaminas my issue is different.
Mine is the ban that Instagram puts when posting same comment even on different posts, "[Same comment multiple times](https://help.instagram.com/161863397286564)"

timgrossmann:
@sionking @butaminas I'm kind of asking myself why InstaPy comments on one post several times...
As far as I know, it should first test whether the post has already been liked before commenting on it 🤔

butaminas:
> 
> 
> @sionking @butaminas I'm kind of asking myself why InstaPy comments on one post several times...
> As far as I know, it should first test whether the post has already been liked before commenting on it 🤔

If a post has a like it doesn't mean it also has a comment. You might have liked a post in the past or used a configuration that only liked posts and later use a configuration to comment posts.

In my mind, the only solution to this problem is to actually check if you made a comment or not to be 100% sure in all cases.

PR #3179, opened 2018-10-21
Support for Checking Video Inputs with Clarifai

dfundingsland:
Updated a couple of the functions in clarifai_util.py to add support for checking videos via the Clarifai API. Since workflows do not take video inputs and only a few of the models accept video inputs, in many cases the keyframe will still be checked for content. 

Also fixed blocks for compiling the list of concepts for the Celebrity and Demographics models.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ _The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ## [Unreleased] - 2018-11-01
 ### Added
 - Interact with tagged images of users, and validation of a user to be optional
-- Use Clarifai to check video content by setting `check_video=True` in `session.set_use_clarifai`. By default `check_video` is set to `False` and should only be used if necessary (_see **README**_ for further details). Clarifai will check one frame of video for content for every one second of video. Feature to be updated with ability to check frames every X seconds as soon as Clarifai enables that feature in their API.
+- Use Clarifai to check video content. By default deactivated and should only be used if necessary.
 
 ## [Unreleased] - 2018-10-29
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ _The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ## [Unreleased] - 2018-11-01
 ### Added
 - Interact with tagged images of users, and validation of a user to be optional
+- Use Clarifai to check video content by setting `check_video=True` in `session.set_use_clarifai`. By default `check_video` is set to `False` and should only be used if necessary (_see **README**_ for further details). Clarifai will check one frame of video for content for every one second of video. Feature to be updated with ability to check frames every X seconds as soon as Clarifai enables that feature in their API.
 
 ## [Unreleased] - 2018-10-29
 ### Added
--- a/README.md
+++ b/README.md
@@ -1562,6 +1562,21 @@ We have 3 options:
 2. user:pass@ip:port
 3. None
 
+### Checking Video
+**WARNING**: Clarifai checks one frame of video for content for every second of video. **That is, in a 60 second video, 60 billable operations would be run for every model that the video is being checked against.** Running checks on video should only be used if you have special needs and are prepared to use a large number of billable operations.
+
+To have Clarifai run a predict on video posts, you can set the `check_video` argument in `session.set_use_clarifai` to `True`. By default, this argument is set to `False`. Even if you do not choose to check the entire video, Clarifai will still check the video's keyframe for content.
+
+For example:
+
+```python
+session.set_use_clarifai(enabled=True, api_key='xxx', check_video=True)
+```
+
+With video inputs, Clarifai's Predict API response will return a list of concepts at a rate of one frame for every second of a video.
+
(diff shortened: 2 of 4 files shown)


timgrossmann:
@dfundingsland Is this checking multiple images from the video or the whole video itself?

dfundingsland:
@timgrossman They way I have it setup is like this. First it will check to see if the `class` for the `img` tag matches that of a regular photo post. If not, it will check for the `class` for the `video` tag and `img` tag of a video post. If the model being used to check the image is one of the supported models, i.e. Apparel, Food, General, NSFW, Travel, or Wedding, then the whole video is processed at 1 frame per second with concepts given for each frame. If using a model not in that list or using a workflow, then only the keyframe image of the video will be checked.

dfundingsland:
@timgrossman They way I have it setup is like this. First it will check to see if the `class` for the `img` tag matches that of a regular photo post. If not, it will check for the `class` for the `video` tag and `img` tag of a video post. If the model being used to check the image is one of the supported models, i.e. Apparel, Food, General, NSFW, Travel, or Wedding, then the whole video is processed at 1 frame per second with concepts given for each frame. If using a model not in that list or using a workflow, then only the keyframe image of the video will be checked.

dfundingsland:
@timgrossman They way I have it setup is like this. First it will check to see if the `class` for the `img` tag matches that of a regular photo post. If not, it will check for the `class` for the `video` tag and `img` tag of a video post.

 If the model being used to check the image is one of the supported models, i.e. Apparel, Food, General, NSFW, Travel, or Wedding, then the whole video is processed at 1 frame per second with concepts given for each frame. If using a model not in that list or using a workflow, then only the keyframe image of the video will be checked.

dfundingsland:
@timgrossman They way I have it set up is like this. First, it will check to see if the `class` for the `img` tag matches that of a regular photo post. If not, it will check for the `class` for the `video` tag and `img` tag of a video post.

 If the model being used to check the image is one of the supported models, i.e. **Apparel**, **Food**, **General**, **NSFW**, **Travel**, or **Wedding**, then the whole video is processed at 1 frame per second with concepts given for each frame. If using a model not in that list or using a workflow, then only the keyframe image of the video will be checked.

sionking:
TOO EXPENSIVE :)

timgrossmann:
@dfundingsland I think it general it's a really good idea, but as @sionking pointed out, this is really "expensive". We only have 5k free API calls at Clarifai per month, at a rate of 1/s it's going to be lit with very few videos.

I think only taking 4-5 images at random positions might be a better solution. (Even this might be really "expensive").

Still, thank you very much for pointing this out and creating a PR.
It's really important that InstaPy gets new ideas and people with new ideas.

dfundingsland:
@timgrossmann In regards to @sionking's point of it being too expensive, I agree that that may be the case. Unfortunately, that's just how Video predict using Clarifai works at the moment (that is, it's nothing specific to my implementation). A 10s video would consume 10 billable operations. At present, I think the maximum length for a video in Instagram is approximately 60 seconds, give or take a few seconds.

As for taking 4-5 images at random, that may be possible in the future judging from Clarifai's FAQ:

```
Can I change that frame-rate so it only gives me tags once every X seconds?

Not quite yet, but we'll be introducing this very soon! Stay tuned.
```

timgrossmann:
@dfundingsland Nice, good to know that they are working on this!
Thank you for the clarification 

Let's keep this open until they add the feature to only sample every x seconds.

dfundingsland:
@timgrossmann Yes, will keep an eye out for the update. The question is, however, even with the upcoming changes to the API, is checking video for content something that the community wants? If so, should it be on by default or should checking video be optional? I can see a use for it, but I don't know if I'm an outlier. Interested in your thoughts. 

On a side note, there are a couple of fixes in this PR that I still should be made. Happy to make another PR if we wish to wait on checking video for the time being.

timgrossmann:
@dfundingsland Depending on how strictly people want to filter, I definitely think this is a useful feature.
Unfortunately, there is no way to track how many people actually use the calrifai feature at all.

It should definitely be turned off by default.

Sure, please go ahead and separate out those changes from this PR, would love to merge them.

sionking:
> It should definitely be turned off by default.

The video check ? Turn OFF BY DEFAULTS !

dfundingsland:
@sionking Hear you loud and clear ;-) @timgrossmann I'll revisit this PR when I have a chance and keep you posted once I've submitted a revised version.

sionking:
The are holding my credit card and have no quota limits so video check can cost me like a house lol

dfundingsland:
@timgrossmann and @sionking Added an argument `check_video` with default set to `False`. Take a look and let me know if there are any other concerns.

timgrossmann:
@dfundingsland This is definitely the right approach for this. We need to make sure that nobody uses it accidentally.

Could you please add an entry to the CHANGELOG.md file following the format it already has?

I think we can push this feature into the master branch with the "statement" that you'll add the possibility to only take frames every x seconds (or random ones) as soon as clarifai enables that feature in their API.

Thank you very much for your work.

dfundingsland:
@timgrossmann Changelog updated.

I'd be more than happy to provide an update once Clarifai has updated the API.

timgrossmann:
@dfundingsland Awesome to hear that, thank you very much!

dfundingsland:
@timgrossmann I thought I might have gone overboard with the changelog ;-) Just making sure to cover all the bases you previously mentioned.

Issue #3202, opened 2018-10-23
I need help understanding blacklist


redsector72:
HI there,
I am having trouble understanding how the blacklist works and I'm pretty sure there is a bug in there.
I need help understanding the code as my programming skills are newbie.

The fact is that I keep getting "--> Image already liked!" even with the blacklist active.
I considered that "maybe" the already liked was past likes, so I've modified the code to add to blacklist also the users that get into the "--> Image already liked!" process.

The problem is that I see duplicates entries in the blacklist.
This make me think there is a bug in the evaluate user process and the blacklist.

Can someone with more experience take a look inside the blacklist functions and tell if there is a bug?

jeremycjang:
The blacklist works by creating a `.csv` of past users that you’ve interacted with through the program that it then references when it sees a link from a user you’ve already interacted with in the past. 

I haven’t opened the blacklist in a while, but believe it saves a line per action you’ve taken on a user (ex. One line for following user1, one line for liking user1’s posts). Not sure which actions you’re using, but the program grabs links dynamically then compares the links to the blacklist when deciding whether or not to action on them. If whatever list or hashtag you’re pulling from doesn’t have a lot of new posts or users, the program’s going to end up pulling a lot of links you’ve already interacted with then will display that message after realizing you’ve interacted with it before.

It might not even be related to the blacklist and could just be you’re finding posts using multiple hashtags that you’ve specified and is pulling the same link that you’ve already liked multiple times

Hope that helps!

redsector72:
Hi @jeremycjang,
this is clear.
If I look at the source code, function validate_username I see that there is no "open file" o "file check".
there is only "if username in blacklist: return false"
My understanding of python is limited, but I'm pretty sure that the black list is not woking.
If I create a simple test, with any like function, and run it twice, i get 2 entry in the blacklist and I never see "is in blacklist  ~skipping user".
I suppose there is an error in the code, but I can'f figure it out

maybe @sionking can find it out!

redsector72:
I have digged out more.
I suppose I have corrected the behavior.
It is definitely missing all the code regarding the file.
I will make a pull request

Issue #3105, opened 2018-10-09
quota supervisor does account the "already" liked?


redsector72:
@uluQulu it's not clear to me if the quota supervisor does count the call to the like action or the actual successfull call.
I have a bunch of "already liked".
Maybe using the blacklist will help?
Thanks!

uluQulu:
@redsector72 
Quota Supervisor does not have such a misbehavior as you have imagined.
**1**-) It checks the action [e.g. like] availability when entering the action [like] engine
**2**-) If a like is available, it tries to like
**3**-) If like is successful, it updates the likes record in the **recordActivity** table of the database

Shortly,
>it's not clear to me if the quota supervisor does count the call to the like action or the actual successfull call.

No. It will not update the likes' record in **recordActivity** table of the database which QS does use to check action availability cos "_already liked_" label is not a successful like.

If you read a bit of source code, you could realize all of that.

-------

Cheers 😀
-----

redsector72:
thanks @uluQulu, I was unsure from the source code. I've not yet used it.
from the readme seems there is a jump out function, but it is not clear.
If I don't specify sleep, then it jump out?

PR #3206, opened 2018-10-23
Update util.py

redsector72:
Referring to  #3202 and #3105 and #2824 I found out the Blackslist is really buggy.
Here is my "workaround" patch for #3202.
Maybe someone else will dig into it and discover why it is not working at all.
I had trouble finding the "log directory" but my knowledge is limited. this version will work with "multi=true"


timgrossmann:
@redsector72 Thank you very much for your contribution.
I seldomly see people really use this feature so I'm sorry you're having issues.

Could you explain what goes wrong or what the wrong behaviour looks like, please?

redsector72:
Sure.
The blacklist is written but according to my experience (python 2.7) it is never referenced. Maybe it works with higher version of python as is, i can't check this.

I've implemented the check if an user is in the blacklist in the user check functions. As is it works on python 2.7. If the user is in the blacklist it return false and there is not "like" action.

Now I see the "already liked" dropping in my stats, as expected.

instapy/like_util.py
```diff
@@ -558,8 +558,10 @@ def like_image(browser, username, blacklist, logger, logfolder):
     if quota_supervisor("likes") == "jump":
         return False, "jumped"
 
-    like_xpath = "//button/span[@aria-label='Like']"
-    unlike_xpath = "//button/span[@aria-label='Unlike']"
+#    like_xpath = "//button/span[@aria-label='Like']"
+#    unlike_xpath = "//button/span[@aria-label='Unlike']"
+    like_xpath = "//button/span[@class='glyphsSpriteHeart__outline__24__grey_9 u-__7' and @aria-label='Like']"
```
timgrossmann: Could you help me with this part?
Is this a special selector for the elements you need?
They are too specific, what is the problem with the ones you commented out?

If you replace selectors, please remove the old ones and don't comment them out.

Thanks 😊

redsector72: This part should not be in the Commit. 
I'm not really good with GIT.
Anyway this part is to fix the like button change in the new update, like #3229

redsector72:
I'm sorry I'm not able to remove the commit from like_util.
Can some one leave only the one for util or I have to make a new pull request?

timgrossmann:
@redsector72 Sorry for that, I think creating a new one would be better because all the commits will appear in the commit history.

Hope that's ok for you.
Thank you

PR #3214, opened 2018-10-24
Change every place we have following/ followers count to one function

sionking:
Didn't finish change all.


timgrossmann:
@sionking Good catch, thank you very much

uluQulu:
Yeah it's a so good change 😂 @sionking
`get_given_user_following` should benefit that function 😛 too

PR #3230, opened 2018-10-26
Use proxy to access clarifai; return clarifai tags to caller

sionking:
read the title.
@timgrossmann Documented for first time hope it is fine :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1519,7 +1519,7 @@ Let's say you want to comment 'Great shot!' on images of men or women with the h
 
 For example:
 ```python
-session.set_use_clarifai(enabled=True, api_key='xxx', workflow=['your-workflow'], proxy=123.123.123.123:5555)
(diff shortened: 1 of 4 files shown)


timgrossmann:
@sionking Awesome documentation! 😄 
Thank you for PRing this, it's very important to be able to do this.

One thing I saw is that the proxy is not passed as string value.
Is this a definition of Claraifai? There are a lot of proxy services that use an url as IP as well (e.g. Luminati). Will this work with them?

sionking:
can you give an example?

timgrossmann:
@sionking Absolutely!
`http://${username}-session-${sessionId}:${password}@zproxy.lum-superproxy.io:${port}`This is the proxy I'm using for a project. Using Luminati proxy service

sionking:
@timgrossmann it uses Requests HTTP Library So yes it input should be string I fixed the doc.

timgrossmann:
@sionking Perfect, thank you!

PR #3238, opened 2018-10-27
Fix logger in verify commenting

dbknv:
Fix missing argument in verify_commenting() in like_by_users
> TypeError: verify_commenting() missing 1 required positional argument: 'logger'


timgrossmann:
@sterid Awesome, good spot!

PR #3252, opened 2018-10-29
Fixed indentation in clarifai_util 

dfundingsland:
Fixed indentation in `clarifai_util` and added definition for `clarifai_proxy` inside `__init__`


timgrossmann:
@dfundingsland Good catch, thank you very much!

PR #3253, opened 2018-10-29
[Proposition] Add option to inject selenium driver for remote session

janekbaraniewski:
Hey,
I wanted to propose making remote selenium driver injectable. This would allow providing pre-configured driver instance to InstaPy, which now is impossible. 
My use-case is running several sessions in parallel from Celery workers which fails with current configuration of remote Selenium session.
Let me know what you think, sorry for not making an issue first to discuss it but it's such a small change that I just went ahead and did it 😄

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -363,20 +363,30 @@ class InstaPy:
 
         return self
 
-    def set_selenium_remote_session(self, selenium_url=''):
-        """Starts remote session for a selenium server.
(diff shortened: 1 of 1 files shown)


timgrossmann:
@janekbaraniewski Perfect and absolutely not breaking anything 😉 

Thank you very much!

janekbaraniewski:
No problem, glad I could help out ;)

PR #3263, opened 2018-10-30
Update to let Blacklist Actually be referred

redsector72:
This PR add support for the blacklist check.
The actual code does not work for me on python 2.7, my modification works perfectly.
Take care

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,27 +7,12 @@ _The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased] - 2018-10-29
 ### Added
-- This CHANGELOG file to hopefully serve as a useful resource for InstaPy fellas to stay up-to-date with the changes happened so far.
-- **Custom action delays** capability (_see **README**_).
-- Now follow engine has the same compact _action verification procedure_ used in the unfollow engine.
-- Lots of stability in action verification steps which are held at newly added `verify_action` function.
-- New quickstart templates from 11 different people shared at #3033.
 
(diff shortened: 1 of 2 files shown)


timgrossmann:
@redsector72 Perfect, looks way better now!
Let's just check it and merge this asap.
Thank you

timgrossmann:
@redsector72 Just saw that we're doing this for every username that has to be validated.
What is the reason for positioning this code snippet there?
Wouldn't it be better to load the blacklist once?

redsector72:
My programming knowledge is limited. 
The blacklist function was already there in the code as this is where each user get validated.
Probably the original code was in accordance with your idea, but it does not works as expected in python 2.7.
I do not have the knowledge to "load the blacklist once"

timgrossmann:
@redsector72 I think we can merge this for now, I've seen that this method is not called too extensively.
Could you add an entry to the CHANGELOG file as a fix, please?

redsector72:
I hope to have done it correctly.

timgrossmann:
@redsector72 Fixed it 😉 

Thank you for this PR

LotusPainter:
Does this work in regards to when the bot grabs followers from an account? i.e 

```
User 'EXAMPLEUSER' [1/1]
Grabbed 5 usernames from 'EXAMPLEUSER's `Followers` to do interaction.
```

Before grabbing the 5 usernames the bot should check to make sure none are in the blacklist. Otherwise it will waste time skipping each user that is in blacklist, and also subsequently increase the count/amount on `session.interact_user_followers ` which of course we want to avoid for skipped accounts.

PR #3265, opened 2018-10-30
fix: was trying to click login button before language change, so Log …

converge:
fix: was trying to click login button before language change, so **Log In** button couldnt be located in other languages


timgrossmann:
@converge Awesome! How could this go through for so long without anyone noticing it 😅
Thank you very much for the fix!

converge:
hey @timgrossmann !! I was checking the commit history, couldn't find when it was changed 😄

PR #3287, opened 2018-11-04
Fixed the login button xpath detection 

arshiyasolei:


sionking:
Why you all not moving to manual login ? much easier

timgrossmann:
@sionking Could you please explain the process of manually logging in?

instapy/login_util.py
```diff
@@ -253,7 +253,7 @@ def login_user(browser,
         update_activity()
 
     login_button = browser.find_element_by_xpath(
-        "//button[text()='Log in']")
+        "//button[contains(.,'Log in')]")
```
timgrossmann, line 256:
```
        "//button[contains(.,'Log in')]")
```

What does the `.` do before the comma? Is it referring to the currently selected element?

sionking:
@timgrossmann  nothing fancy I just stuck on the login page, do the login manually so if there is a surprise I can handle it. then continue the application after login successful.

converge:
hey guys, looking for the Log In button in this way will match two buttons, (Log In using the Facebook button) and the (Instagram Log In button).

login screen

instapy/login_util.py
arshiyasolei: https://stackoverflow.com/questions/46486477/why-wont-my-xpath-select-link-button-based-on-its-label-text

Hope this helps

arshiyasolei: https://stackoverflow.com/questions/38240763/xpath-difference-between-dot-and-text

This one has better examples

fgisslen:
@sionking If you are running 200-300 accounts it's pretty hard with manual login...

converge:
thank you for your suggestion @arshiyasoleimani , unfortunatelly it will break InstaPy.

arshiyasolei:
@converge how so?

arshiyasolei:
This does not break instapy

converge:
@arshiyasoleimani as said above, there are two Login buttons
```"//button[contains(.,'Log in')]")```
will match two buttons, when we use Selenium, we need to be specific, if not, it will click on the wrong button.

arshiyasolei:
@converge currently, your method is not working. I used this to fixed my own code. It does not cause any conflict with the mentioned login buttons. My code takes the **first** login button.

arshiyasolei:
@timgrossmann it would be great if you guys reconsider this.

converge:
@arshiyasoleimani no worries, you guys can keep working on this topic.

timgrossmann:
@arshiyasoleimani @converge Is right, this is way to breakable... As soon as something changes, the first element might not be the login button anymore.

We need to find a better solution

arshiyasolei:
@timgrossmann if we ever decide to be more descriptive with the xpath stuff (compared to what I worte) , there will be a higher chance of code breaking because they might change the UI (we never know).
I don't think they will change the login button order soon. 
Also keep in mind that the current original version is broken on my test pc.

timgrossmann:
@arshiyasoleimani Just had to set it up from scratch on my server, the login process with the current approach works fine on my side. Could you share the error message on your end?

PR #3292, opened 2018-11-05
adds headless mode Firefox support

converge:


timgrossmann:
@converge Awesome, Joao, thank you!

What do you need this for, if I may ask? What's the reason you're using this over chrome?

converge:
Hey Tim! I prefer  this approach because in this way I don't need to work with chrome extensions to use proxy. And Firefox is sexier 💃

sionking:
Many configurations don't exist in Chrome unless extenstion. and no extenstion in headless. I might switch soon.

PR #3313, opened 2018-11-07
Update .travis.yml

converge:
PEP8 enforce us to improve the code quality, E127 propose that this is a bad thing (example):
```python
def get_links_for_username(browser,
                            username,
                             amount,
                             logger,
                             randomize=False,
                             media=None,
                             taggedImages=False):
```
And this is a good thing:
```python
def get_links_for_username(browser,
                           username,
                           amount,
                           logger,
                           randomize=False,
                           media=None,
                           taggedImages=False):
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ install:
     - pip install flake8  # pytest  # add another testing frameworks later
 before_script:
     # stop the build if there are Python syntax errors or undefined names
-    - flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
+    - flake8 . --count --select=E127,E901,E999,F821,F822,F823 --show-source --statistics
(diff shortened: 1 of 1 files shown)


timgrossmann:
@converge Good addition, could you maybe please also reformat the code according to E127 so that the CI isn't failing?

converge:
I dont think this is a good idea, there is 1500 lines of travis warnings to fix. anyone to help ? @world

timgrossmann:
@converge Hmm... That's the point, I don't think it's a good thing to keep the CI on failed.
Maybe someone has good tooling for changes like this.

converge:
@timgrossmann you're right.. I'll try another approach in the next days to have it.

timgrossmann:
@converge Sounds perfect, looking forward to it.
Thank you

PR #3337, opened 2018-11-09
Update README.md

converge:
Hey guys, we have actually 1.500+ open issues, I would like to help improve it. I see two good ideas:

- have a better channel support (Discord)
- setup an issue bot (to close inactive issues)

In this way we could focus on bug fixes and general project improvements.


timgrossmann:
@converge I think we need both. 
The issue bot is absolutely necessary. 

The switch from Slack to Discord is a good idea since we're running close to the limits of free slack usage.

PR #3347, opened 2018-11-11
requirements update

converge:
It will help people with **no module named ___** issues, now they just have to 
```
pip3 install -r requirements
```


requirements.txt
```diff
@@ -1,10 +1,18 @@
...
+clarifai==2.4.1
+configparser==3.5.0
+EasyProcess==0.2.3
+emoji==0.5.1
+future==0.17.1
+googleapis-common-protos==1.5.5
+grpcio==1.16.0
+idna==2.7
+jsonschema==2.6.0
+plyer==1.3.1
```
uluQulu, line 12: Hey @converge 
Good that plyer `1.3.1` was released. It solves some bugs in older release.

But why don't you prefer to use `>=` in versions?
You know, using `==` requires big care. And I don't think some new version of e.g. _selenium_ will not work for some people.
It's just _webdrivers_ which are buggy in latest releases and they are not in PyPI.

But in general scope, it doesn't matter. Users just will see warnings...

converge: Hey @uluQulu , sounds good!

I'll be busy all day long today, if you have some time, could you update it ? If not, I'll be able to update it tomorrow.

PR #3362, opened 2018-11-12
Add issue bot for automatic closing of stale issues

timgrossmann:
Based on @converge idea about having a bot that closes old and stale issues to get a better overview on the still remaining issues and especially the most recent ones that haven't been fixed, this PR introduces a bot that marks issues where there was no activity in the last 30 days. After 7 more days the issue will then be closed.

Any comments or requests about the time till stale and time till close?

Thank you for this very good idea, @converge! 👍 

> P.S. we really need to get that PEP change working with the pipeline...

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,57 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+ # Number of days of inactivity before an Issue or Pull Request becomes stale
+ daysUntilStale: 30
+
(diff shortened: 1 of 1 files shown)


converge:
I think it's not working..

timgrossmann:
@converge Why do you think that? 
I had >500 commented issues 😉 

It just takes a while to start up

converge:
@timgrossmann I was anxious to see it working 😆 
It's working now.

PR #3393, opened 2018-11-19
Issue with cookie for user Creating new cookie fix

ugurozturk:
Hi, I was had the error. Then i did set Headless to false and debug myself. So i realized after cookie inserted the browser was staying on the login page. So everything goes wrong.


IH4xx3R:
thanks for this fix !

timgrossmann:
@ugurozturk Thank you for helping fix this problem.

PR #3420, opened 2018-11-23
fix the xpath for likers grabbing and some PEP8 ind

sionking:


timgrossmann:
@sionking Nice one, thank you very much!

PR #3428, opened 2018-11-25
Check_link > Switch mandatory_words to ANY instead of ALL

visualheroes:
Change mandatory_words to ANY instead of ALL under check_link def. Mandatory words for commenting is already using ANY so maybe this was a typo.

Mandatory Words ALL might be useful on rare occasions but it is very restrictive and use quite a lot of connections. 

**Edit: if you want to keep both ANY and ALL, we can introduce any_mandatory_words and all_mandatory_words (for commenting and liking). Let me know!**

### DIFF
--- a/README.md
+++ b/README.md
@@ -843,7 +843,7 @@ session.set_mandatory_words(['#food', '#instafood'])
 ```
 
 `.set_mandatory_words` searches the description, location and owner comments for words and
-will like the image if **all** of those words are in there
+will like the image if **any** of those words are in there
 
 ### Restricting Likes
 
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -526,7 +526,7 @@ def check_link(browser, post_link, dont_like, mandatory_words, ignore_if_contain
         image_text = image_text + '\n' + location_name
     
     if mandatory_words :
-        if not all((word in image_text for word in mandatory_words)) :
+        if not any((word in image_text for word in mandatory_words)) :
             return True, user_name, is_video, 'Mandatory words not fulfilled', "Not mandatory likes"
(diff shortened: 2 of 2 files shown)


timgrossmann:
@visualheroes First of all, thank you for contributing!

The all in the method comes from the wording "mandatory_words" which implies that all words given have to be in there.
Instead of creating two distinct methods we should use a `all` parameter that, by default is `False` and can be set to `True` to enable a more restrictive checking.

Could you also please update the Changelog file with the modifications you did?

Thank you very much

oskarkrawczyk:
> The all in the method comes from the wording "mandatory_words" which implies that all words given have to be in there.

Just wanted to chime in. I also got the impression that it was `ANY` not `ALL` – to me, a non-native speaker `mandatory_words` doesn't imply `ALL` at... all.

timgrossmann:
@oskarkrawczyk That is very good to know! Thank you very much for mentioning this.

We'll merge this separately then and create a new PR if we should need the `all`.

visualheroes:
@timgrossmann Thanks for the merge!
You are right tho, we need an option for any in case someone needs it. 
I'll try to add it later, i'm working on another pull request at the moment (mandatory language). ;)

timgrossmann:
@visualheroes Sounds interesting, looking forward to it.
Maybe this can work together with the PR of @uluQulu about the language

visualheroes:
@timgrossmann 
I'll do the pull request now, well, i'm trying as i'm pretty new to this github thing. ;)

It wont interfer with @uluQulu as it's based on character set and it's only implemented in check_links.

I tested it the last two week without problem.

PR #3469, opened 2018-12-01
Changed link to english

meinert:


timgrossmann:
@meinert Good catch, thank you very much for updating the documentation 😉

PR #3475, opened 2018-12-02
fix record_activity been recored multiple times per day for the same account

converge:
issue: duplicated profile_id ->
database

When checking for a previous record on the same day , we were considering the HOUR also, long story short:

///
**before**
- 1st run:
```record activity for 2018-12-02 10:00:00```
- 2nd run at 11:00:00
```check if exists todays activity record checking as "date AND hour"```

**result:**
1 record for 2018-12-02 10:00:00 (with activity records)
1 record for 2018-12-02 11:00:00 (with activity records)

///
**now**
- 1st run:
```record activity for 2018-12-02 10:00:00```
- 2nd run at 11:00:00
```check if exists todays activity record checking as "date"```

**result:**
1 record for 2018-12-02 10:00:00 (with all activities that ocurred this day)

### DIFF
--- a/instapy/util.py
+++ b/instapy/util.py
@@ -324,15 +324,20 @@ def update_activity(action="server_calls"):
         conn.row_factory = sqlite3.Row
         cur = conn.cursor()
         # collect today data
-        cur.execute("SELECT * FROM recordActivity WHERE profile_id=:var AND "
-                    "STRFTIME('%Y-%m-%d %H', created) == STRFTIME('%Y-%m-%d %H', 'now', 'localtime')",
-                    {"var": id})
+        # check if there is existent records for the current day
+        cur.execute("SELECT * FROM recordActivity "
+                    "WHERE profile_id = ? "
+                    "AND STRFTIME('%Y-%m-%d', created) == "
+                    "    STRFTIME('%Y-%m-%d','now', 'localtime')", (id,))
         data = cur.fetchone()
 
         if data is None:
-            # create a new record for the new day
-            cur.execute("INSERT INTO recordActivity VALUES "
-                        "(?, 0, 0, 0, 0, 1, STRFTIME('%Y-%m-%d %H:%M:%S', 'now', 'localtime'))", (id,))
+            # create a new record for the new day if there isnt records
+            # for the current day
+            cur.execute(
+                "INSERT INTO recordActivity "
+                "VALUES (?, 0, 0, 0, 0, 1, "
+                "STRFTIME('%Y-%m-%d %H:%M:%S', 'now', 'localtime'))", (id,))
 
         else:
             # sqlite3.Row' object does not support item assignment -> so,
@@ -348,11 +353,14 @@ def update_activity(action="server_calls"):
                 data["server_calls"] += 1
                 quota_supervisor("server_calls", update=True)
 
-            sql = ("UPDATE recordActivity set likes = ?, comments = ?, "
-                   "follows = ?, unfollows = ?, server_calls = ?, "
-                   "created = STRFTIME('%Y-%m-%d %H:%M:%S', 'now', 'localtime') "
-                   "WHERE  profile_id=? AND STRFTIME('%Y-%m-%d %H', created) == "
-                   "STRFTIME('%Y-%m-%d %H', 'now', 'localtime')")
+            sql = (
+                "UPDATE recordActivity "
+                "SET likes = ?, comments = ?, follows = ?, unfollows = ?, "
+                "server_calls = ?, "
+                "created = STRFTIME('%Y-%m-%d %H:%M:%S', 'now', 'localtime') "
+                "WHERE profile_id = ? "
+                "AND STRFTIME('%Y-%m-%d', created) == "
+                "STRFTIME('%Y-%m-%d', 'now', 'localtime')")
 
             cur.execute(sql, (data['likes'], data['comments'], data['follows'],
                               data['unfollows'], data['server_calls'], id))

### CODE REVIEW

File: instapy/util.py [hunk 1]
```diff
@@ -324,15 +324,20 @@ def update_activity(action="server_calls"):
         conn.row_factory = sqlite3.Row
         cur = conn.cursor()
         # collect today data
-        cur.execute("SELECT * FROM recordActivity WHERE profile_id=:var AND "
-                    "STRFTIME('%Y-%m-%d %H', created) == STRFTIME('%Y-%m-%d %H', 'now', 'localtime')",
-                    {"var": id})
+        # check if there is existent records for the current day
+        cur.execute("SELECT * FROM recordActivity "
+                    "WHERE profile_id = ? "
+                    "AND STRFTIME('%Y-%m-%d', created) == "
```
converge, line 330: before '%Y-%m-%d %H'
now '%Y-%m-%d'

instapy/util.py
```diff
@@ -324,15 +324,20 @@ def update_activity(action="server_calls"):
...
+                    "    STRFTIME('%Y-%m-%d','now', 'localtime')", (id,))
         data = cur.fetchone()
 
         if data is None:
-            # create a new record for the new day
-            cur.execute("INSERT INTO recordActivity VALUES "
-                        "(?, 0, 0, 0, 0, 1, STRFTIME('%Y-%m-%d %H:%M:%S', 'now', 'localtime'))", (id,))
+            # create a new record for the new day if there isnt records
```
converge, line 335: just organized it better, same code

instapy/util.py
```diff
@@ -348,11 +353,14 @@ def update_activity(action="server_calls"):
...
-                   "follows = ?, unfollows = ?, server_calls = ?, "
-                   "created = STRFTIME('%Y-%m-%d %H:%M:%S', 'now', 'localtime') "
-                   "WHERE  profile_id=? AND STRFTIME('%Y-%m-%d %H', created) == "
-                   "STRFTIME('%Y-%m-%d %H', 'now', 'localtime')")
+            sql = (
+                "UPDATE recordActivity "
+                "SET likes = ?, comments = ?, follows = ?, unfollows = ?, "
+                "server_calls = ?, "
+                "created = STRFTIME('%Y-%m-%d %H:%M:%S', 'now', 'localtime') "
+                "WHERE profile_id = ? "
```
converge: before '%Y-%m-%d %H'
now '%Y-%m-%d'

timgrossmann:
@converge Will the entry for the day be updated?
If we don't record every session, it might get hard to know when the session were.

What is the reason for this change?

an0key:
Interesting approach. This is due to a bug I raised on the dashboard, where it will only show one of the instances.

This isn't how I would have suggested fixing it, instead, this should be handled the dashboard side. It's much more useful to have the per-session statitics, and then either show all stats, or summerise on the dashboard.

converge:
from my perspective, recording hour based sessions isn't useful, because we don't have control over the sessions, for example:
```
a) 1 record for 2018-12-02 10:00:00 (with activity records)
b) 1 record for 2018-12-02 11:30:00 (with activity records)
c) 1 record for 2018-12-02 16:00:00 (with activity records)
```
We don't know if the session started at 10am stopped like 10:59am and started again at 11am
OR
the session started 10am and made the last server_call of session at 11:30am

------

My suggestion is just group the interactions by day. @timgrossmann answering the question, we're still inserting/update the hour as before.

------

A good exercise to see my suggestion working is retrieve data while the bot is running, it works as before, the only change is that it's grouping all interactions per profile_id and date(0000/00/00).

converge:
@an0key for that design, we would need to add a session key, so we would be able to track the sessions, wihout a session key we don't have an indicator to know the session.

the way we have the database design today, it's all together, we're unable to identify sessions.

let me know what you guys think.

uluQulu:
Hi all

@converge 
I have learnt about SQL by playing with your `update_activity()` function 😛 and you initially had records in the daily basis.
But when QS came out, I switched to hourly basis to be able to track the hourly statistics and _inspect_/_control_ the quota usage in both _hourly_ & _daily_ intervals.
I also thought of switching to momentary interval and offer user flexible delays (_e.g. per each 10 minutes_) but then I thought it would be a bit complex 😄 for the initial release..
So the change you are going to do will break whole Quota Supervisor 🎳

Not only QS, the `dump_record_activity()` function is also using the _hourly_/_daily_ structure of the `recordActivity` table.
_It writes a local file for the user to benefit at the end of every session_.

> ...because we don't have control over the sessions,...

I think, you can easily add a new tracker in **DB** to store session _start_ & _end_ if it is important.

@an0key 
> "_...per-session statitics_..."

Well, you can also use the `live_report()` method to view sessional live statistics, easily.

---

Also, I think DB is quite good cos I have used **referencing** unique profile IDs (_real numbers_) all around rather than hard recording profile names or IDs.
That's why it saves tons of data, e.g., while storing 24 hourly records per day rather than 1 daily record.
```erlang
`id` INTEGER PRIMARY KEY AUTOINCREMENT,
...
`profile_id` INTEGER REFERENCES `profiles` (id),
```
you can see it at **database_engine.py**.

---

Cheers 😁
---

converge:
hey @uluQulu , I totally understand your point, thank you for clarify! I didn't notice that it would reach QS, good that you showed me that. I tested it with other functions, it all worked well, but I didn't test it with QS.

I came to this solution because it's showing duplicate records and we don't have a session identifier atm, the logic way to fix it was grouping it by day, but if it will break something, I prefer to step back and rethink it.

I always like to go deep in the rabbit hole, for now I will go to the quick solution that's grouping data by date on the dashboard.

Thank you guys for the healthy discuss. 🤘

PR #3495, opened 2018-12-04
PEP8 Fixes and On Exit Fix

w4ngel:
- Better formatting in PEP8 across the board, mainly getting rid of unnecessary returns and formatting/indentations.

- Couple more self.aborts for the follow followers and follow following because otherwise InstaPy will not exit properly on them.

### DIFF
--- a/instapy/clarifai_util.py
+++ b/instapy/clarifai_util.py
@@ -6,20 +6,19 @@ from selenium.common.exceptions import NoSuchElementException
 
 
 def check_image(
-    browser,
-    clarifai_api_key,
-    img_tags,
-    img_tags_skip_if_contain,
-    logger,
-    clarifai_models,
-    workflow,
-    probability,
-    full_match=False,
-    check_video=False,
-    proxy=None,
-    picture_url=None,
+        browser,
+        clarifai_api_key,
+        img_tags,
+        img_tags_skip_if_contain,
+        logger,
+        clarifai_models,
+        workflow,
+        probability,
+        full_match=False,
+        check_video=False,
+        proxy=None,
+        picture_url=None,
 ):
-
     try:
         """Uses the link to the image to check for invalid content in the image.
         If a workflow has been selected, get list of tags from Clarifai API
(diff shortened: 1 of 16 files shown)


timgrossmann:
@w4ngel It's a good idea to move the whole codebase to PEP8, I totally agree.
We had some discussions with @uluQulu @sionking @CharlesCCC in one of the discord channels that it is pretty hard to handle since some of the PEP8 things may break some of the logics (e.g. using if condition: instead of if condition is True: since the first one is True for more cases due to coercion etc.)

We have to introduce this slowly over time since looking over the whole codebase in one run is really complex.

Thank you very much for this PR, though.

Could you create a separate PR for the second point you mentioned, introducing those fixes without the PEP8 change, please?

CharlesCCC:
@w4ngel thank you for the PR, at the PEP-8 point, I would agree with @timgrossmann . Currently, the Instapy has a huge code-base, the change will be too dramatical to handle it all.  moving forward I suggest to use it, but I'm hesitate to enforce it, as not everybody have the same level of skill and style.

w4ngel:
> @w4ngel It's a good idea to move the whole codebase to PEP8, I totally agree.
> We had some discussions with @uluQulu @sionking @CharlesCCC in one of the discord channels that it is pretty hard to handle since some of the PEP8 things may break some of the logics (e.g. using if condition: instead of if condition is True: since the first one is True for more cases due to coercion etc.)
> 
> We have to introduce this slowly over time since looking over the whole codebase in one run is really complex.
> 
> Thank you very much for this PR, though.
> 
> Could you create a separate PR for the second point you mentioned, introducing those fixes without the PEP8 change, please?

Sure will do a bit later.

w4ngel:
One thing to bear in mind all PEP8 changes here are formatting, no changes to the logic have been made.

sionking:
This merge can kill birds :()
I think one at the time is better :)

timgrossmann:
@w4ngel Thank you, good to know those are only formatting changes.

We'll introduce this over time.

Issue #3451, opened 2018-11-29
`Can't multiply sequence by non-int of type 'float'` after newest PR


oskarkrawczyk:
After pulling the newest InstaPy I get the following error – most likely caused by this #3318 PR.

- `pip install .` has been run successfully
- running `python 2.7`

```
→ python qs-likes.py
Traceback (most recent call last):
  File "qs-likes.py", line 17, in <module>
    disable_image_load=True)
  File "/Users/oskarkrawczyk/Sites/InstaPy/instapy/instapy.py", line 250, in __init__
    self.set_selenium_local_session()
  File "/Users/oskarkrawczyk/Sites/InstaPy/instapy/instapy.py", line 391, in set_selenium_local_session
    highlight_print(self.username, message, "initialization", "info", self.logger)
  File "/Users/oskarkrawczyk/Sites/InstaPy/instapy/util.py", line 929, in highlight_print
    print("\n{}".format(upper_char * ceil(output_len/len(upper_char))))
TypeError: can't multiply sequence by non-int of type 'float'
```

This is the quickstart file:

```python
import random
from instapy import InstaPy
from instapy.util import smart_run

# login credentials
insta_username = '...'
insta_password = '...'

# get an InstaPy session!
# set headless_browser=True to run InstaPy in the background
session = InstaPy(username=insta_username,
                  password=insta_password,
                  headless_browser=True,
                  disable_image_load=True)

with smart_run(session):
    """ Activity flow """

    like_tag_list = [
        'taintedmag',
        'thinkverylittle',
        'verybusymag',
        'dreamermagazine',
        'lightzine',
        'indiependentmag',
        'collecmag',
        'noicemag',
        'everybodyfilm',
        'believeinfilm',
        'theanalogclub',
        'specialfilm',
        'kodaklosers'
    ]

    session.set_simulation(enabled=False)

    session.set_skip_users(
        skip_private=True,
        private_percentage=100
    )

    session.set_quota_supervisor(
        enabled=True,
        sleep_after=["likes", "comments_d", "follows", "unfollows", "server_calls_h"],
        sleepyhead=True,
        stochastic_flow=True,
        notify_me=False,
        peak_likes=(107, 585),
        peak_comments=(31, 182),
        peak_follows=(48, None),
        peak_unfollows=(35, 402),
        peak_server_calls=(None, 4700)
    )

    session.set_relationship_bounds(
        enabled=True,
        # potency_ratio=1,
        delimit_by_numbers=True,
[...]

Issue #3503, opened 2018-12-05
quickstart failing


tariquesani:
Hi, 
Can anyone look at this traceback and let me know what is wrong.

`Traceback (most recent call last):
  File "quickstart.py", line 14, in <module>
    headless_browser=False)
  File "/home/tariquesani/Projects/instapy/venv/local/lib/python2.7/site-packages/instapy/instapy.py", line 250, in __init__
    self.set_selenium_local_session()
  File "/home/tariquesani/Projects/instapy/venv/local/lib/python2.7/site-packages/instapy/instapy.py", line 391, in set_selenium_local_session
    highlight_print(self.username, message, "initialization", "info", self.logger)
  File "/home/tariquesani/Projects/instapy/venv/local/lib/python2.7/site-packages/instapy/util.py", line 929, in highlight_print
    print("\n{}".format(upper_char * ceil(output_len/len(upper_char))))
TypeError: can't multiply sequence by non-int of type 'float'
`

octavius-1:
Had the same issue, try casting to int the result of the division, otherwise you’re going to have n.0 instead of n 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

BrianneHostutler:
I have the same issue, @octavius-1 can you elaborate on what needs to be changed?

BrianneHostutler:
Actually found the answer here: https://github.com/timgrossmann/InstaPy/issues/3451. I had to update to python 3 and pip 3 and it is working now.

tariquesani:
Using `virtualenv -p python3 venv` fix it for me. May be the instructions in README should be changed

PR #3510, opened 2018-12-05
Python 2.7 Ceil Fix

w4ngel:
Fix for python 2.7 users, ceil returns a float in python 2. 
Fixes issues:
#3451
#3503


timgrossmann:
@w4ngel Thanks for the fix. good catch there.

PR #3512, opened 2018-12-06
Updated virtualenv instructions

tariquesani:
Updated virtualenv instructions to reflect python3 requirement


tariquesani:
OK I see that the issue has been fixed and we are not being forced to use Python 3 :)

PR #3528, opened 2018-12-08
remove  api call; fix logging uncertain; some more error catching

sionking:
-- remove https://i.instagram.com/api/v1/users/{}/info/ as it not working and killing the unfollow with error
-- fix logging uncertain
-- fix get active users  when Video have no likes button / no posts in page

### DIFF
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -29,6 +29,7 @@ from .util import truncate_float
 from .print_log_writer import log_followed_pool
 from .print_log_writer import log_uncertain_unfollowed_pool
 from .print_log_writer import log_record_all_unfollowed
+from .print_log_writer import get_log_time
 from .relationship_tools import get_followers
 from .relationship_tools import get_nonfollowers
 from .database_engine import get_database
@@ -348,6 +349,9 @@ def unfollow(browser,
                                                              relationship_data,
                                                               logger,
                                                               logfolder)
+
+                    post_unfollow_actions(browser, person, logger)
+
                     if unfollow_state == True:
                         unfollowNum += 1
                         sleep_counter += 1
@@ -645,6 +649,7 @@ def get_users_through_dialog(browser,
         amount = int(users_count*0.85)
     try_again = 0
     sc_rolled = 0
+
     # find dialog box
     dialog_address = "//div[3]/div/div/div[2]"
     dialog = browser.find_element_by_xpath(dialog_address)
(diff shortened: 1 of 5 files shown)


sionking:
@timgrossmann this is urgent since unfollow can be skipped by error until it fixed.

timgrossmann:
@sionking Just saw it and looked through it. 
I don't really like the comment out tactic. But I guess it's fine for now.

Looks like it's fine. 
Could you add an CHANGELOG entry and then merge this, please?

sionking:
@timgrossmann can you check the CHANGELOG ?

timgrossmann:
@sionking Perfect, thank you very much!
Thank you for updating it.

Sorry for the delay

instapy/unfollow_util.py
```diff
@@ -1187,10 +1199,12 @@ def post_unfollow_cleanup(state, username, person, relationship_data, logger, lo
...
         # this user was found in our unfollow list but currently is not being followed
-        log_uncertain_unfollowed_pool(username, person, logger, logfolder)
+        logtime = get_log_time()
+        log_uncertain_unfollowed_pool(username, person, logger, logfolder, logtime, person_id)
         # save any unfollowed person
-        log_record_all_unfollowed(username, person, logger, logfolder)
-        sleep(3)
+
+    log_record_all_unfollowed(username, person, logger, logfolder)
+    sleep(3)
```
uluQulu, line 1207: Why did you indent it back?

**Successful** unfollows has their own custom or default sleep delay being executed in the unfollow engine.
But **uncertain** unfollows take that static 3 seconds of sleep at this `post_unfollow_cleanup()` function.

Now it will misbehave with 3 more extra sleep seconds per each successful unfollow...

instapy/unfollow_util.py
```diff
@@ -1306,3 +1320,9 @@ def verify_action(browser, action, track, username, person, person_id, logger, l
                 return False, "unexpected"
 
     return True, "success"
+
+
+def post_unfollow_actions(browser, person, logger):
```
uluQulu, line 1325: What is the use of this deserted function?

instapy/unfollow_util.py
sionking: I should only indent the log_record_all_unfollowed correct.

sionking: I own you 3 seconds :)

instapy/unfollow_util.py
sionking: future.
I am using it for some of my tools but it might be used in future.

PR #3529, opened 2018-12-08
unix file name style for template files

converge:
Rename template files for easier use on *nix terminals.


timgrossmann:
@converge Good one, thank you very much!

PR #3540, opened 2018-12-09
feature/log_liked_and_commented_posts

artem-panin:
Feature: logging of liked or commented users and posts. 
Logs template: logtime ~ liked_user(commented_user) ~ liked_post(commented_post). 
It will be useful for some metrics, like retention rate of users after likes or comments.


sionking:
maybe it can be merged with SOME PR already logging each action

artem-panin:
> maybe it can be merged with SOME PR already logging each action

I suppose it isn't necessary to log every action, but it can be useful to log some activities like comments or likes. Follows and unfollows already logged and saved in logs directory, so it is convenient to make your own follow/unfollow strategies. For me, likes and comments are important parts of activities and very important too for follow/unfollow strategies and I had been using it for a weeks. Do you agree that it will be useful to save not just followed and unfollowed accounts, but liked and commented posts too?

timgrossmann:
@artem-panin Please change your credentials now.
Also, you've committed several ipynb files, those have to be removed, too.

PR #3543, opened 2018-12-10
Fix dont_include override after set a blacklist

gdurandvadas:
### Scenario
`dont_include` is overwrite when `set_blacklist` is enabled, but never recovered after `set_blacklist` disabled.
With some item in `blacklist` do:
```phython
bot = InstaPy(username=insta_username, password=insta_password)
with smart_run(bot):
  bot.set_dont_include(friendlist)
  print(bot.dont_include)
  bot.set_blacklist(enabled=True, campaign='blacklist')
  print(bot.dont_include)
  bot.set_blacklist(enabled=False, campaign='blacklist')
  print(bot.dont_include)
```


timgrossmann:
@gdurandvadas That's an excellent catch!
Thank you very much for fixing it.

PR #3637, opened 2018-12-27
PEP8 layout changes

w4ngel:
This PR deals with pushing InstaPy to a cleaner code base with PEP8 as in the roadmap, a step towards the PEP8 format is to do spacing correctly, have the correct amount of blank lines after a function, at the end of the file etc, having spaces after the start of comments. This PR only deals with that, it does not attempt to do anything outside of formatting for PEP8 to give a much more readable codebase.

All files edited are done as separate commits. No changes par the ones stated will be in this branch.

### DIFF
--- a/instapy/browser.py
+++ b/instapy/browser.py
@@ -63,7 +63,6 @@ def set_selenium_local_session(proxy_address,
     else:
         chromedriver_location = Settings.chromedriver_location
         chrome_options = Options()
-        # chrome_options.add_argument("--disable-infobars")
         chrome_options.add_argument("--mute-audio")
         chrome_options.add_argument('--dns-prefetch-disable')
         chrome_options.add_argument('--lang=en-US')
--- a/instapy/browser.py
+++ b/instapy/browser.py
@@ -76,7 +76,8 @@ def set_selenium_local_session(proxy_address,
             chrome_options.add_argument('--no-sandbox')
 
             if disable_image_load:
-                chrome_options.add_argument('--blink-settings=imagesEnabled=false')
+                chrome_options.add_argument(
+                    '--blink-settings=imagesEnabled=false')
 
             # replaces browser User Agent from "HeadlessChrome".
             user_agent = "Chrome"
@@ -101,7 +102,8 @@ def set_selenium_local_session(proxy_address,
 
         # using saved profile for chrome
         if browser_profile_path is not None:
-            chrome_options.add_argument('user-data-dir={}'.format(browser_profile_path))
+            chrome_options.add_argument(
+                'user-data-dir={}'.format(browser_profile_path))
 
         chrome_prefs = {
             'intl.accept_languages': 'en-US',
@@ -124,7 +126,8 @@ def set_selenium_local_session(proxy_address,
 
         # prevent: Message: unknown error: call function result missing 'value'
         matches = re.match(r'^(\d+\.\d+)',
-                           browser.capabilities['chrome']['chromedriverVersion'])
+                           browser.capabilities['chrome'][
+                               'chromedriverVersion'])
         if float(matches.groups()[0]) < Settings.chromedriver_min_version:
             err_msg = 'chromedriver {} is not supported, expects {}+'.format(
                 float(matches.groups()[0]), Settings.chromedriver_min_version)
--- a/instapy/browser.py
+++ b/instapy/browser.py
@@ -63,7 +63,7 @@ def set_selenium_local_session(proxy_address,
     else:
         chromedriver_location = Settings.chromedriver_location
         chrome_options = Options()
-        #chrome_options.add_argument("--disable-infobars")
+        # chrome_options.add_argument("--disable-infobars")
         chrome_options.add_argument("--mute-audio")
         chrome_options.add_argument('--dns-prefetch-disable')
         chrome_options.add_argument('--lang=en-US')
@@ -113,13 +113,13 @@ def set_selenium_local_session(proxy_address,
         chrome_options.add_experimental_option('prefs', chrome_prefs)
         try:
             browser = webdriver.Chrome(chromedriver_location,
-                                            desired_capabilities=capabilities,
-                                            chrome_options=chrome_options)
+                                       desired_capabilities=capabilities,
+                                       chrome_options=chrome_options)
 
         except WebDriverException as exc:
             logger.exception(exc)
             err_msg = 'ensure chromedriver is installed at {}'.format(
-                            Settings.chromedriver_location)
+                Settings.chromedriver_location)
             return browser, err_msg
 
         # prevent: Message: unknown error: call function result missing 'value'
@@ -127,7 +127,7 @@ def set_selenium_local_session(proxy_address,
                            browser.capabilities['chrome']['chromedriverVersion'])
         if float(matches.groups()[0]) < Settings.chromedriver_min_version:
             err_msg = 'chromedriver {} is not supported, expects {}+'.format(
-                            float(matches.groups()[0]), Settings.chromedriver_min_version)
+                float(matches.groups()[0]), Settings.chromedriver_min_version)
             return browser, err_msg
 
     browser.implicitly_wait(page_delay)
--- a/instapy/browser.py
+++ b/instapy/browser.py
@@ -185,7 +185,7 @@ def close_browser(browser,
                 logger.exception(
                     "Error occurred while deleting cookies "
                     "from web browser!\n\t{}"
-                        .format(str(exc).encode("utf-8")))
+                    .format(str(exc).encode("utf-8")))
 
         # close web browser
         try:
@@ -195,4 +195,4 @@ def close_browser(browser,
                 logger.exception(
                     "Error occurred while "
                     "closing web browser!\n\t{}"
-                        .format(str(exc).encode("utf-8")))
+                    .format(str(exc).encode("utf-8")))
--- a/instapy/clarifai_util.py
+++ b/instapy/clarifai_util.py
@@ -5,20 +5,18 @@ from clarifai.rest import Workflow
 from selenium.common.exceptions import NoSuchElementException
 
 
-def check_image(
-        browser,
-        clarifai_api_key,
-        img_tags,
-        img_tags_skip_if_contain,
-        logger,
-        clarifai_models,
-        workflow,
-        probability,
-        full_match=False,
-        check_video=False,
-        proxy=None,
-        picture_url=None,
-):
+def check_image(browser,
+                clarifai_api_key,
+                img_tags,
+                img_tags_skip_if_contain,
+                logger,
+                clarifai_models,
+                workflow,
+                probability,
+                full_match=False,
+                check_video=False,
+                proxy=None,
+                picture_url=None):
     try:
         """Uses the link to the image to check for invalid content in the 
         image.
--- a/instapy/clarifai_util.py
+++ b/instapy/clarifai_util.py
@@ -18,7 +18,7 @@ def check_image(browser,
                 proxy=None,
                 picture_url=None):
     try:
-        """Uses the link to the image to check for invalid content in the 
+        """Uses the link to the image to check for invalid content in the
         image.
         If a workflow has been selected, get list of tags from Clarifai API
         by checking link against models included in the workflow. If a workflow
--- a/instapy/clarifai_util.py
+++ b/instapy/clarifai_util.py
@@ -20,16 +20,19 @@ def check_image(
         picture_url=None,
 ):
     try:
-        """Uses the link to the image to check for invalid content in the image.
+        """Uses the link to the image to check for invalid content in the 
+        image.
         If a workflow has been selected, get list of tags from Clarifai API
         by checking link against models included in the workflow. If a workflow
-        hasn't been provided, InstaPy will check images against given model(s)"""
+        hasn't been provided, InstaPy will check images against given model(
+        s)"""
         clarifai_api = ClarifaiApp(api_key=clarifai_api_key)
         clarifai_tags = []
 
         if proxy is not None:
             clarifai_api.api.session.proxies = {'https': proxy}
-        # Set req image or video source URL to given one or get it from current page
+        # Set req image or video source URL to given one or get it from
+        # current page
         if picture_url is None:
             source_link = get_source_link(browser)
         else:
@@ -39,14 +42,19 @@ def check_image(
         if not source_link:
             return True, [], []
 
-        # Check image using workflow if provided. If no workflow, check image using model(s)
+        # Check image using workflow if provided. If no workflow,
+        # check image using model(s)
         if workflow:
-            clarifai_workflow = Workflow(clarifai_api.api, workflow_id=workflow[0])
-            # If source is video, checks keyframe against models as video inputs not supported when using workflows
+            clarifai_workflow = Workflow(clarifai_api.api,
+                                         workflow_id=workflow[0])
+            # If source is video, checks keyframe against models as video
+            # inputs not supported when using workflows
             if source_link[0].endswith('mp4'):
-                clarifai_response = clarifai_workflow.predict_by_url(source_link[1])
+                clarifai_response = clarifai_workflow.predict_by_url(
+                    source_link[1])
             else:
-                clarifai_response = clarifai_workflow.predict_by_url(source_link[0])
+                clarifai_response = clarifai_workflow.predict_by_url(
+                    source_link[0])
 
             for response in clarifai_response['results'][0]['outputs']:
                 results = get_clarifai_tags(response, probability)
@@ -67,17 +75,20 @@ def check_image(
             )
         )
 
-        # Will not comment on an image if any of the tags in img_tags_skip_if_contain are matched
+        # Will not comment on an image if any of the tags in
+        # img_tags_skip_if_contain are matched
         if given_tags_in_result(img_tags_skip_if_contain, clarifai_tags):
             logger.info(
                 'Not Commenting, image contains concept(s): "{}".'.format(
-                    ', '.join(list(set(clarifai_tags) & set(img_tags_skip_if_contain)))
+                    ', '.join(list(
+                        set(clarifai_tags) & set(img_tags_skip_if_contain)))
                 )
             )
             return False, [], clarifai_tags
 
         for (tags, should_comment, comments) in img_tags:
-            if should_comment and given_tags_in_result(tags, clarifai_tags, full_match):
+            if should_comment and given_tags_in_result(tags, clarifai_tags,
+                                                       full_match):
                 return True, comments, clarifai_tags
             elif given_tags_in_result(tags, clarifai_tags, full_match):
                 logger.info(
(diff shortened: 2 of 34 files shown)


w4ngel:
Just to reiterate, this does not change any logic, just format cleans up the code to match PEP8 formatting guidelines.

timgrossmann:
@w4ngel Thank you for clarifying, we still have to discuss if we should do all the changes in one step or introduce it one after each other.
We had PRs like this before,  I have to admit I really like this separation of refactors per commit. 

Thank you for this, let me discuss this with the other core devs.

w4ngel:
> @w4ngel Thank you for clarifying, we still have to discuss if we should do all the changes in one step or introduce it one after each other.
> We had PRs like this before, I have to admit I really like this separation of refactors per commit.
> 
> Thank you for this, let me discuss this with the other core devs.

Yeah just let me know, if you ever want to get closer to PEP8, honestly going one step at a time won't be the way as the code base is so extensive, at least a massive formatting change like this gets rid of one of the large parts of PEP8 which is merely simple layout. The more complex parts like the logic changes should be done step by step as things like `if condition:` aren't necessarily applicable to all functions in InstaPy.

uluQulu:
Nice job, @w4ngel!

Just wonder, have you used auto formatter for that?

Also, as @timgrossmann mentioned, file-specific commits makes it too much easier to differentiate.
Will try to look at the changes in depth, soon as possible.

--- 

Cheers :grin: 
---

w4ngel:
> Nice job, @w4ngel!
> 
> Just wonder, have you used auto formatter for that?
> 
> Also, as @timgrossmann mentioned, file-specific commits makes it too much easier to differentiate.
> Will try to look at the changes in depth, soon as possible.
> 
> ## Cheers 😁

We have a few auto format tools for work that a colleague made, they go to varying degrees of detail, but the best one to use just does formats it in PEP8 layout :)

Also use the pep8.py tool to flag issues.

w4ngel:
The last few commits are just similar changes for all the quickstart files, proxy extension files and setup. So the changes would be uniform across the whole project now :D

timgrossmann:
@w4ngel @uluQulu Alright, I've looked though every commit to make sure that there are no logical changes (or malicious additions. Nothing personal @w4ngel just making sure 🙈) 

Everything looks good and much cleaner now, thank you for this PR.

As soon as @uluQulu approves this PR, we should be ready to go. 
Considering that @sionking is also fine with the changes made here, I know he has a lot of local changes to his InstaPy version.

timgrossmann (approved):

w4ngel:
> @w4ngel @uluQulu Alright, I've looked though every commit to make sure that there are no logical changes (or malicious additions. Nothing personal @w4ngel just making sure 🙈)
> 
> Everything looks good and much cleaner now, thank you for this PR.
> 
> As soon as @uluQulu approves this PR, we should be ready to go.
> Considering that @sionking is also fine with the changes made here, I know he has a lot of local changes to his InstaPy version.

Any changes that @sionking has I can also format appropriately as and when they are committed.

w4ngel:
Resolved Merge conflicts. Can I suggest this is merged sooner rather than later so I won't have to keep resolving them?

timgrossmann:
@w4ngel If no-one shouts out with a big no, I'll merge it this evening (8pm German time)
Thank you for resolving

instapy/text_analytics.py
```diff
@@ -120,14 +117,14 @@ def text_analysis(text, text_type, logger):
                 des_pol = MEANINGCLOUD_CONFIG["score_tag"]
 
                 polarity_level = (3 if pol == "P+" else 2 if pol == 'P' else
-                                  1 if pol == "NEU"else
```
uluQulu, line 123: a bit strange here

w4ngel: The way that is done is PEP8 visual under indentation- believe it or not the previous version follows PEP8 to the dot but is harder to read. I can change it back if you want?

instapy/unfollow_util.py
```diff
@@ -315,18 +309,21 @@ def unfollow(browser,
...
                     logger.warning("--> Unfollow quotient reached its peak!\t~leaving Unfollow-Users activity\n")
                     break
 
                 if sleep_counter >= sleep_after and sleep_delay not in [0, None]:
-                    delay_random = random.randint(ceil(sleep_delay*0.85), ceil(sleep_delay*1.14))
+                    delay_random = random.randint(ceil(sleep_delay * 0.85), ceil(sleep_delay * 1.14))
                     logger.info("Unfollowed {} new users  ~sleeping about {}\n".format(sleep_counter,
-                                    '{} seconds'.format(delay_random) if delay_random < 60 else
-                                    '{} minutes'.format(truncate_float(delay_random/60, 2))))
+                                                                                       '{} seconds'.format(
```
uluQulu: a lot of spacing here,,

w4ngel: Yeah, I believe it's these if statements that are not necessarily following PEP8 convention themselves causing the format to do some strange things.

instapy/util.py
```diff
@@ -1287,7 +1287,7 @@ def explicit_wait(browser, track, ec_params, logger, timeout=35, notify=True):
     elif track == "PFL":
         ec_name = "page fully loaded"
         condition = (lambda browser: browser.execute_script("return document.readyState")
-                     in ["complete" or "loaded"])
+                                     in ["complete" or "loaded"])
```
uluQulu, line 1452: ow 16 spaces..

w4ngel: > ow 16 spaces..

Yeah that's due to it being inline with the browser.execute_script statement as that's the statement that it's part of.

uluQulu:
I have a busy schedule but spent a few minutes to review this PR in details.

I have checked whole code and else than those 3 comments, everything is amazing.
_Those comments are not necesary but just for good visual look_..

@w4ngel, if your auto formatter makes it look ugly either by PEP 8 rules or not, it must be changed.
Remember, PEP 8 is just for making code look good, readable,..

BTW if you said all of the changes here were ONLY **spacing** and **lining**, then it would be much cooler to review. I know you said but there you say "_etc._" .. 🙂

It is difficult to prepare such a huge change list, so, you have done a really good job.
I use `AutoPEP8` in my _Sublime Text_, but I don't think I can configure it to just pepize spacing and lining, you got a nice tool 😎 there..

Also, how many spaces did you take as the line length limit?

---

Cheers 😁
---

w4ngel:
> I have a busy schedule but spent a few minutes to review this PR in details.
> 
> I have checked whole code and else than those 3 comments, everything is amazing.
> _Those comments are not necesary but just for good visual look_..
> 
> @w4ngel, if your auto formatter makes it look ugly either by PEP 8 rules or not, it must be changed.
> Remember, PEP 8 is just for making code look good, readable,..
> 
> BTW if you said all of the changes here were ONLY **spacing** and **lining**, then it would be much cooler to review. I know you said but there you say "_etc._" .. 🙂
> 
> It is difficult to prepare such a huge change list, so, you have done a really good job.
> I use `AutoPEP8` in my _Sublime Text_, but I don't think I can configure it to just pepize spacing and lining, you got a nice tool 😎 there..
> 
> Also, how many spaces did you take as the line length limit?
> 
> ## Cheers 😁

Gotcha, and 121 is the line length limit I have on my current laptop.

w4ngel:
@timgrossmann let me know if you want me to make any changes before merge!

timgrossmann:
@w4ngel Just making sure that this really is standard PEP8 without any highly modified custom styling.
It will be hard for people to contribute if this is something you configured without sharing your linter config

uluQulu:
@w4ngel 
https://pep8.org/#maximum-line-length
 - "_Limit all lines to a maximum of 79 characters._"

It explains clearly there and I now, solely, agree with those rules if not earlier times.

timgrossmann:
@w4ngel Thank you very much @uluQulu for referencing.

I think this PR should stay open a little bit longer. I won't merge anything that is not crucial to bug fixing until we've merged this.

@w4ngel Could you reason about the changes you have done to the PEP8 standard definition and why you adjusted them.

w4ngel:
> @w4ngel Just making sure that this really is standard PEP8 without any highly modified custom styling.
> It will be hard for people to contribute if this is something you configured without sharing your linter config

The linter is using standard definition library for PEP8 without optimisations/code cleanup options.



> Could you reason about the changes you have done to the PEP8 standard definition and why you adjusted them.

The only thing that I have customised is line length but that is set auto to my screen size! Which if I reset, I'll just need to make some adjustments to the merge to make it fit. Certain stuff hasn't been formatted, mainly if you have a string such as `foo = "Some very long string"` it will not split up the string to multiple lines. Other than that no problem.

I am just gonna do a recheck of all the files with the PEP8 line length and do a re-review.

w4ngel:
Just a note, biggest things flagged on making the lines shorter are the following;

1. Variable and parameter naming conventions should all be lower case, in some instances they aren't - this hasn't been changed in this commit.

2. Lines too long - as mentioned above, can be fixed but would suggest a staged approach for this unless it's obvious. Edit; I can fix this easily.

3. And last; continuation lines missing indentation or being outdented, i.e. due to the above, some strings that are formatted and put over multiple lines then get weird auto-indentation. 

@timgrossmann @uluQulu If you want I can reformat with line limit of 79, will need to go through to solve issue 3, but I can do that easily. Just need to know before I do it.

Edit; just so you know I use the PEP8.py tool to flag issues.

w4ngel:
Here is an example of what things would look like with corrected issues above (Par the naming change): https://gist.github.com/w4ngel/117bb223c398b2ba9d89379f9955cf65

timgrossmann:
@w4ngel Thank you very much for the quick response.
You're absolutely right, we need to, iteratively, adjust the naming and too long lines.

I think the line length is hard to decide, in my opinion, 80 chars is a little bit short, but it's PEP standard and it's probably best to keep everything at default so people won't have to adjust the definition when they are creating PRs.

> Thanks for mentioning your way of listing. We will link that in the contributors guide once we're done with this.

w4ngel:
> @w4ngel Thank you very much for the quick response.
> You're absolutely right, we need to, iteratively, adjust the naming and too long lines.
> 
> I think the line length is hard to decide, in my opinion, 80 chars is a little bit short, but it's PEP standard and it's probably best to keep everything at default so people won't have to adjust the definition when they are creating PRs.
> 
> > Thanks for mentioning your way of listing. We will link that in the contributors guide once we're done with this.

If you like the look of the example file given above, I will set about making those changes across the board, I do agree with having it as a default. But I also agree that it's really short! If you have a monitor larger than 13in it looks ridiculous :D

uluQulu:
>1-) Variable and parameter naming conventions should all be lower case. 

I agree that it is a usual regulation.
But having constant variables upper cased is better.
And there are very less parameters with uppercase letter in its name and one of them is
```pyhon
interruption_handler(SIG_type=signal.SIGINT, handler=signal.SIG_IGN, notify=None, logger=None)
```
That is intentional for technical [not logical] reasons and should not be renamed.

If you try to lowercase names, it wouldn't be not be a trivial change.

---

@timgrossmann @w4ngel 
For example, there are lots of companies with every programmer having wide-screen displays and they are clearly right to choose 100 ll. But it's different here, e.g. my display is `1366x768` and such as, while opening multi tabs in a TE/IDE, and can't see longer lines fullly.
Majority of userbase should be taken into account for this.

---

Cheers 😁
---

w4ngel:
> @timgrossmann @w4ngel
> For example, there are lots of companies with every programmer having wide-screen displays and they are clearly right to choose 100 ll. But it's different here, e.g. my display is `1366x768` and such as, while opening multi tabs in a TE/IDE, and can't see longer lines fullly.
> Majority of userbase should be taken into account for this.

Ok, with that being said, @timgrossmann do you want me to go about reformatting with the PEP8 line standard? Will be able to get it done tomorrow.

timgrossmann:
@w4ngel @uluQulu I think we should try sticking to PEP8 as close as possible.
I'm a little bit worried about the sometimes really long xPaths. We don't want any line breaks in the urls themselves.

Please feel free to adjust this to the actual PEP8 line length. 
Thank you very much
Sorry for the long discussion

w4ngel:
> @w4ngel @uluQulu I think we should try sticking to PEP8 as close as possible.
> I'm a little bit worried about the sometimes really long xPaths. We don't want any line breaks in the urls themselves.
> 
> Please feel free to adjust this to the actual PEP8 line length.
> Thank you very much
> Sorry for the long discussion

I will commit changes to the actual PEP8 Line length, though I would suggest another review as things like XPaths/URLs can now be over multiple lines.

instapy/browser.py
```diff
@@ -63,7 +63,7 @@ def set_selenium_local_session(proxy_address,
     else:
         chromedriver_location = Settings.chromedriver_location
         chrome_options = Options()
-        #chrome_options.add_argument("--disable-infobars")
+        # chrome_options.add_argument("--disable-infobars")
```
converge: if not used, we can remove it

instapy/clarifai_util.py
```diff
@@ -6,31 +6,33 @@
 
 
 def check_image(
```
converge: the old version is correct..
in this way we're adding 8 tabs indentation.

instapy/database_engine.py
```diff
@@ -74,7 +73,7 @@ def create_database(address, logger, name):
     except Exception as exc:
         logger.warning(
             "Wah! Error occurred while getting a DB for '{}':\n\t{}"
-            .format(name, str(exc).encode("utf-8")))
+                .format(name, str(exc).encode("utf-8")))
```
converge: the old version is correct..

instapy/database_engine.py
```diff
@@ -132,7 +131,7 @@ def get_profile(name, address, logger):
     except Exception as exc:
         logger.warning(
             "Heeh! Error occurred while getting a DB profile for '{}':\n\t{}"
-            .format(name, str(exc).encode("utf-8")))
+                .format(name, str(exc).encode("utf-8")))
```
converge: the old version is correct..

quickstart_templates/friends_last_post_likes_and_interact_with_user_based_on_hashtahs.py
```diff
@@ -58,87 +62,113 @@
...
 if data['friends_interaction']:
-  friends = InstaPy(username=insta_username, password=insta_password, selenium_local_session=False, disable_image_load=True, multi_logs=False)
-  friends.set_selenium_remote_session(selenium_url='http://selenium:4444/wd/hub')
-  with smart_run(friends):
-    print(u'💞 Showing friends some love 💖')
-    friends.set_relationship_bounds(enabled=False)
-    friends.set_skip_users(skip_private=False)
-    friends.set_do_like(True, percentage=100)
-    friends.interact_by_users(friendlist, amount=2, randomize=False)
+    friends = InstaPy(username=insta_username, password=insta_password,
```
converge: one argument per line is better

converge (changes requested):
I couldn't review all the code, but added some quick comments.

instapy/clarifai_util.py
w4ngel: Can I please have a line ref? The changes with this file have all been to fit with the 79 char limit.

w4ngel: Wait I see.

w4ngel: Made it fit much better and corrected additonal comma that was there for some reason...

converge:
we can add E712 (PEP8) to travis setup later.

E712 (^) | comparison to True should be ‘if cond is True:’ or ‘if cond:’

converge (approved):

timgrossmann:
Love the team work here!
Thank you very much for this, guys! ✌️

uluQulu:
>           we can add E712 (PEP8) to travis setup later.
> E712 (^) | comparison to True should be ‘if cond is True:’ or ‘if cond:’

@converge 
We have discussed it several times with @sionking and @timgrossmann before.
It will break logic and code. If you want it, you must change behaviours of affected functions which changes the way they work.
I am against that regulation on this project.


---

Cheers 😁
---

converge:
@uluQulu can you show me an example ?

uluQulu:
sure @converge - such as `verify_action()` function;

If you wanna find more code that relies on multi-type return values without reading whole source code line by line, then run this short code which I just wrote to find an example for you,
```python
target_file = "unfollow_util.py"
k = ["def ", "return ", "if ", "elif ", "while ", "True", "False", "None"]

with open(target_file, 'r') as f:
    c = f.readlines()

for e, l in enumerate(c):
    if "def " in l:
        print('\n')

    if any(i in l for i in k):
        print("{}\t{}".format(e, l))
```


---

Cheers 😁
---

converge:
I see, but isn't a real situation, I can't see the point of parsing a .py file to check if the reserved word is inside the .py file.

Can you point it where will it break in InstaPy ?

uluQulu:
Sorry for a very late reply, @converge 
I have been very busy last days.

---

>"_Can you point it where will it break in InstaPy ?_"


For example, in the function- `verify_action()` that I talked about before,
image
`action_state` variable takes either a _Boolean_ or `None` type value.
And checking `action_state` as _PEP_ advices,
`if not action_state:`
will trigger both _Boolean_ `False` and `None` type values which will cause a misbehaviour.


---

>".. _can't see the point of parsing a .py file_ .. "

By printing those key lines, you can easily find out which statements rely on multi-type values.
Sorry for my basic techniques, I have not much tooling experience 😊

---

Cheers 😁
---

converge:
I see @uluQulu, the point is.. in the end, it's just the same verification with the same result, it's just a "PEP8 thing".

```python
x = None
if x is True:
    print('ok')

if x is False:
    print('ok')

if x == False:
    print('ok')

if x == True:
    print('ok')
# ---
if x is None:
    print('ok')
output: ok
if x == None:
    print('ok')
output: ok
```

Does it make sense.. ?

timgrossmann:
@converge Ahh, just checked that and I think you're right.

I think the point was the assumption that python coerces `None` to `False` when checking for `is False`.
So the assumption was this output
```python
if None is False:
    print('ok')
# will print ok

if False is False:
    print('ok')
# will print ok
```

When testing this in the console, the **actual** output is
```python
if None is False:
    print('ok')
# will NOT print ok

if False is False:
    print('ok')
# will print ok
```

So implementing PEP8 with the structural changes should not make a difference

sionking:
@w4ngel hi what tool did you use to align the whole code ?

w4ngel:
I use a custom linter that a colleague built in my IDE - PyCharm it uses pep8.py tool

uluQulu:
@converge @timgrossmann 

Sorry for delay, but it is a very sensitive logic in Python, so I could not write easily.

**1**-) @converge your tool that says is checking PEP8 is misleading cos actually PEP8 [as of today] highly disgraces that style of usage.
image
What PEP8 actually says is,
image

It's the same with my linter - **pycodestyle** (_latest version - 2.5.0 [as of today]_).
image
I was shocked by it once learnt, I will look for another linter now.

So, why PEP8 does not like `'is'` to compare booleans?
There are several reasons I am unaware of. But it is fact that `'is'` must be used only for checking `None`.

Fine, what is `'is'`?
`'is'` is not checking the value itself. It checks for identity.
```python
if zx is True:
```
is roughly equal to
```python
if id(zx) == id(True)
```

Drawbacks of using `'=='` to compare bools:
There is 1 case there it behaves differently;
```python
1 == True
0 == False
```
Yes, it's cos True and False are themselves integers...
That's the **only** caveat to be cautious while using direct comparisons on Booleans in python.


Further,
 Before booleans became part of Python they were often defined as part of a program. Even now (_Python `2.7.15`_) they are only names that can be pointed to different objects:
```python
True 👉🏼 <type 'bool'>
True = 1 👉🏼 <type 'int'>
None = 1 👉🏼 SyntaxError: cannot assign to None
```
If so,
```python
>>> True = 1
>>> (2 > 1)
True
>>> (2 > 1) == True
True
>>> (2 > 1) is True
False
```

If you try to assign a value to `True` in a newer Python e.g. `3.7.1`,
```python
True = 1 👉🏼 SyntaxError: can't assign to keyword
None = 1 👉🏼 SyntaxError: can't assign to keyword
```

#### So, shortly.
 - Booleans **must** be checked/compared as `if cond:` as PEP8 rules.
 - It isn't a strict rule, we can compare it differently UNLESS fully covered.

To have the `if cond:` style applied to our code base:
As I have previously said, bunch of blocks, their logic has to be converted into the single-type return values.
Or _type_+`None` pair.
RATHER THAN multi-type return values.


#### Conclusion
Why there are lots of that situations?
 - I can say, most of them are written by me xD (_sorry for that :P_)
 - I have used multi-type return values JUST FOR the simplicity and flexibility.
 - I really think the future of the programming is going towards [working with] multi-type return values [EVEN THOUGH many masters might disagree].
 - People think multi-type return values are unsafe and dangerous.
 - I agree with people but I have taken care of all of the possible situations.
 - It's just a bit difficult to understand for some programmers when your function returns multi-type values and if not fully covered, it can misbehave.

### Also
You should also know the confusion about the identity of big integers ..
```python
256 is 256
True

256 == 256
True

256 is (255+1)
True

256 == (255+1)
True

257 is (256+1)
False

257 == (256+1)
True
```
Yeah,
```python
257 is (256+1)
False
```
is cos of the identity of `257` is different than `256+1`, you can print it,
```python
id(257)
id(256+1)
```
>Note:
Newer Python like `3.7.1` does not have this caveat but it still generates different identities..

### Also 2

```python
>>> for i in range(250, 260): a = i; print("{}: {}".format(i, a is int(str(i))));
... 
250: True
251: True
252: True
253: True
254: True
255: True
256: True
257: False
258: False
259: False
```
This is another obvious reason not to use `'is'`: the behavior is left up to implementations when you're erroneously using it for _value equality_.

---
I will continue to use direct comparison operator `==` to check for value equality once relying on multi-type return values.

Cheers 😁
---

<sub>Special thanks to the guys at Freenode-IRC #python and threads at SO.</sub>

converge:
alright, PEP8 is our master, let's use if cond: in case cond is True/False. :D

is there some PEP code to force it? so we can add it to travis.yml

PR #3638, opened 2018-12-28
How to Raspberry with Python 2.7 & Python 3

JoeFertig:
How to Raspberry with Python 2.7 & Python 3


timgrossmann:
@JoeFertig Awesome, thank you very much for updating and extending the documentation!
This will help a lot of users that are working with Raspberries.

PR #3641, opened 2018-12-29
Update Quickstart Readme

w4ngel:
Simple readme update for the quickstart templates as some of the limits in some of the templates are quite dangerous to have depending on various factors.


timgrossmann:
@w4ngel Good point, thank you for pointing this out and adding it to the readme.

Issue #3602, opened 2018-12-16
Unfollowers not working correctly


notahooper15:
## Expected Behavior
Unfollow 400 people then quit
## Current Behavior
Unfollowers anywhere from 20-150 then quits

## InstaPy configuration

this is in my quickstart.py file

session.unfollow_users(amount=400, InstapyFollowed=(True, "all"), style="RANDOM", unfollow_after=48*60*60, sleep_delay=600)

getting the error code

Unfollow loop error:  Expecting value: line 1 column 1 (char 0)

Randomly stops unfollowing after unfollowing for some time. Any help would be appreciated thanks.

matthesv:
Change it to unfollow_after=48*60*60

Tr1pke:
i use 
48*60*60
, maybe this helps?

Tr1pke:
update you unfollow_util.py file , or download it again, i did it and no problems anymore (for now at least)

Issue #3523, opened 2018-12-08
unfollow users


wrhsescp:
## InstaPy configuration

all_unfollowers, active_unfollowers = session.pick_unfollowers(username="XXXXX",compare_by="earliest", compare_track="first", live_match=True, store_locally=True, print_out=True)
sleep(200)
session.unfollow_users(amount=len(active_unfollowers), customList=(True, active_unfollowers, "all"), style="RANDOM", unfollow_after=None, sleep_delay=600

# after picking the data and sleeping i get this : 
 Pow! Error occurred while dumping follow restriction data to a local JSON:
        b'Expecting value: line 1 column 1 (char 0)'

sionking:
I do also

sionking:
```
WARNING [2018-12-08 16:01:05] [123]  Maybe 'b'shoshykogman'' has changed the username!    ~verifying through the user ID
INFO [2018-12-08 16:01:05] [123]  Trying to find the username from the given user ID by loading a post
INFO [2018-12-08 16:01:16] [123]  Failed to get the username from 'https://www.instagram.com/p/BrGmmU_HVQK/' page
INFO [2018-12-08 16:01:16] [123]  Trying to find the username from the given user ID by a quick API call
ERROR [2018-12-08 16:01:17] [123]  Unfollow loop error:  Expecting value: line 1 column 1 (char 0)

INFO [2018-12-08 16:01:17] [123]  --> Total people unfollowed : 0

```

sionking:
@uluQulu @CharlesCCC  as I thought so any non-human call is a fail to the system and can be worst as a way to detect bot.
the "https://i.instagram.com/api/v1/users/198566306/info/" call is not working anymore.

Moreover it log you out.... isn't it was the https://www.instagram.com/web/friendships/249229606/follow/
that logged you out ?

What is the solution ?

sionking:
I have a solution:
1. save all users we can't get username to a file (we have it in a form of uncertain unfollow list).
2. once in a week grab all following of running user (sync).
3. compare id from file == id from sync, if id match - > the username is in the sync.

CharlesCCC:
@sionking Yea, you are right, i believe the second link is the one kick you out. but apparently now is the reverse. (I just tried both). I guess Instagram is playing with their URL APIs

As for your solution, how does that gonna solve the problem, what happen if the id didn't match ?

uluQulu:
@sionking 
It rather a workaround than a solution.

[...]

PR #3643, opened 2018-12-29
Catch exceptions for unfollow_user inside loop

mariusglauer:
Prevents crashing when unfollowing fails for a certain user

Should provide a workaround for
#3602
#3523

### DIFF
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -340,7 +340,8 @@ def unfollow(browser,
                     person_id = (automatedFollowedPool["all"][person]["id"] if
                                 person in automatedFollowedPool["all"].keys() else False)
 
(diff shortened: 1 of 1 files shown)


timgrossmann:
@gbhrdt Good one.
Thank you very much and congrats to your first PR to InstaPy 🙌

sionking:
Is it just me ? if exception occuer the `unfollow_state, msg`  will not be internalized and another exception will be.

timgrossmann:
@sionking Sorry, could you elaborate a bit more, please?

sionking:
@timgrossmann  We are using `unfollow_state` after we catch the exception, but if we have exception this var is not exist ...

timgrossmann:
@sionking Now I understand, meaning this PR will lead to another error once an exception is encountered

PR #3659, opened 2019-01-01
fix time_stamp will be last follow profile

sionking:
in case:
time_stamp, user
user

the user without timestamp will use the timestamp of privious user.


timgrossmann:
@sionking Sorry for the late response, this is a good fix.
Thank you very much

Issue #3656, opened 2018-12-31
Firefox + geckodriver doesn't really work


oskarkrawczyk:
I had some issues with `chromedriver` so tried to use `geckodriver`, unfortunately that doesn't work at all:

```zsh
Traceback (most recent call last):
  File "qs-follow.py", line 17, in <module>
    with smart_run(session):
  File "/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "/Users/oskarkrawczyk/Sites/InstaPy/instapy/util.py", line 1415, in smart_run
    session.login()
  File "/Users/oskarkrawczyk/Sites/InstaPy/instapy/instapy.py", line 349, in login
    self.bypass_with_mobile):
  File "/Users/oskarkrawczyk/Sites/InstaPy/instapy/login_util.py", line 207, in login_user
    .move_to_element(login_elem)
  File "/usr/local/lib/python3.7/site-packages/selenium/webdriver/common/action_chains.py", line 80, in perform
    self.w3c_actions.perform()
  File "/usr/local/lib/python3.7/site-packages/selenium/webdriver/common/actions/action_builder.py", line 76, in perform
    self.driver.execute(Command.W3C_ACTIONS, enc)
  File "/usr/local/lib/python3.7/site-packages/selenium/webdriver/remote/webdriver.py", line 321, in execute
    self.error_handler.check_response(response)
  File "/usr/local/lib/python3.7/site-packages/selenium/webdriver/remote/errorhandler.py", line 242, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.MoveTargetOutOfBoundsException: Message: (901.4916915893555, 695) is out of bounds of viewport width (1280) and height (694)
```

Doesn't look like many people use it so perhaps the easiest way would be just to remove it from the docs.

**EDIT**

OS: macOS Mojave
Firefox: 64.0
Geckodriver: 0.23.0

gitpatrickhub:
It depends on many variables; Platform, Operating System, version of OS,  version of FireFox, version of Geckodriver. 

I'm using it on;
Platform = Raspberry Pi 3
OS = Debian Jessie
Firefox = 60.3.0 ESR
Geckodriver = 0.22.0

w4ngel:
> ov

Hi,

I use geckodriver exclusively and I don't get this- whereabouts does this happen?

[...]

PR #3662, opened 2019-01-02
#3656 (bug fix) change the way the login link is clicked

converge:
Sometimes the screen is smaller than the Instagram page, and the Login button wont be visible. ActionChains works looking for elements visible on viewport:

```python
(ActionChains(browser)
         .move_to_element(login_elem)
         .click()
         .perform())
```

throws this exception if the Login link isn't visible on viewport:
```selenium.common.exceptions.MoveTargetOutOfBoundsException: Message: (901.4916915893555, 640) is out of bounds of viewport width (1280) and height (638)```

(Login button isn't visible)
captura de tela 2019-01-02 as 08 59 34

--
The fix is click the link using the click() function.

### DIFF
--- a/instapy/login_util.py
+++ b/instapy/login_util.py
@@ -16,6 +16,7 @@ from .util import check_authorization
 # import exceptions
 from selenium.common.exceptions import NoSuchElementException
 from selenium.common.exceptions import WebDriverException
+from selenium.common.exceptions import MoveTargetOutOfBoundsException
 
 
 def bypass_suspicious_login(browser, bypass_with_mobile):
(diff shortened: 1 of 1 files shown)


timgrossmann:
@converge Thank you very much for the fix, could we maybe quickly discuss if there is another way to handle this while keeping the action chain movements?
(Thinking about InstaPy being more "human-like")

converge:
sure.. we could scroll down, but isn't a fancy solution.. if we're using it in a big screen, there is no need to scroll down..

I will sleep on it.

timgrossmann:
@converge Perfect, thank you very much.
I'm aware that it's only a problem on small screens and maybe it's unnecessary to think about the difference but maybe it comes in handy at one point, maybe.

Thanks for considering.
If we agree that there is no need, we'll merge this.

w4ngel:
@timgrossmann @converge this isn't happening across the board, would we be able to put it in a try catch instead?

`try:
     (ActionChains(browser)
         .move_to_element(login_elem)
         .click()
         .perform())
except MoveTargetOutOfBoundsException:
     login_elem.click()`

Then we can keep the more "human" way initially.

converge:
@w4ngel sounds like the best idea for me.

w4ngel:
Committed. @timgrossmann this is ready to merge.

timgrossmann:
@converge @w4ngel Awesome teamwork guys!
Thank you very much

Issue #3676, opened 2019-01-03
Problem with proxy and headless browser


patacalas:
I'm using proxy server with credidentials. MITMProxy creates a local http proxy that transferred network traffic through the auth proxy. [Get more help here](https://stackoverflow.com/questions/46002582/chrome-in-headless-mode-provide-credentials-auth-for-proxy)

## Expected Behavior
Http proxy without authentication should work fine in headless browser mode

## Current Behavior
Getting this error
> selenium.common.exceptions.InvalidArgumentException: Message: invalid argument: cannot parse capability: proxy
from invalid argument: Specifying 'socksProxy' requires an integer for 'socksVersion'

Chromedriver version 2.45.615279

## Possible Solution (optional)
Solutions that worked for me.

First solution: add proxy config directly in browser.py 
```
        if headless_browser:
            chrome_options.add_argument('--headless')
            chrome_options.add_argument('--no-sandbox')
            chrome_options.add_argument("--proxy-server=http://127.0.0.1:8090")
```

Second solution: delete `prox.socks_proxy = proxy` in browser.py and use standard proxy configuration in quickstart.py

w4ngel:
This is only for chrome correct?

w4ngel:
And did you make any other changes other than the solutions provided?

patacalas:
> This is only for chrome correct?

Yes

> And did you make any other changes other than the solutions provided?

I only made that modification in browser.py. Actually using second solution.

w4ngel:
Can you check if that pull request fixes the issue, please?

patacalas:
Yes, it's fully working. Thanks!

w4ngel:
@timgrossmann please close this issue.

octavius-1:
@patacalas Where are you getting the proxies? I’m getting my accounts banned after running them directly in AWS.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

patacalas:
@octavius-1  like block?

I got like block issue running Instapy in my residential Internet connection. 
Using proxies don't solve the problem.
I think the problem is Instagram detecting the headless browser. 
[Here is some information about making Chrome headless undetectable](https://intoli.com/blog/making-chrome-headless-undetectable/).

w4ngel:
@patacalas can you close this issue please?

[...]

PR #3679, opened 2019-01-04
Proxy Fix For Headless Chrome

w4ngel:
Fix for non-authenticated proxies in chrome headless browser as mentioned in issue #3676.

Credits to @patacalas for pointing out the solution.


timgrossmann:
@w4ngel @patacalas Thanks guys for this fix!

PR #3705, opened 2019-01-09
Runtime and skeleton implementation

lucanello:
**For easier use of the script and better multi-account implementation I wrote a runtime which should have every function in it and a skeleton which contains the configuration.**

When running the runtime with `python3 runtime.py <username> <mode (e.g. "only_unfollow" to skip all commands and just execute)>`, the configuration file is loaded and the bot is executed. This configuration file should be like skeleton.py and must be named `run_<username>.py`.

It is important that this runtime contains all commands that InstaPy has to offer in the future. Since I was unfortunately quite inactive lately, I could not implement some things.
This includes e.g. @uluQulu's Quota Supervisor. But I also couldn't finish the unfollow commands properly.

But I added more features like an integration to Pushover. I will extend this in the future. With Pushover it is possible to get a notification on your mobile phone about the last run bot.

Please ignore the `write_result` command at the bottom. This feature will be added later. It will add the possibility to write the results to a database in order to monitor your bots.

In the next step I would like to extend this runtime and the skeleton in such a way that we gradually approach an API. 

Thanks for your help!
Best regards,
Luca

### DIFF
--- a/runtime.py
+++ b/runtime.py
@@ -1,0 +1,277 @@
+from InstaPy.instapy import Settings
+from InstaPy.instapy import InstaPy
+
+from plugins.statistics import write_result
+from plugins.mail_support import mail_result
+
+from datetime import datetime
+
+import logging
+import json
+
+import argparse
+
+import requests
+
+global followed_this_time
+global liked_this_time
+global already_liked_this_time
+global commented_this_time
+global like_inap_this_time
+global follow_inap_this_time
+global followed_by
+global unfollowed
+followed_this_time = 0
+liked_this_time = 0
+already_liked_this_time = 0
+commented_this_time = 0
+like_inap_this_time = 0
+follow_inap_this_time = 0
+followed_by = 0
+unfollowed = 0
+
+# PARSE USERNAME FROM ARGUMENT #
+parser = argparse.ArgumentParser()
+parser.add_argument('username', action="store")
+parser.add_argument('type', nargs='?')
+parse_results = parser.parse_args()
+
+insta_username = parse_results.username
+type = parse_results.type
+
+if type != None:
+    print('---> Type argument given: ' + str(type))
+else:
+    print('---> No type argument given')
+
+# LOAD USER-SPECIFIC SETTINGS #
+import importlib
+libname = 'run_' + insta_username
+module = importlib.__import__(libname)
+
+print('MODULE LOADED: ' + str(module))
+
+# SET UP TIME AND LOGGING #
+now = datetime.now().strftime('%Y-%m-%d_%H:%M:%S')
+timestart = datetime.now().replace(microsecond=0)
+logging.basicConfig(filename='statistics/logs/' + now + '-' + insta_username +  '_instabot.log',level=logging.INFO)
+
+# SET UP INSTAPY #
+Settings.database_location = './db/instapy.db'
+Settings.chromedriver_location = './browser/chromedriver'
+session = InstaPy(username=module.insta_username,
+                  password=module.insta_password,
+                  headless_browser=True,
+                  multi_logs=True
+                  )
+
+try:
+    session.switch_language=False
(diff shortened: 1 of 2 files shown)


uluQulu:
Awesome, @lucanello 
I will read it soon.

---

Cheers 😁
---

w4ngel (changes requested):
run_skeleton.py - why is this a python file? It could just be a JSON file then you load it in the runtime.py

runtime.py - things should be run `with smart_run` this is due to that handling a load of errors and making sure the program closes properly, you can put the try/finally code. Though it isn't necessary I would highly recommend it, otherwise you can encounter issues like hanging processes etc.

lucanello:
Hi @w4ngel, thanks for your feedback. It was easier for me to load the configuration from a python file. I think I had some problems with JSON decoding. I'm sure you can do better and help me with that :)

Also, what do you mean with smart_run?

w4ngel:
> Hi @w4ngel, thanks for your feedback. It was easier for me to load the configuration from a python file. I think I had some problems with JSON decoding. I'm sure you can do better and help me with that :)
> 
> Also, what do you mean with smart_run?

Smart_run is a feature that most of the quickstart templates use. If you can add me as a collaborator I will be able to edit your fork for you.

w4ngel:
@lucanello if you let me know when you have added me so I can make the change.

w4ngel:
Ok, so the reason your JSON wasn't working is that it wasn't formatted like JSON properly. I am correcting this now as well as implementing smart_run on the skeleton. Actually, on second thought, I am not going to format it in JSON, I will just tidy it up as it is all meant to be in python, I will also fix any errors.

runtime.py
```diff
@@ -0,0 +1,277 @@
+from InstaPy.instapy import Settings
+from InstaPy.instapy import InstaPy
+
+from plugins.statistics import write_result
```
w4ngel, line 4: Are these two plugin.stats and plugin.mail_support modules custom?

runtime.py
```diff
@@ -0,0 +1,277 @@
...
+from plugins.mail_support import mail_result
```
w4ngel, line 5: And here as well...

runtime.py
```diff
@@ -0,0 +1,277 @@
...
+    #
+    # requests.post("https://api.pushover.net/1/messages.json",
+    #               data={'token': '<token>',
+    #                     'device': module.devices,
+    #                     'user': '<user>',
+    #                     'title': 'Instagram bot @' + insta_username + ' finished',
+    #                     'message': str(session.liked_img) + ' were liked and ' + str(session.followed) + ' followed. I posted ' + str(session.commented) + ' comments. I unfollowed ' + str(session.unfollowed) + ' users. You have ' + str(followed_by) + ' followers!'},
+    #               verify=False)
+
+    session.end()
```
w4ngel, line 354: Did you do all of this on a separate branch or your master branch?

runtime.py
lucanello: Yes, I'm planning to create a PR as soon as possible. Unfortunately it seems like everyone was writing on an analytics dashboard. And I don't think we need 4.

runtime.py
lucanello: Look above :)

runtime.py
lucanello: I did these changes on my master branch.

w4ngel:
@lucanello I am going through the runtime skeleton right now, and I am not exactly too sure what you are trying to achieve... There is a lot of bespoke configuration for your PC here. Along with unnecessary logging, parsing etc. When you have used InstaPy in the past have you managed to use any quickstart templates or tried to?

lucanello:
Hi @w4ngel, the skeleton file is only a template to setup the runtime. The runtime and skeleton are meant to replace the quickstart scripts.

PR #3724, opened 2019-01-12
Update preferred installation to use builtin venv and not virtualenv

ReagentX:
for versions >= 3.6


timgrossmann:
@ReagentX Thank you for the docs update! :+1:

PR #3752, opened 2019-01-16
Optimizing Dockerfile for smaller docker image

Herrox:
Hello all,

I am running `InstaPy` with `Docker` and a best practice when creating Docker containers is keeping the size of the final image to the minimum possible. For a number of reasons including:

- Disk on space
- Reducing attack surface
- Upload time (if so)
- Performance (for old Docker version)

So, I rewrite `./docker_conf/python/Dockerfile` to reduce by half the size the final image from `1.34GB` to `512MB`. I'm pretty sure there is still few optimization possible and the size might be lowered, but it's a first step.
More generally regarding Docker, I'm pretty sure we can do some cleaning in the configuration:

- I also think it could be possible to reduce the size of `selenium` image (which is around `1GB` too) but I haven't already look at it.
- Cleaning `docker-*.yml` files: it's better to stay with only one file
- Cleaning `docker_conf` directory

I also include in the `.gitignore` file `docker_quickstart.py` to avoid overwriting of the file that happened sometimes.

WAiting for your feedback.

Regards,
Herrox

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ __pycache__/
 *$py.class
 
 quickstart.py
+docker_quickstart.py
 
 # C extensions
 *.so
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ The **goal** of this file is explaining to the users of our project the notable 
 
 _The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)_.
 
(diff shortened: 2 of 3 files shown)


timgrossmann:
@Herrox Thank you very much for the update of the docker file size!
Congrats to your first contribution to InstaPy! 🙌

timgrossmann:
@Herrox Are you part of the discord channel? 
@scriptworld-git would like to ask you some questions in regards to the docker files in the repository since we want to completely extract the docker part out to this repository: https://github.com/InstaPy/instapy-docker

Thank you

PR #3759, opened 2019-01-16
Delay unfollow followbackers

sionking:
We better delay the unfollow to the follow backers.


sionking:
Review someone ?

timgrossmann:
@sionking Sorry for the delay, I'll try to do that in the coming week.
Would be perfect if you could test it once just to make sure it's working on your end.

What's the reason you are using `"undefined"` as a string here?

sionking:
I did test it till now :)
undefined is what used for userid so I keep on syntec

timgrossmann:
@sionking Perfect, thank you!
I'll give it a try tomorrow

timgrossmann:
@w4ngel Could you PEP8 this, please?

> Also, would it be possible to document your PEP8ify steps somewhere (best in research repo) and provide the formatter somewhere? (or is it private?)

w4ngel:
> @w4ngel Could you PEP8 this, please?
> 
> > Also, would it be possible to document your PEP8ify steps somewhere (best in research repo) and provide the formatter somewhere? (or is it private?)

I cannot for a couple of days while PC is still out of action. 

I am afraid I cannot share the PEP8 steps but I can provide some similar suggestions, the issue is most people use different IDEs/text editors and mine is built for PyCharm + it's private plugin.

Get my PC on Wednesday.

timgrossmann:
@w4ngel Sounds good, would be great if you could share it! 👍 
Thank you!

sionking:
Anyone tested ? I did .

instapy/unfollow_util.py
```diff
@@ -382,6 +376,31 @@ def unfollow(browser,
                                  person in automatedFollowedPool[
                                      "all"].keys() else False)
 
+                    if delay_follow_back:
```
uluQulu: Using it with the `nonFollowers` method is an overuse.

sionking:
> @sionking Sorry for the delay, I'll try to do that in the coming week.
> Would be perfect if you could test it once just to make sure it's working on your end.
> 
> What's the reason you are using `"undefined"` as a string here?

@timgrossmann you can't set None into test...

instapy/unfollow_util.py
```diff
@@ -44,7 +46,7 @@ def set_automated_followed_pool(username, unfollow_after, logger, logfolder):
     """ Generare a user list based on the InstaPy followed usernames """
     pool_name = "{0}{1}_followedPool.csv".format(logfolder, username)
     automatedFollowedPool = {"all": {}, "eligible": {}}
-    time_stamp = None
+    time_stamp = "undefined"
```
uluQulu: As @timgrossmann,
I will also not agree with the usage of the `"undefined"` (_JS style value_);

image

I have used the value of `"undefined"` out of a requirement for the `user_id` variable.
image

image
image
image
image


As you can easily see,
`"unknown"`, `"undefined"` and `None` serve for different purposes and can't be merged or replaced.

But I don't see that requirement for the `time_stamp` variable.

That noted, we must always try to go with the Python recommended values and types, so that, it must use `None`.
Otherwise, it will cause a confusion.

instapy/instapy.py
```diff
@@ -223,7 +223,7 @@ def __init__(self,
         self.skip_business_percentage = 100
         self.skip_no_profile_pic_percentage = 100
         self.skip_private_percentage = 100
-
+        self.delay_unfollow_followbackers = 864000  # 864000 = 10 days, 0 = don't delay
```
uluQulu: Should be added as a parameter to the `unfollow_users()` feature to able to be configured by the user easily.
And by default it should be turned off.

instapy/util.py
```diff
@@ -1849,3 +1849,23 @@ def save_account_progress(browser, username, logger):
             conn.commit()
     except Exception:
         logger.exception('message')
+        
+
+def get_epoch_time_diff(time_stamp):
+    ''' time diff in seconds from input to now'''
+    log_time = datetime.strptime(time_stamp, '%Y-%m-%d %H:%M')
```
uluQulu: In original, it was handling `ValueError` exceptions.
I see that it is lost here.
image

Value error can natively occur for certain reasons.
One of them is a user changes the datetime value in the entry line to an unexpected or invalid [datetime] format.

It must be handled.

instapy/unfollow_util.py
```diff
@@ -382,6 +376,31 @@ def unfollow(browser,
...
+
+                        if is_follow_me(browser, person):
+                            # delay follow-backers to 15 days.
+                            time_stamp = (automatedFollowedPool["all"][person]["time_stamp"] if
+                                         person in automatedFollowedPool["all"].keys() else False)
+                            if time_stamp not in [False, "undefined"]:
+                                try:
+                                    time_diff = get_epoch_time_diff(time_stamp)
+
+                                    if time_diff < delay_follow_back:  # N days in seconds
```
uluQulu: A bit weird logic 🤔

considering a `delay_follow_back=864000`  (_10 days_)  [`self.delay_unfollow_followbackers`]

and you've followed a user for 9 days,

then it will check if it has been followed at least 10 days.

As it hasn't done it, then it will change datetime value in the entry line and delay unfollow 10 more days.

It results, that user will be able to be unfollowed 19 days later, if does not reset again.

uluQulu:
The idea is neat, @sionking 
I have analyzed the code and outlined my concerns as single comment reviews.

instapy/unfollow_util.py
sionking: How can you set None for text file followpool ?

instapy/instapy.py
sionking: ok

instapy/unfollow_util.py
sionking: I know, but it can save extra argument.
user should just expect:
real_Delay = delay_follow_back + unfollow_after

sionking: I don't follow a user more than 2-3 days. it will explode the following.

instapy/util.py
sionking: I don't see why handling such a case should be done.
If you got your DB corrupted, fix the DB, don't just ignore it !
BTW, if you Ignored is by setting continue, you will **never unfollow** the user !!!! you will just ignore him forever.

instapy/unfollow_util.py
sionking: So do you suggest also checking for "nonFollowers" inside this IF?

sionking: Sometimes it is better to not complicate things ...

uluQulu: I don't think so.
All times, things must do what they are intended to do. 
Nothing more, nothing less.

instapy/unfollow_util.py
uluQulu: >..text file followpool ?

It's just a variable waiting for a value to be assigned to.

instapy/util.py
uluQulu: If it was not handled previously, I would probably not raise that issue.
But it was already handled clearly by good people before.
And you have digged that solution into history.
It is what must be mentioned.

Some programmers think like so, but not me since I think,
1. There are too many newbies using this software
2. How stable is software, that strong it is.

If a datetime entry is invalid, your DB is not corrupt. Actually it's not a DB, rather a CSV file.
So, it's just your one line of entry having invalid datetime value.

If you are about totally flushing the CSV file, then you should add a new function to do it.
But not have your code expose its weakness just to observe that bug-scenario.

Also, you have noticed or not, in your newly added code, you have already catched that `ValueError` exception for the same case.
image
image
_My comment in-between separated that block so attached 2 screenshots for it_ xD

My suggestion is handle it inside your new  `get_epoch_time_diff()` function which will handle it everywhere it has been called.

instapy/unfollow_util.py
uluQulu: I think you should clearly explain what this new addition can do.
Maybe we (_I_) don't understand it as you do.

Also, the solution has to be targeting general audience and be generic..
At this stage, it seems to cause a misbehavior off my [personal] style.

Waiting for your clear explanation on the usage/scope of this feature before trying to share my further review.

---

Cheers 😁
---

instapy/util.py
sionking: but don't you agree that one line of user that will be catches over and over again and will never be handled correctly ? also will never be removed from CSV! that a big hole.
stability is not everything, I prefer data to be correct.

instapy/unfollow_util.py
sionking: but setting none to csv file and retrieving that none, will not give you type None, right ?

instapy/util.py
sionking: @timgrossmann @uluQulu this is a bug don't you agree?

uluQulu: @sionking, I will re-caption my previous statement;
image
When your code is _unstable_, you lose a user.
A user can always _correct_ an invalid entry if you (_by code_) tell him/her to. 

It's not a bug, instead a realtime scenario which **can** occur.

instapy/unfollow_util.py
uluQulu: Ok, @sionking 
You mean,
"_Why `user_id` gets a value of `None` and also `"undefined"`, .._
_and I (_`time_stamp`_) also want to get the value of `"undefined"`_ >.<"

I will re-caption, again.
image


If `"undefined"` looks really annoying cos of it is also a JS type, I can rename it to something else for the `user_id` variable.
I just liked its definition could be used for our case needed to differentiate absence levels [of user ID] out of various reasons which comes to help e.g. while debugging..

BTW, print this,
```python
eval("None")
```

--- 

Cheers 😁

instapy/util.py
timgrossmann: Unfortunately, in my opinion, both of you are right... Data integrity is extremely important but we also can't lose users because the tooling is less stable.

Files can't give us the same guarantees DBs can. Maybe we should take the remaining things that are not completely written into the DB (if there are any at all) and get rid of the files themselves.
sqlite also gives us very nice tooling to view the data if that is an aspect.
https://sqlitebrowser.org

What is the reason we haven't completely removed the files for the DB up to this point?

instapy/unfollow_util.py
sionking: done

instapy/util.py
sionking: Since most of the users know how to handle csv files and read it while others don't know to handle db.

timgrossmann: True, but using a tool like sqlitebrowser is super simple and has very nice visuals, too.
It seems like keeping text based files just introduces complex code because we have to handle data integrity and stability ourselves which we don't with the DB.

sionking:
@timgrossmann @uluQulu 
I committed the changes according to the comments please confirm.

uluQulu:
@sionking

My points on `ValueError` remains the same.
Sadly, you have mishandled it unlike it is in the existing code;

Existing code [block],
```python
try:
    # nothing special
except ValueError:
    # handle
```

Your code,

```python
# block moved into a new function
def new():
    # nothing handled

# 1st function call [UNhandled]
new()

# 2nd function call [handled]
try:
    new()
except ValueError:
    # handle
```
What you should do
```python
def new():
    try:
        # all same
    except ValueError:
        # handle
```
And do not handle it inside function call(s).
If you don't want to get loads of invalid entries inside a data file.
Then make another function to check data integrity and please, don't unstabilize the existing code.
Also, you can simply print the line with its number saying it is invalid and tell the user he/she can fix if wants (_that is the same and even better than crashing to console traceback_).

I also agree with @timgrossman's point on wannabe usage of DB in such structures.
We can't provide data integrity as e.g. SQLite databases and even if we try, it is so unwise.
And yeah, we want data integrity ..


---


Also, you have left my concern unanswered on my last code review of this PR,
image

Please, remember that, not everyone has your style of use, @sionking.

---


Your new commits are nice, e.g. using PYTHONIC `None`,..
But you should not add the optional parameter- `delay_unfollow_followbackers` into _InstaPy_'s constructor.
Maybe add it to `unfollow_users()` arguments?



---

Cheers 😁
---

sionking:
> Your new commits are nice, e.g. using PYTHONIC `None`,..
> But you should not add the optional parameter- `delay_unfollow_followbackers` into _InstaPy_'s constructor.
> Maybe add it to `unfollow_users()` arguments?

I don't get you here....

uluQulu:
@sionking 
Which part?
image

sionking:
@uluQulu 
This delay_unfollow_followbackers is relevant to any unfollow request, not only to  `unfollow_users` method.

uluQulu:
Alright, @sionking 

Standard unfollow procedure is being operated by the `unfollow_users()` feature;
Other places that also run the unfollow engine- `unfollow_user()` function are:

1-) Simulated follow from a dialog box and then user has been found to be non-valid.
It will unfollow directly.

2-) Removing follow request - `remove_follow_requests()` also runs the `unfollow_user()` function.

3-) `like_by_feed()` feature unfollows the user if its `unfollow` parameter is set as `True` and user has an inappropriate content.

All of those 3 processes must be accomplished immidiately and not be delayed at all.
So, there is no reason to store the new `delay_unfollow_followbackers` parameter inside _InstaPy_'s constructor.
Even there were some little reason, then you should add a new function to set it rather than adding it to the constructor cos it does not really fit in there.

---

Cheers 😁
---

sionking:
> @sionking
> 
> My points on `ValueError` remains the same.
> Sadly, you have mishandled it unlike it is in the existing code;
> 
> Existing code [block],
> 
> ```python
> try:
>     # nothing special
> except ValueError:
>     # handle
> ```
> 
> Your code,
> 
> ```python
> # block moved into a new function
> def new():
>     # nothing handled
> 
> # 1st function call [UNhandled]
> new()
> 
> # 2nd function call [handled]
> try:
>     new()
> except ValueError:
>     # handle
> ```
> 
> What you should do
> 
> ```python
> def new():
>     try:
>         # all same
>     except ValueError:
>         # handle
> ```
> 
> And do not handle it inside function call(s).
> If you don't want to get loads of invalid entries inside a data file.
> Then make another function to check data integrity and please, don't unstabilize the existing code.
> Also, you can simply print the line with its number saying it is invalid and tell the user he/she can fix if wants (_that is the same and even better than crashing to console traceback_).
> 
> I also agree with @timgrossman's point on wannabe usage of DB in such structures.
> We can't provide data integrity as e.g. SQLite databases and even if we try, it is so unwise.
> And yeah, we want data integrity ..
> 
> Also, you have left my concern unanswered on my last code review of this PR,
> image
> 
> Please, remember that, not everyone has your style of use, @sionking.
> 
> Your new commits are nice, e.g. using PYTHONIC `None`,..
> But you should not add the optional parameter- `delay_unfollow_followbackers` into _InstaPy_'s constructor.
> Maybe add it to `unfollow_users()` arguments?
> 
> ## Cheers 😁

I still don't agree here.
If someone is using Instapy he must have some knowledge in programming and problem solving and not just running scripts. This entire application is far from stable, so better fix issue manually then avoid them.

uluQulu:
@sionking 
>If someone is using Instapy he must have some knowledge in programming and problem solving and not just running scripts. This entire application is far from stable, so better fix issue manually then avoid them.

Oppositely, most of the users do not have any knowledge in programming and copy/paste code from other threads, templates, etc. and change numbers to get it work.
Even after easy installation by **pip** gets into action, then newbies' ratio will vastly increase.

It can be true that entire app is ~unstable cos of it is fully open-source and relies on live external server(s) which makes it hardly to ever become fully stable.
But, it does not mean you can make it unstable as you like.
Oppositely, you have to solve it to make it more stable with your solutions.

---

Cheers 😁
---

sionking:
@uluQulu I still don't see how avoiding corrupted data make the code stable.
A user have defected follow pool resulting in **not unfollowing** 100's of instapy followed, resulting in following number gets 7500. is that stable for you ? ok...

sionking:
@uluQulu please check if this solution suite you

uluQulu:
Yes, @sionking 
That part looks good to me 🙂

sionking:
@timgrossmann So basically some documentation is missing ? did you test it ?

timgrossmann:
@sionking So sorry for the delay... So much stuff is going on at the moment.
I'll look over it now.

instapy/instapy.py
```diff
@@ -98,7 +98,8 @@ def __init__(self,
                  disable_image_load=False,
                  bypass_suspicious_attempt=False,
                  bypass_with_mobile=False,
-                 multi_logs=True):
+                 multi_logs=True,
+                 delay_unfollow_followbackers=0):
```
timgrossmann: Could you please add remove this from the InstaPy constructor and only keep it as a parameter for the `unfollow_users` method? 😊

instapy/instapy.py
```diff
@@ -223,7 +224,7 @@ def __init__(self,
         self.skip_business_percentage = 100
         self.skip_no_profile_pic_percentage = 100
         self.skip_private_percentage = 100
-
+        self.delay_unfollow_followbackers = delay_unfollow_followbackers  # 864000 = 10 days, 0 = don't delay
```
timgrossmann: Please use the default value to 0 in the argument list and add documentation in the README that explains how to use that feature with an example with the `864000` to delay it 10 days.

instapy/unfollow_util.py
```diff
@@ -179,6 +174,7 @@ def unfollow(browser,
              white_list,
              sleep_delay,
              jumps,
+             delay_follow_back,
```
timgrossmann: is this meant to be the same variable as `delay_unfollow_followbackers`?
If yes, could we please keep the naming consistent?

instapy/util.py
```diff
@@ -1972,3 +1972,26 @@ def close_dialog_box(browser):
     except NoSuchElementException as exc:
         pass
 
+
+def get_epoch_time_diff(time_stamp, logger):
```
timgrossmann: Nice! 😮

instapy/util.py
```diff
@@ -1972,3 +1972,26 @@ def close_dialog_box(browser):
...
+        return None
+
+
+def is_follow_me(browser, person=None):
+    # navigate to profile page if not already in it
+    if person:
+        user_link = "https://www.instagram.com/{}/".format(person)
+        web_address_navigator(browser, user_link)
+
+    return getUserData("graphql.user.follows_viewer", browser)
```
timgrossmann: Maybe we have to backup this with the xPath version of reading the text from the follow button
=> extract the checking to a new method that checks if we are already following that person.
What do you think?

timgrossmann (changes requested):

timgrossmann:
@sionking I've added a few comments, would be awesome if you could check them and discuss them with me.

I've also changed the merge branch to `dev` since we'll start to only merge changes/additions into `dev` and then merge `dev` into `master` once we have some changes aggregated.
Only bug fixes will be PRed to `master` from now on 😊 

Thank you very much for your contributions

instapy/util.py
sionking: I don't know xpath very well, it can be added later.
I think we shall keep the graphql method as default, also add some testing for changes.

instapy/unfollow_util.py
sionking: done

instapy/instapy.py
sionking: done

sionking:
added docs

timgrossmann (approved):

timgrossmann:
@sionking Awesome, thank you very much for the changes and documentation! 👍 
I've adjusted the changelog and docs a little bit and added an example on the usage, feel free to double check if what I wrote is correct.

README.md
```diff
@@ -880,7 +880,16 @@ session.unfollow_users(amount=200, allFollowing=True, style="FIFO", unfollow_aft
...
+`delay_unfollow_followbackers`  
+Delay the unfollow for users that follows you, the delay is in seconds and will delay from the moment that we "checked" to unfollow the user.
+It will add the delay_unfollow_followbackers seconds on top of that.   
+0 seconds == disabled   
+864000 (or `240*60*60`) = 10 days delay  
 
+```python
+session.unfollow_users(amount=60, InstapyFollowed=(True, "all"), style="FIFO", unfollow_after=48*60*60, delay_unfollow_followbackers=240*60*60)
+```
+This will unfollow all users that are not following you back after **2 days**. The unfollowing for users that are following you back will be **delayed by an additional 10 days** (12 days to unfollow in total).
```
sionking, line 926: great

instapy/unfollow_util.py
```diff
@@ -384,6 +380,36 @@ def unfollow(browser,
...
+                                try:
+                                    time_diff = get_epoch_time_diff(time_stamp, logger)
+                                    if time_diff is None:
+                                        continue
+
+                                    if time_diff < delay_unfollow_followbackers:  # N days in seconds
+                                        refresh_follow_time_in_pool(username,
+                                                                    person,
+                                                                    person_id,
+                                                                    delay_unfollow_followbackers,
```
timgrossmann: This should be `time_stamp` instead of the the time difference. Otherwise we will extend the delay every time we encounter the user, don't we? 🤔

instapy/unfollow_util.py
```diff
@@ -1550,3 +1576,16 @@ def get_follow_requests(browser, amount, sleep_delay, logger, logfolder):
         .format(len(users_to_unfollow), users_to_unfollow))
 
     return users_to_unfollow
+
+
+def refresh_follow_time_in_pool(username, person, person_id, extra_secs, logger, logfolder):
+    # set the new time to now plus extra delay
+    logtime = (datetime.now() + timedelta(seconds=extra_secs)).strftime('%Y-%m-%d %H:%M')
```
timgrossmann, line 1638: This shouldn't be necessary, since the `extra_secs` should be the timestamp instead.
When using it like this we create a new time stamp every time that we encounter the user.

What is the intention with this logic?
This way we will only unfollow the user if we encounter it once after the delays have been surpassed.

timgrossmann:
@sionking Sorry, it's kind of weird but the code seems to be off when updating the entry of the file.
Also, I think we have to rename the variable from `delay_unfollow_followbackers` to `delay_followbackers` since it already is a parameter of the unfollow method. 🙈

instapy/unfollow_util.py
sionking: I just realized this an hour ago also !
I can't find a proper solution, since not changing the timestamp will result in checking the user page each unfollow loop.
I think we shall have another database/flag, "followback" next to userid and timestamp? that way we won't need to change the timestamp and won't enter the user page if time diff is not sufficient.

timgrossmann:
@sionking I've updated the branch to latest `dev` and renamed the parameter.
Would be great if you could double check the logic for the delays.

timgrossmann:
@sionking Could we please finish this PR, first?
The only thing missing before merging this is that the timestamp won't get updated every time it is checked but only once when saved to the DB so we only have `time_stamp + delay_followbackers` once and it won't be refreshed every time it is checked. 😊

sionking (commented):
> @sionking Could we please finish this PR, first?
> The only thing missing before merging this is that the timestamp won't get updated every time it is checked but only once when saved to the DB so we only have `time_stamp + delay_followbackers` once and it won't be refreshed every time it is checked. 😊

So that is what I was working on, but it is not **simple** at all.
I will add another field next to userid in the followpool, that will set to true when followback check is true (this should be better used with rational database approach, but good for now.
later I will filter followbackers by delay follow back at the automatedFollowedPool.
when the time delay finished I will check that the flag is not yet set, only then I will set it otherwise we will unfollow the user.

Kind of complicated, but we must save the state for the person somehow.
Also, it will be good for the future to get the followback statistics.

timgrossmann:
@sionking Sorry for the heavy work you have to put into this 🙁 
I thought we could simply pass the `time_stamp` from this part

```
                    if is_follow_me(browser, person):
                             # delay follow-backers in delay_follow_back.
                             time_stamp = (automatedFollowedPool["all"][person]["time_stamp"] if
                                          person in automatedFollowedPool["all"].keys() else False)
```

into the `refresh_follow_time_in_pool` and use that to reinsert the user into the file. 
But I'm probably missing something here.

I'd actually prefer the solution with the database since implementing it for the files will lead to more, additional, work later on when removing the file based approach.

What do you think? 
Is it doable for you?
It sounds sophisticated, would love to see that implemented.

sionking:
@timgrossmann @uluQulu 
Please review the changes as it is not minor changes.
Also @uluQulu there was a bug fix regarding userid.

instapy/unfollow_util.py
```diff
@@ -36,24 +36,31 @@
...
     try:
         with open(pool_name, 'r+') as followedPoolFile:
             reader = csv.reader(followedPoolFile)
 
             for row in reader:
+
+                # init
+                followedback = None
+                user_id = "undefined"  # 'undefined' rather than None is *intentional
+                eligle = True  # 'undefined' rather than None is *intentional
```
sionking, line 62: remove comment

sionking (commented):
> @sionking Sorry for the heavy work you have to put into this 🙁
> I thought we could simply pass the `time_stamp` from this part
> 
> ```
>                     if is_follow_me(browser, person):
>                              # delay follow-backers in delay_follow_back.
>                              time_stamp = (automatedFollowedPool["all"][person]["time_stamp"] if
>                                           person in automatedFollowedPool["all"].keys() else False)
> ```
> into the `refresh_follow_time_in_pool` and use that to reinsert the user into the file.
> But I'm probably missing something here.
> 
> I'd actually prefer the solution with the database since implementing it for the files will lead to more, additional, work later on when removing the file based approach.
> 
> What do you think?
> Is it doable for you?
> It sounds sophisticated, would love to see that implemented.

If we only won't change the timestamp, we will get a working solution.
But, every unfollow loop we will access the followbackers profile to see that they still follow-backed, we don't want this, entering unnecessary pages 10-100's of times, so we need to set a flag for the user.

Regarding DB, this solution I implemented is very easy to transfer into db, it is just another field near userid and timestamp.

 I tested the solution and it seems to not break anything and working as planned.
Be my guest to test it.

instapy/unfollow_util.py
timgrossmann: @sionking ?

sionking:
@timgrossmann  TESTED ?

timgrossmann:
@sionking Sorry, not yet.
I won't make it until tomorrow for the new release, sorry.

Also, next week are my exams so I'll have to move this to earliest coming Friday.
Those are not small changes so I need to take some time with it.
I'll let you know, I'll add it to my todo list for the upcoming week after the exams.

sionking:
Can anyone ease test it ?
It  seems to work fine.

sionking:
@CharlesCCC @converge ?

timgrossmann:
@sionking Sorry 🙁 it's quite busy at the moment...

CharlesCCC:
@sionking I will find sometime to test it out tomorrow night. (if nobody else test before that)

README.md
```diff
@@ -546,11 +547,18 @@ session.follow_user_followers(['friend1', 'friend2', 'friend3'], amount=10, rand
...
+session.follow_by_tags(['tag1', 'tag2'], amount=10),
+```
+#####  Follow user based on hashtags (interact with user, liking images)
+* You can also **interact** with the users you just started by activating `interact = True`, which will use the `set_user_interact` configuration:
+```python
+session.set_user_interact(amount=2,
+                            percentage=70,
+                            randomize=True,
+                            media='Photo')
+session.follow_by_tags(['tag1', 'tag2'], amount=10, interact=True),
```
CharlesCCC: super minor issue, i would remove the `,` or replace it with `;` here.

CharlesCCC:
Started testing this tonight, 
one error showing on the log
```
ERROR [2019-02-27 19:38:07] [username]  Error occurred while generating a user list from the followed pool!
	b'get_epoch_time_diff() takes 1 positional argument but 2 were given'
```
I currently have set the following
```
session.unfollow_users(amount=60, InstapyFollowed=(True, "all"), style="FIFO", unfollow_after=24*60*60, delay_followbackers=48*60*60)
```

so it will take sometime to confirm if everything worked as expected. (need have someone follow me back)

sionking:
> ERROR [2019-02-27 19:38:07] [username]  Error occurred while generating a user list from the followed pool!
> 	b'get_epoch_time_diff() takes 1 positional argument but 2 were given'

thanks ! I updated the version to contain logger arg now. please pull.

CharlesCCC:
@sionking Pull with latest change and everything seems working fine this time. 
I'm good with Merge this PR. (of course after resolve the merge conflicts). 
Thank you for the contribution.

sionking:
> @sionking Pull with latest change and everything seems working fine this time.
> I'm good with Merge this PR. (of course after resolve the merge conflicts).
> Thank you for the contribution.

did you test a follow backer ? is that works well ?

CharlesCCC:
@sionking yes, i have the set that while I was testing
```python
session.unfollow_users(amount=60, InstapyFollowed=(True, "all"), style="FIFO", unfollow_after=24*60*60, delay_followbackers=48*60*60)
```
seems working fine with me.

timgrossmann:
@CharlesCCC Thank you very much for testing 🙏 
Let's merge this and see how it turns out over time 🙌 

Thank you very much @sionking for the feature 😊

meinert:
This new parameter needs documentation.......it is a bit unclear to me what a "follow backer" is and thus how this is different from the "unfollow_after"

CharlesCCC:
@meinert "follow backer" is someone you followed and followed your back. LOL .

unfollow_after is set the time when you choose to unfollow someone. a day, a week or a month.

PR #3783, opened 2019-01-20
Added Beta Extensions, Singleton (need fix)... See commits or comments

antwal:
I have to finish the implementation; but this works very well.

WARNING:
If you use the schedule library, an error is generated when using chrome in local session.


timgrossmann:
@antwal This looks really interesting, could you explain some more what this PR is useful for, please?

Maybe we should create a separate file for the singleton class.
Also, could we please make sure that we stick to single quotes (we have to unify that at one point).
Really nice addition, we'll discuss this PR.

What is your usage with this feature?

Thank you for your contribution!

antwal:
@timgrossmann Currently I removed the singleton because errors occur; I try first to solve and then we can implement it.
In this way we can have a unique access to the loaded instance useful if we end the implementation of the sqlite and mysql database, also it is useful in the server version if a restful API is implemented.

the other part for the extensions I'm still working on it because it does not work perfectly; however I have implemented currently in the local version, the extensions for:
- telegram-bot
- chrome-load-extensions
- watch-stories
- publish-photos-feed

without having to make changes to the original instapy code every time

antwal:
@timgrossmann in days I send you a PR to make all the code more dynamic; I close it

antwal:
@timgrossmann I had to send the changes to Config Manager, but the recent commit with Workspace compromises the structure. As soon as possible I will make the changes, currently you can find it only on my fork

timgrossmann:
@antwal Thank you very much for letting me know and your adaptability. 
You're right, the workspace is a big structural change but it will really add much value to the project.

Thanks for considering to adapt your PR.
Best
Tim

antwal:
@timgrossmann in a few hours I send the PR to you; I am doing some tests even if I have not yet converted and adapted all the code

PR #3764, opened 2019-01-17
Fix "_Unable to locate element: ...xpath","selector":"//div[text()=\'Likes\'..._" error

uluQulu:
I have been reported it occurred with `follow_likers()` feature.

Fixes are applied to the `likers_from_photo()` function and each change is explained in the commit description.

---

Cheers 😁
---

PR #3796, opened 2019-01-22
Fix `set_dont_unfollow_active_users()` feature completely

uluQulu:
I wanted to solve this issue after reading @sionking said that feature does not work anymore.

He wanted [proposed] to roughly apply the steps in the newly fixed `likers_from_photo()` function inside this feature- `get_active_users()` at #3764.

But `get_active_users()` function is not as simple as `likers_from_photo()` since `get_active_users()` is written to behave smartly.
E.g. I remember, I made it realize when it hits the bottom of the dialog box and stop scrolling at that point so that no any extra server call would be sent.

That was the main reason I solved it - to save `get_active_users()` from possible loss 😎

---

Also, now `set_dont_unfollow_active_users()` feature also has a Progress Tracker support.

Each commit made is unique with its specific explanation.
Please, read commit descriptions to view the changes occurred inside this PR.

---

Cheers 😁
---

### DIFF
--- a/instapy/settings.py
+++ b/instapy/settings.py
@@ -56,5 +56,27 @@ class Settings:
 
 class Storage:
     """ Globally accessible standalone storage """
+
     # store realtime record activity data
     record_activity = {}
+
+
+class Selectors:
+    """
+    Store XPath, CSS, and other element selectors to be used at many places
+    """
+
+    likes_dialog_body_xpath = (
+        "//h1[text()='Likes']/../../following-sibling::div/div")
+
+    likes_dialog_close_xpath = "//span[contains(@aria-label, 'Close')]"
+
+
+
+
+
+
+
+
+
+
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,18 @@ The **goal** of this file is explaining to the users of our project the notable 
 _The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)_.
 
 
+
+## [Unreleased] - 2019-01-22
+### Added
+- Now `set_dont_unfollow_active_users()` feature also has a Progress Tracker support.
+
+### Fixed
+- Fix `set_dont_unfollow_active_users()` feature completely.
+
+
 ## [Unreleased] - 2019-01-17
 ### Changed
-- Optimizing Dockerfile for smaller docker image
+- Optimizing Dockerfile for smaller docker image.
 
 ### Fixed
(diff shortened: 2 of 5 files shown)


sionking:
> E.g. I remember, I made it realize when it hits the bottom of the dialog box and stop scrolling at that point so that no any extra server call would be sent.

why not just add this to likers_from_photo ? and save extra function?

timgrossmann:
@uluQulu Really nice start of the `Selectors` in the `Settings` class 😉 
Extracting all of them to a central file will help ease up work done on changes! 👍

sionking:
BTW I implemented also a feature that store the grabbed data into db, so it won't rerun this like grabbing for the next X hours.

timgrossmann:
@sionking @uluQulu Can we find a solution to merge those two features into one working state? 😊

uluQulu:
I don't understand.
Previously it was not working at all and I fixed it in its best..

---

Can you make those changes [without loss of any logic and behaviour in code] and PR then, @sionking?

I also don't look warm to the idea of storing likers data that one-directional even though you say per X hours.

But if you do it, add a parameter to turn it off, too.
Like in the **relationship tools** - `live_match=False`.

---

>why not just add this to likers_from_photo ? and save extra function?

It's not just the example I mentioned, just look at the source code of those functions.
They are different.

`likers_from_photo()` is very simple since `get_active_users()` is very compact handling many scenarios and they have separate requirements.

What can be done?
- `likers_from_photo()` can be integrated to use some portion of `get_active_users()`

---

Cheers 😁
---

sionking:
@uluQulu please done with your pr I will add the logic later.

instapy/settings.py
```diff
@@ -56,5 +56,27 @@ class Settings:
...
 class Storage:
     """ Globally accessible standalone storage """
+
     # store realtime record activity data
     record_activity = {}
+
+
+class Selectors:
+    """
+    Store XPath, CSS, and other element selectors to be used at many places
```
sionking, line 66: good

sionking (approved):

JoeFertig:
Tested @uluQulu 's PR on a Raspberry Pi 3b+ and it's working!

instapy/settings.py
uluQulu: thanks, @sionking, it is part of your idea 😇 from recent discussions..
I have thought of adding a new file but then realized it will be an overuse.
So the existing **settings** file smelt more appropriate with a new storage class added for it.

Let's see, how it will look when the `Selectors` class grow huge..

BTW **settings.py** module might be renamed to sth like **globals.py**, hem?
@timgrossmann, @sionking?

sionking: leave it for now :()

PR #3812, opened 2019-01-23
Fix inconsistent use of spaces before parameters

obendev:
I commitet it directly in the master branch this time @timgrossmann 😁
Again only a small change like the thing with the commas. (It's still only fixed in dev 😝)


timgrossmann (approved):

timgrossmann:
@obendev Thank you for this fix, it's really important to have the documentation as close to just copy and pasting, to make it work, as possible.

Congrats to your PR 😉

PR #3813, opened 2019-01-23
Added follow_by_locations.

patacalas:
If you like or comment on many photos from users who you are not following, Instagram's algorithm recognizes it as SPAM and blocks actions from your account. 
I think it's important to add the "follow_by_locations" functionality without liking or commenting photos needed.


w4ngel (approved):
Looks good from me, needs PEP8 but other than that it's ok!

w4ngel (approved):

timgrossmann:
@w4ngel Thank you very much for reviewing it. 
I agree, it definitely looks good, but since we are approaching big change, we should try to reduce the amount of code by generalising everything (over time) to a `interact_by` e.g. `interact_by_locations` and then decide what should be done by using `set_do_like`, `set_do_follow`, `set_do_comment` etc.
We need to reduce the places of possible errors to a minimum.

@patacalas Do you think you can refactor your code to resemble that idea? 😊

potentowl:
I had this idea the other day lol

finley777:
> I had this idea the other day lol

You managed to embody it?

timgrossmann:
@patacalas After considering this and looking through the docs once more (we have like and comment by location) already in there, I think adding this feature, will be what people expect until we are able to reduce the amount of methods by using `interact_by_...` for all of them 🤔 

@w4ngel Do you think you can adjust add the PEP8 changes?

I think we definitely need some standardised PEP8 tooling that people can use for their code to commit.

@patacalas Do you think you could enrich the documentation with this feature? I think it's enough to have an example and one sentence or something like that just like with the others (like and comment)

patacalas:
> @patacalas Do you think you could enrich the documentation with this feature? I think it's enough to have an example and one sentence or something like that just like with the others (like and comment)

Done ...there is a new pull request  😉

timgrossmann:
@w4ngel Let's PEP8 the codebase again at one point using a custom PEP8 tool that can be made available to the users.

timgrossmann:
@patacalas Thank you very much for your contribution and congrats to your first merged PR to InstaPy 🙌 😊

PR #3825, opened 2019-01-24
fix no posts exception; add track

sionking:
Fix no posts exception when scraping likes
Add track post/profile
Avoid prints for only one user

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -464,7 +464,7 @@ class InstaPy:
             return self
 
         self.do_like = enabled
-        self.like_percentage = percentage
+        self.like_percentage = min(percentage,100)
 
         return self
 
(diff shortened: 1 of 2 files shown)


timgrossmann (approved):

timgrossmann:
@sionking Thank you very much for those updates!

Why don't we want to log the activities if we only have one account?

sionking:
too many logs when calling this function each time in a loop every time with one log only. and all the actively for the one user is already printed in the actions it those

timgrossmann:
@sionking If we can get one more approval/tester I'll merge this, thank you very much for this addition! 👍

timgrossmann:
@sionking Tested those small additions again, they work fine.
Let's just merge this.

Thank you for the additions!

PR #3829, opened 2019-01-25
added instapy-chromedriver package if no chromedriver is in path

breuerfelix:
In requirements.txt I used the specific chromedriver version 2.44 since 2.45 gets me an chromedriver error:  
Chromedriver version needs to be between 70-73.

I don't know if this error is only present for me.

Otherwise we need to change `instapy-chromedriver==2.44` to `instapy-chromedriver>=2.44` in file requirements.txt.

I also converted a random method description string into a pydoc string. Just noticed this while scrolling a file :)

### DIFF
--- a/instapy/file_manager.py
+++ b/instapy/file_manager.py
@@ -224,7 +224,7 @@ def get_chromedriver_location():
 
         CD = binary_path
         chrome_version = pkg_resources.get_distribution("instapy_chromedriver").version
-        message = f"Using built in instapy-chromedriver executable (version {chrome_version})"
+        message = "Using built in instapy-chromedriver executable (version {})".format(chrome_version)
         highlight_print(Settings.profile["name"],
                         message,
                         "workspace",
--- a/instapy/file_manager.py
+++ b/instapy/file_manager.py
@@ -221,7 +221,6 @@ def get_chromedriver_location():
         assets_path = "{}{}assets".format(workspace_path, native_slash)
         validate_path(assets_path)
 
-        Settings.chromedriver_location = binary_path
         CD = binary_path
         message = "Using built in instapy-chromedriver executable."
         highlight_print(Settings.profile["name"],
@@ -230,6 +229,8 @@ def get_chromedriver_location():
                         "info",
                         Settings.logger)
 
+    # save updated path into settings
+    Settings.chromedriver_location = CD
     return CD
 
 
--- a/instapy/file_manager.py
+++ b/instapy/file_manager.py
@@ -7,6 +7,8 @@ from os.path import isfile as file_exists
 from os.path import sep as native_slash
 from platform import python_version
 
+from instapy_chromedriver import binary_path
+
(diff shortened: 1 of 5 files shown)


instapy/file_manager.py
```diff
@@ -219,11 +221,16 @@ def get_chromedriver_location():
         assets_path = "{}{}assets".format(workspace_path, native_slash)
         validate_path(assets_path)
 
-        msg = ("Oops! Please, put chromedriver executable to the \"{}\" folder"
-               " and start again :]"
-               .format(assets_path))
-        raise InstaPyError(msg)
+        CD = binary_path
+        message = "Using built in instapy-chromedriver executable."
```
uluQulu: Any chance to also print the version in use? 😌
```erlang
"Using built-in 'instapy-chromedriver' executable (version 2.44)"
```

uluQulu:
I've just finished testing.
It works flawlessly!

Good job, @scriptworld-git ☕
I also like the idea of local **chromedriver** having higher priority than **pip**ped **chromedriver**.


I thought you were to do extra logic like we discussed in your initial PR about auto-chromedriver..
But this solution is pretty neat, simple & straight-forward and can stay around as long as sophisticated controller gets ready.

And this way,
- users can get the desired version of **chromedriver** binary by installing the same version of **instapy-chromedriver** package by **pip** [per their _python_ version].
- optionally, can manually download and put the **chromedriver** binary into the **assets** folder [at their **workspace**] and then _InstaPy_ will **always** use it.

That little part should be noted in Readme somewhere, so that users [mostly newbies] can easily understand how it works under the hood.

---

Cheers 😁
---

breuerfelix:
Just found a way on how to print the version. I will need the pkg_resources package which ships with setuptool package.  
Am I right that I don't need to mention the setuptools package in the requirements.txt since it gets installed with python and pip ?  

Since I just broke my local python version this morning.. I need some more hours to fix that, Im gonna update during the day and hopefully we get that new packaging up soon :)

instapy/file_manager.py
breuerfelix: done :))

instapy/file_manager.py
```diff
@@ -222,7 +223,8 @@ def get_chromedriver_location():
         validate_path(assets_path)
 
         CD = binary_path
-        message = "Using built in instapy-chromedriver executable."
+        chrome_version = pkg_resources.get_distribution("instapy_chromedriver").version
+        message = f"Using built in instapy-chromedriver executable (version {chrome_version})"
```
uluQulu: _f-strings_ added to python with v3.6 and is not backwards compatible.
You should use the `format()` function to format strings.

breuerfelix: done :))

timgrossmann:
@scriptworld-git Awesome one! 🙌 
Thank you!

@uluQulu Thanks for the review 😉

PR #3840, opened 2019-01-27
Added Config Manager, Release, New Directory Structure

antwal:
The ConfigManager allows us to avoid any kind of modification within the InstaPy library even if it is installed as a system user and used as a user.

For more information on paths, refer to: https://github.com/ActiveState/appdirs

Currently I am still making changes to this last revision where the Workspace was introduced; in the previous version the ConfigManager worked perfectly and allows you to easily manage many accounts.

I'm using it to manage 36 accounts simultaneously in one application.

The main configuration file is created automatically if it does not exist and the structure is very simple.

An example file can be found in configmanager_default.py, to understand how to use it.

I changed InstaPy's initialization method in this way it was possible to manage multiple accounts from a single application.

Soon I will give you an example of using multiple accounts using the schedule library


HCWcoder:
@antwal  wow that is so interesting !

we definitely need more instruction for this amazing add-on ! 

thanks !

antwal:
@timgrossmann @HCWcoder 

I'm updating the code with the change made in PR #3829, as it is very useful. 

I divided the sections in the configuration file for convenience and dynamism in use:

**Section [instapy]**
there is all the basic configuration, for the management of folders, logs and database, currently also the list of instagram accounts enabled in the bot

**Section [selenium]**
The section is dedicated to all common parameters used by selenium for chrome or firefox

**Section [ig_default] or [ig_ <username_instagram>]**
If you need to manage a single account by default the [ig_default] section is created, but if you want to manage 2/4 accounts, just create the appropriate sections [ig_account1], [ig_account2], etc.

Within the account section I put all the parameters that I considered most useful, such as the use of different proxies for each account.

In this way, once InstaPy is initialized, it is sufficient to choose by code which account to use, so that the dedicated parameters are loaded.

The Config Manager, if modified correctly, will be able to manage multiple configuration files in the future; for example, even if this comes from an extension

To load the Config Manager in any part of the code, just use:
`from instapy.tools import config`

To get the values, just use:
`config.instapy[param], config.selenium[param], config.account[param]`

for third party parameters, refer to:
`config.misc[param]`

An example with standard initialization:
`instance = InstaPy(config_store=True)`

the file config, in my case, on Mac OSX, is generated in the user home directory:
`/Users/antwal/.instapy`

the application folder of InstaPy is this:
`/Users/antwal/Library/Application Support/InstaPy`

The configuration file generated is this:
```
[instapy]
assets_dir = /Users/antwal/Library/Application Support/InstaPy/assets
console_logs = True
data_dir = /Users/antwal/Library/Application Support/InstaPy
db_dir = /Users/antwal/Library/Application Support/InstaPy/db
extensions_path = /Users/antwal/Projects/InstaPy/instapy/extensions
ig_accounts = {'default'}
log_level = None
logs_dir = /Users/antwal/Library/Application Support/InstaPy/logs
nogui = False

[selenium]
browser_profile_path = None
disable_image_load = False
headless_browser = False
proxy_chrome_extension = None
selenium_local_session = True
use_firefox = False

[ig_default]
bypass_suspicious_attempt = False
bypass_with_mobile = False
page_delay = 25
password = insta_password
proxy_address = None
proxy_port = 0
save_logs = True
username = insta_username
```

**Important, be careful** 
the root path of InstaPy Library is: 
`/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/instapy/`

the "root_path" parameter is not included in the config file because it is loaded every time the config is initialized

antwal:
I Have some problems with tox tests, but it working... Try fix it.

antwal:
@timgrossmann @HCWcoder 

Added example and changes to manage multiple accounts

TOX Tests Failed only with Python 2.7, 3.4, 2.7.6

antwal:
@timgrossmann I answer you here, that the other PR has been merged. Yes, I'll load all the changes, then when you want, you can revise them; in the end at the moment I should do as I always did that I update the code with important changes to instapy. Any conflicts or errors with Python 2.7, 3.4 for the moment I ignore them

timgrossmann:
@antwal Perfect, sounds good! 👍
Thank you

quickstart_templates/configmanager_multi.py
```diff
@@ -0,0 +1,83 @@
...
+        account1.set_dont_include(['user1', 'user2'])
+        account1.set_dont_like(["pizza", "#store"])
+        account1.like_by_tags(["kiss"], amount=10)
+
+        account1.end()
+    except InstaPyError:
+        pass
+
+
+def insta_account2():
```
CharlesCCC, line 57: I don't think this is the best approach to have multiple account. this seems a lots of duplicated code and it required to copy and paste to add additional account ?  I would just stick with the currently approach by passing argument to quickstart.py with multiple CMD/Terminal window open (that will be much easier).

antwal: I can not currently send the changes; however using the command line is integrated but I have not published the thing.

I am currently converting all the code into individual extensions that can be modified or overwritten by other user extensions.

I have also added server mode instead of using a multi-process via terminal

antwal: on this product [neo-z83](http://minix.com.hk/it/products/neo-z83-4) I can easily run 28 accounts via proxy without problems; I'm testing the latest version in server mode with a graphical interface using only 2 lines of code:
```
import instapy
instapy.server.initialize()
```
as soon as possible I create a separate branch; so you can try it and change the code if you find it useful

antwal:
I can close it, since I have changed almost everything and the code in this PR has been adapted to the previous commit

Issue #3920, opened 2019-02-06
ImportError: No module named 'instapy_chromedriver'


nikospaz:
Traceback (most recent call last):
  File "quickstart.py", line 3, in <module>
    from instapy import InstaPy
  File "/home/pi/Projects/InstaPy/instapy/__init__.py", line 3, in <module>
    from .instapy import InstaPy
  File "/home/pi/Projects/InstaPy/instapy/instapy.py", line 71, in <module>
    from .browser import set_selenium_local_session
  File "/home/pi/Projects/InstaPy/instapy/browser.py", line 16, in <module>
    from .file_manager import get_chromedriver_location
  File "/home/pi/Projects/InstaPy/instapy/file_manager.py", line 11, in <module>
    from instapy_chromedriver import binary_path
ImportError: No module named 'instapy_chromedriver'

In raspberry pi i get this error

aliceEddie:
Try this: #3907

MalteIwanicki:
same problem

aliceEddie:
from @PhilippSPOTTERON
> Thanks that was very helpful!
> For everybody having the same issue:
> With the new package, you just need to do the following steps:
> 
>     * Installed new version by "pip3 install instapy" on Debian Stretch (Raspberry Pi),
> 
>     * go to /home/pi/InstaPy/assets
> 
> 
> > from https://github.com/timgrossmann/InstaPy/blob/master/docs/How_to_Raspberry.md
> 
>     * just do the following steps:
> 
> 
>     1. wget https://github.com/electron/electron/releases/download/v3.0.0-beta.5/chromedriver-v3.0.0-beta.5-linux-armv7l.zip
> 
>     2. unzip chromedriver-v3.0.0-beta.5-linux-armv7l.zip
> 
>     3. chmod 755 chromedriver
> 
>     4. chmod +x chromedriver
> 
>     5. sudo apt-get remove chromium
> 
> 
>     * edit thenameofyourscripttemplate.py and add your username and password
> 
>     * Run Instapy by "python3 thenameofyourscripttemplate.py" in the right directory.
> 
> 
> If you want to have your script executed by crontab automatically, do not forget to add "headless_browser=True" in it, otherwise there will by another chromdriver error!
> 
> Hope that helps others, too,
> thanks again for solving.

my installation was at /home/cherry/pi/.local/lib/python3.7/site-packages/ for some reason. But the explanation above should fix it

nikospaz:
[...]

PR #3929, opened 2019-02-07
Only import instapy-chromedriver when necessary

breuerfelix:
Accoring to #3920 people have issues when using their own chromedriver, since they fail on importing instapy-chromedriver package.

I don't know why this happens, since when installing instapy on Pi, it should also install the chromedriver package. The chromedriver might be the wrong version but it should still install the package.

Well but it still happens so user have to uncomment the import for instapy-chromedriver when using their own chromedriver.  
I just moved the import down where I use the Path so if you put your own chromedriver in the assets folder, it won't even try to import the one from the package.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,12 @@ The **goal** of this file is explaining to the users of our project the notable 
 
 _The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)_.
 
+## [0.1.4] - INSERT MERGING DATE HERE
+### Fixed
+- Only import instapy-chromedriver package when needed
 
-## [0.1.3] - UNKNOWN
+
+## [0.1.3] - 2019-02-05
 ### Fixed
 - Fix "_Failed to load desired amount of users!_" issue.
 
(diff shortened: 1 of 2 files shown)


timgrossmann:
@scriptworld-git Good idea, having the import only when it really is wanted might avoid a lot of problems! 👍 

I think since it's kind of "common" to locate imports at the top of the files it seems weird at first (at least for me) but I've seen that structure in some other python libs before.

We might want to do this, too with other deps that only get used in very specific places 🤔 

We'll merge this into dev for now, there are a few other PRs open already that might get merged in the coming days and since it is not something extremely critical we won't make a bug fix release out of this for now, or do you think we should?

Could you please quickly add a section in the Changlog file? 😋

timgrossmann (approved):

breuerfelix:
@timgrossmann damn I forgot about that we would need a new PyPI version for that !

its okay merging this into dev for now :) just realised that pi users now are able to use the chromedriver version 2.36! really glad that this works easy on pi.

also this 'only import when needed' technique saves a lot of RAM :) since dependencies are only loaded into RAM when needed. maybe this makes it a little bit slower during runtime, but with a bot that got plenty of delay built in, I would rather go with the low RAM :))

timgrossmann:
@scriptworld-git True 😄 

Good to hear that!
Thanks for the PR 👍

Issue #3902, opened 2019-02-05
Bug fix: Stale element


timgrossmann:
#### One very common error is `stale element reference: element is not attached to the page document`. The biggest problem with this error is that it is non-deterministic and sometimes appears and fades with the next execution. 

Advantages of this feature:
- Error proneness when discovering a stale element exception

Possible ways of achieving:
- A decorator is added to each method automatically catching this error and triggering a reload of the page (which in most cases will solve the error). 
- This is done up to three times. If the error persists after that, an actual error is thrown and the tool still stops.
- The decorator wraps the whole method which won't add much code to the tool but only one line on the top e.g. `@stalereload(....)`. 

> [Bountysource this Issue](https://www.bountysource.com/issues/69650593-bug-fix-stale-element)

--- 

###### Feel free to comment with adjustments and improvements

sionking:
I think it should be a part of browser file - class.

timgrossmann:
@sionking Good idea, this will fit nicely into #3581

timgrossmann:
#3962

PR #3931, opened 2019-02-07
Fix get links for username

sionking:
We shall reload and refetch the dom element on StaleElementReferenceException

### DIFF
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -404,7 +404,7 @@ def get_links_for_username(browser,
     # then do not navigate to it again
     web_address_navigator(browser, user_link)
 
-    if "Page Not Found" in browser.title:
+    if not is_page_available(browser, logger):
         logger.error(
             'Intagram error: The link you followed may be broken, or the '
             'page may have been removed...')
@@ -424,6 +424,8 @@ def get_links_for_username(browser,
     # Get links
     links = []
     main_elem = browser.find_element_by_tag_name('article')
+    elem_expresstion = "browser.find_element_by_tag_name('article')"
+    main_elem = eval(elem_expresstion)
     posts_count = get_number_of_posts(browser)
     attempt = 0
 
@@ -444,7 +446,7 @@ def get_links_for_username(browser,
 
         # using `extend`  or `+=` results reference stay alive which affects
         # previous assignment (can use `copy()` for it)
-        links = links + get_links(browser, person, logger, media, main_elem)
+        links = links + get_links(browser, person, logger, media, main_elem, elem_expresstion)
         links = sorted(set(links), key=links.index)
 
         if len(links) == len(initial_links):
@@ -733,9 +735,20 @@ def get_tags(browser, url):
     return tags
 
 
-def get_links(browser, page, logger, media, element):
+def get_links(browser, page, logger, media, element, elem_expression=None):
     # Get image links in scope from hashtag, location and other pages
-    link_elems = element.find_elements_by_tag_name('a')
+    for i in range(2):
+        try:
+            link_elems = element.find_elements_by_tag_name('a')
+            break
+        except StaleElementReferenceException:
+            # notifiy
+            logger.error("link_elems StaleElementReferenceException")
+            # handle
+            if elem_expression:
+                reload_webpage(browser)
+                element = eval(elem_expression)
+
     sleep(2)
     links = []
     try:
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -422,11 +422,7 @@ def get_links_for_username(browser,
     elif following_state == 'Follow':
         wait_for_valid_authorization(browser, username, logger)
 
-    if following_state == 'Following':
-        following = True
-    else:
-        following = False
-    is_private = is_private_profile(browser, logger, following)
+    is_private = is_private_profile(browser, logger, following_state == 'Following')
     if (is_private is None
             or (is_private is True and following_state not in ['Following', True])
             or (following_state == 'Blocked')):
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -423,7 +423,6 @@ def get_links_for_username(browser,
 
     # Get links
     links = []
-    main_elem = browser.find_element_by_tag_name('article')
     elem_expresstion = "browser.find_element_by_tag_name('article')"
     main_elem = eval(elem_expresstion)
     posts_count = get_number_of_posts(browser)
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -88,16 +88,17 @@ def get_links_for_location(browser,
     sleep(1)
 
     if skip_top_posts:
-        main_elem = browser.find_element_by_xpath('//main/article/div[2]')
+        elem_expression = '''browser.find_element_by_xpath('//main/article/div[2]')'''
     else:
-        main_elem = browser.find_element_by_tag_name('main')
+        elem_expression = '''browser.find_element_by_tag_name('main')'''
 
+    main_elem = eval(elem_expression)
     link_elems = main_elem.find_elements_by_tag_name('a')
     sleep(1)
 
-    if not link_elems:  # this location does not have `Top Posts` or it
-        # really is empty..
-        main_elem = browser.find_element_by_xpath('//main/article/div[1]')
+    if not link_elems:  # this location does not have `Top Posts` or it really is empty..
+        elem_expression = '''browser.find_element_by_xpath('//main/article/div[1]')'''
+        main_elem = eval(elem_expression)
         top_posts = []
     sleep(2)
 
@@ -128,7 +129,7 @@ def get_links_for_location(browser,
         # keeps counted for the location
 
     # Get links
-    links = get_links(browser, location, logger, media, main_elem)
+    links = get_links(browser, location, logger, media, main_elem, elem_expression)
     filtered_links = len(links)
     try_again = 0
     sc_rolled = 0
@@ -153,7 +154,7 @@ def get_links_for_location(browser,
 
             sleep(3)
             links.extend(
-                get_links(browser, location, logger, media, main_elem))
+                get_links(browser, location, logger, media, main_elem, elem_expression))
 
             links_all = links  # uniqify links while preserving order
             s = set()
@@ -185,12 +186,13 @@ def get_links_for_location(browser,
                         try_again = 0
                         sleep(10)
 
-                        main_elem = (browser.find_element_by_xpath(
+                        elem_expression = '''(browser.find_element_by_xpath(
                             '//main/article/div[1]') if not link_elems else
                                      browser.find_element_by_xpath(
                                          '//main/article/div[2]') if
                                      skip_top_posts else
-                                     browser.find_element_by_tag_name('main'))
+                                     browser.find_element_by_tag_name('main'))'''
+                        main_elem = eval(elem_expression)
                     else:
                         logger.info(
                             "'{}' location POSSIBLY has less images than "
@@ -239,15 +241,18 @@ def get_links_for_tag(browser,
     sleep(1)
 
     if skip_top_posts:
-        main_elem = browser.find_element_by_xpath('//main/article/div[2]')
+        elem_expression = '''browser.find_element_by_xpath('//main/article/div[2]')'''
     else:
-        main_elem = browser.find_element_by_tag_name('main')
+        elem_expression = '''browser.find_element_by_tag_name('main')'''
+
+    main_elem = eval(elem_expression)
     link_elems = main_elem.find_elements_by_tag_name('a')
     sleep(1)
 
     if not link_elems:  # this tag does not have `Top Posts` or it really is
         # empty..
-        main_elem = browser.find_element_by_xpath('//main/article/div[1]')
+        elem_expression = '''browser.find_element_by_xpath('//main/article/div[1]')'''
+        main_elem = eval(elem_expression)
         top_posts = []
     sleep(2)
 
@@ -292,7 +297,7 @@ def get_links_for_tag(browser,
     # counted for the tag
 
     # Get links
-    links = get_links(browser, tag, logger, media, main_elem)
+    links = get_links(browser, tag, logger, media, main_elem, elem_expression)
     filtered_links = len(links)
     try_again = 0
     sc_rolled = 0
@@ -316,7 +321,7 @@ def get_links_for_tag(browser,
                 # you sent scoll command...
 
             sleep(3)
-            links.extend(get_links(browser, tag, logger, media, main_elem))
+            links.extend(get_links(browser, tag, logger, media, main_elem, elem_expression))
 
             links_all = links  # uniqify links while preserving order
             s = set()
@@ -348,12 +353,13 @@ def get_links_for_tag(browser,
                         try_again = 0
                         sleep(10)
 
-                        main_elem = (browser.find_element_by_xpath(
+                        elem_expression = '''(browser.find_element_by_xpath(
                             '//main/article/div[1]') if not link_elems else
                                      browser.find_element_by_xpath(
                                          '//main/article/div[2]') if
                                      skip_top_posts else
-                                     browser.find_element_by_tag_name('main'))
+                                     browser.find_element_by_tag_name('main'))'''
+                        main_elem = eval(elem_expression)
                     else:
                         logger.info(
                             "'{}' tag POSSIBLY has less images than "
@@ -420,7 +426,7 @@ def get_links_for_username(browser,
     if following_state is None:
         wait_for_valid_connection(browser, username, logger)
     elif following_state == 'Follow':
-        wait_for_valid_authorization(browser, username, logger)
+        wait_for_valid_authorization(browser, username, logger, notify=False)
 
     is_private = is_private_profile(browser, logger, following_state == 'Following')
     if (is_private is None
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -14,6 +14,8 @@ from .util import get_number_of_posts
 from .util import get_action_delay
 from .util import explicit_wait
 from .util import extract_text_from_element
+from .util import wait_for_valid_connection
+from .util import wait_for_valid_authorization
 from .quota_supervisor import quota_supervisor
 from .unfollow_util import get_following_status
 
@@ -411,14 +413,24 @@ def get_links_for_username(browser,
         return False
 
     # if private user, we can get links only if we following
-    following, follow_button = get_following_status(browser, 'profile',
+    following_state, follow_button = get_following_status(browser, 'profile',
                                                     username, person, None,
                                                     logger, logfolder)
-    if following == 'Following':
+
+    if following_state is None:
+        wait_for_valid_connection(browser, username, logger)
+    elif following_state == 'Follow':
+        wait_for_valid_authorization(browser, username, logger)
+
+    if following_state == 'Following':
         following = True
+    else:
+        following = False
     is_private = is_private_profile(browser, logger, following)
-    if (is_private is None) or (is_private is True and not following) or (
-            following == 'Blocked'):
+    if (is_private is None
+            or (is_private is True and following_state not in ['Following', True])
+            or (following_state == 'Blocked')):
+        logger.info('This user is private and we are not following')
         return False
 
     # Get links
--- a/instapy/util.py
+++ b/instapy/util.py
@@ -2093,3 +2093,29 @@ class CustomizedArgumentParser(ArgumentParser):
         """
         return []
 
+def wait_for_valid_connection(browser, username, logger):
+    while True:
+        sirens_wailing, emergency_state = emergency_exit(browser, username, logger)
+        if sirens_wailing and emergency_state == "not connected":
+            logger.warning('there is no valid connection')
+            sleep(60)
+        else:
+            break
+
+
+def wait_for_valid_authorization(browser, username, logger):
+    # save current page
+    current_url = get_current_url(browser)
+
+    # stuck on invalid auth
+    auth_method = "activity counts"
+    while True:
+        login_state = check_authorization(browser, username, auth_method, logger)
+        if login_state is False:
+            logger.warning('not logged in')
+            sleep(60)
+        else:
+            break
+
+    # return to original page
+    web_address_navigator(browser, current_url)


timgrossmann:
@sionking Good quick fix to take on #3902 

Don't you think we should remove passing the element itself and only keep the element expression? 🤔 
Having the element itself seems to be obsolete then, what do you think?
Searching the whole codebase, the `get_links` method only gets used in the like_util file

sionking:
@timgrossmann But the find element by tag only done once, if you look it up in the get_links you will end up triggering it over and over again.

timgrossmann:
@sionking Sorry I'm not sure if I understand what you mean.
If we use:
```python
def get_links(browser, page, logger, media, elem_expression):
     element = eval(elem_expression)
     for in range(2):
         try:
             link_elems = element.find_elements_by_tag_name('a')
             break
         except StaleElementReferenceException:
             # notifiy
             logger.error("link_elems StaleElementReferenceException")
             # handle
             if elem_expression:
                 reload_webpage(browser)
                 element = eval(elem_expression)
```

I don't understand the problem when just passing in the element_expression and getting the element in the method only

Could you explain that to me, please?

sionking:
get_links is called from a while loop.
the  element = eval(elem_expression) is done before the while loop.

timgrossmann:
@sionking Ok I see.
We should probably add the `elem_expression` to all the methods using `get_links`not only the `get_links_for_username` method, shouldn't we?

Thank you

sionking:
@timgrossmann @uluQulu 
Please check additional commit regarding waiting for valid connection / valid auth state.

sionking:
@timgrossmann please tell me what you think of the approch.
I think we should stuck and waqit for admin when no connection / not logged in.

Maybe it can be checked in other places.

sionking:
> `get_links`not

done,
Please merge this before merging other features I don't have time for diffs.
If the changes suite all it is better to merge.

instapy/like_util.py
```diff
@@ -86,16 +88,17 @@ def get_links_for_location(browser,
...
+    main_elem = eval(elem_expression)
     link_elems = main_elem.find_elements_by_tag_name('a')
     sleep(1)
 
-    if not link_elems:  # this location does not have `Top Posts` or it
-        # really is empty..
-        main_elem = browser.find_element_by_xpath('//main/article/div[1]')
+    if not link_elems:  # this location does not have `Top Posts` or it really is empty..
+        elem_expression = '''browser.find_element_by_xpath('//main/article/div[1]')'''
+        main_elem = eval(elem_expression)
```
uluQulu, line 101: ⚠ `eval()` is **dangerous**, please don't use it in a production code 🥺

instapy/util.py
```diff
@@ -2093,3 +2093,29 @@ def _get_option_tuples(self, option_string):
...
+def wait_for_valid_authorization(browser, username, logger):
+    # save current page
+    current_url = get_current_url(browser)
+
+    # stuck on invalid auth
+    auth_method = "activity counts"
+    while True:
+        login_state = check_authorization(browser, username, auth_method, logger)
+        if login_state is False:
+            logger.warning('not logged in')
```
uluQulu, line 2115: `check_authorization(notify=True)` already prints that message;
```python
logger.critical(
                    "--> '{}' is not logged in!\n".format(username))
```

But there is another problem,
You are repeatedly checking authorization state in case of it is not logged in.

And in case of it **really** is not logged in, that loop will never end...

instapy/util.py
```diff
@@ -2093,3 +2093,29 @@ def _get_option_tuples(self, option_string):
         """
         return []
 
+def wait_for_valid_connection(browser, username, logger):
+    while True:
+        sirens_wailing, emergency_state = emergency_exit(browser, username, logger)
+        if sirens_wailing and emergency_state == "not connected":
+            logger.warning('there is no valid connection')
```
uluQulu, line 2100: `ping_server()` already prints that message;
```python
        logger.critical(
            "There is no connection to the '{}' server!".format(host))
```

Again, in case of a **real** connection outage, it will never leave that loop.

instapy/util.py
sionking: That is exactly what we want to achieve. endless loop on not logged in.
This issue need Admin involved usuelly happen when a user change his password during runtime,
you dont want to unfollow/follow in this case.
Worst case is unfollowing -> follow_status is "Follow" which is seems that the users unfollowed before and will be removed from followpool but he is just logged out.

instapy/like_util.py
sionking: What is your suggestion ?
I think Eval in this case can be just fine.

instapy/util.py
sionking: again, you don't want to "eat" your sources when a connection error occur, better stuck in an endless loop until network back to normal.

instapy/util.py
CharlesCCC: Instead of looping forever, I would recommend retry 3~5 times and then break the loop.

sionking: @CharlesCCC But then what ?ANy of you are using real tools ? if the proxy / login is broken the user status is **STUCK** .

sionking: I do aggree notify(logger) should done once. it can be changed.

instapy/like_util.py
uluQulu: @sionking 

`eval()` built-in function alongside `exec` statement and `exec()` function or their low level parent `compile()` built-in with them must be avoided at max expenses in a general codebase.

In the very first unexpected source, it will execute it and then booom 🌋
But good news is, almost always, there are alternative safer ways you can find to avoid using them.

My suggestion is using normal methods.
Your code,
```python
elem_expression = '''browser.find_element_by_xpath('//main/article/div[1]')''' 
# 1st call
main_elem = eval(elem_expression)

# 2nd call for/in get_links(elem_expression=None)
links = get_links(elem_expression)
element = eval(elem_expression)
```
It can see a little bit less characters spent to achieve it.
Yeah but at the same time it is,
- Highly dangerous ⚡
- Confusing: _Maintainer always guesses what is going on and needs to return back the read the source of that expression passed to `eval()`_..


My suggestion,
```python
main_elem_XPath = "//main/article/div[1]"
# 1st call
main_elem = browser.find_element_by_xpath(main_elem_XPath)

# 2nd call for/in get_links(main_elem_XPath=None)
links = get_links(main_elem_XPath)
element = browser.find_element_by_xpath(main_elem_XPath)
```
So your easy style was intended for easing multiple calls,
by using
`element = eval(elem_expression)`
to replace a bit longer call,
`element = browser.find_element_by_xpath(main_elem_XPath)`
But well :octocat: 
You should know that people who still [decide to] use `eval()` or `exec`/`exec()` has very specific reasons that can hardly be achieved by other styles.
And your case is really not one of them.

Apart from that, using my suggestion,
- Code and execution is super healthy as expected;
- Easily understood by other programmers - nothing special can exist since it is a structured context;

Also, actually, 
`element = browser.find_element_by_xpath(main_elem_XPath)`
is not a long code.
It is just a normal line of code.
You want to get a the reference to the main element at real time and there is no other way of doing it rather than calling that line.
So, I can't think of if it was a bit longer and can be shortened..

If you think, we can open a new naming class - namespace and shorten those names e.g. as,
`browser.find_element_by_xpath` becomes `fex` 😆
but yeah it is not really needed.

Or another suggestion can be,
```python
def get_main_element():
    main_elem_XPath = "//main/article/div[1]"
    return browser.find_element_by_xpath(main_elem_XPath)


# 1st call
main_elem = get_main_element()

# 2nd call for/in get_links()
links = get_links()
element = get_main_element()
```    


---

Cheers 😁
---

instapy/util.py
uluQulu: @sionking 
>That is exactly what we want to achieve. endless loop on not logged in.

If you want to know my point, endless loop is simply illogical in all aspects.
That shows the weakness of a code.
I WOULD never try to achieve it, at all.

>This issue need Admin involved usuelly happen when a user change his password during runtime,

User changed password? 
Fine. **End** the session cos whatever you do further is misleading.
_Such as, you manually log in and want it to continue == data read/write is literally broken_.

>Worst case is unfollowing -> follow_status is "Follow" which is seems that the users unfollowed before and will be removed from followpool but he is just logged out.

We have the function to check authorization.
Use it, enhance it, integrate it, upgrade it and then catch those real situations and end the session in case of a real failure.
But never do loop infinitively. It is opposite to logic xD

>@CharlesCCC But then what ?ANy of you are using real tools ? if the proxy / login is broken the user status is STUCK .

As above, verify the failure state in depth and then **end** the session in case of a real critical failure.
Apart from that, for your 'real tool',
- If your proxy/VPS system or any other system or connectivity tool/software is in inconsistent state, then look for changing it, fixing it, upgrading it and even if it is severe, quitting it.
Please, research on that and don't have its affect on general codebase with illogical statements.

---

Cheers 😁
---

instapy/util.py
uluQulu: @sionking my thoughts are the same for this as in https://github.com/timgrossmann/InstaPy/pull/3931#discussion_r256927907.

instapy/like_util.py
sionking: @uluQulu But please see other cases which are not xpath.
find_element_by_tag_name in line 93.
and some other cases which are not xpath.

We maybe can execute a function call and pass it as argument, but hi, eval fit nice here and works well.

instapy/util.py
sionking: @uluQulu 
Stupid unstable application just quit on an issue.
No internet ? quit
not loggin ? quit

But ok, quieting is ok for most application **if it start from where it ended** but
Starting over again in instapy will make the same actions over again, we don't save last state.

And please you should know better, once said your area has limited internet access, please fix your area of leaving it is in inconsistent state.

instapy/util.py
sionking: @uluQulu you stuck in the past, in the way you handle things, no one can contribute here.
If you think been the biggest code contributor make you the only ruler here, than welcome.

@timgrossmann  I can't try contribute here anymore, it mostly political what is going on with @uluQulu, it is easy to find spots that you don't like and speak tons on words regarding it like you saying something smart but saying nothing at all.

I became who I am today by listening to others and learning, been chief engineer at my company and having million dollar business at home. try to build things not destroy others.

Not saying those are my last words, but I do wish good luck for every one, moving to my next journey. find peace everyone.

timgrossmann: @sionking Sorry to hear that, I totally understand your point and appreciate your honesty with that. I can understand your point of view on @uluQulu style and especially his "love" to detail and correctness. Judging from my own POV and @CharlesCCC comment, we all 3 agree that just keeping the tool running and hoping that the logout/interrupt in connection will be handled somehow (and we wait for it to be resolved) is not the approach we'd like to take.
If there are some cases in which it display "logged out" other than actually being logged out, we should do a retry of (at most) up to three times with increasing delay times.
Same for the lost connection.
Of course the user/admin or whoever is responsible should be notified, I absolutely agree with you here since that is not a normal behaviour.

I've been doing a lot of dev with JavaScript in the past and considering `eval` it's the same there. You should avoid it all cost, there are a few options where it actually is suitable, but only if there is not a single way around it. Judging by the comments and usage in the code, we can avoid it here.

Again, I'm really sorry to see you feeling like that and I can totally understand it since you've contributed a lot to this project and invested many hours.
@uluQulu developed into the "ruler" in regards to code since he's doing tons of dev and knows more about the code base than most (probably all) the others. He also introduced many genius features and improvements that improved the quality of this tool immensely.

The potential for this tool is huge and even if I missed that in the beginning, having a stable and good thought through codebase is extremely important so code reviews will only tighten in the future with, at one point, having all decisions discussed to provide the best possible quality.

I really appreciate and respect your previous work and experience, as this project grows we want to make sure that we, at one point, reach the same quality the project you've worked at as chief engineer had.
It's absolutely your decision, I just wanted to say that you'll definitely be missed in this project. Wish you good luck for all your next journeys, maybe you'll find back to InstaPy some day in the future.

Thank you for everything you have done for this project and the community.
Sorry to see you go

CharlesCCC: @sionking Tons of respect to both you and @uluQulu, You made great contribution to this awesome InstaPy community, I would like to echo what @timgrossmann said and hate to see you go,  as any community, there will be conflicts and debate; I don't want pick the side, I'm here to learn and do the right thing(I knew, even the "right-thing" is could debateable, as couple years went by, I will look at the same code I wrote and says 'who wrote this piece of crap').   Regardless who wins the debate, as long as we still moving to the right-direction, which is helping others to learn and to grow, that is all the matters.   "Stay hungry, Stay foolish".  --- Steve jobs

sionking: Some improvement can be at it times.
For now best option IMO is to **not fail the user into wrong action**, in any cost.
Restarting the application is a bad idea, how can you not see that ?
If user have a connection error he should fix it, then instapy can continue working from where it stopped and not from the start, I have been working in this way for a long time.

Connectivity issue ? getting instapy into endless loop, call it "IDLE" mode ( or make a nice clean code that print IDLE may satisfy you all), receiving telegram msg, fixing the issue, instapy continue. not all is about right coding, but correct behavior of the application is more important, even if you need to use eval for it. I do real time FW in my profession writing in pure c, almost don't give a shit on how the code looks, only correct flow.

@uluQulu debate in a way that not make you feel comfortable, "lecturing" you about how things should be done, like he know shits, like he know you.
You have somethings to say ? please say it, but don't act as the teacher.

uluQulu: Hey @sionking 
Better behave your words!

I kept silence for couple of hours until now to maybe make you feel better but you, misunderstood it and does not worth a piece of respect.

If you think I live in old, you are wrong.

This is Python and you write in Python.
Python is a Turing ready language and there is no need for `eval()` or `exec`/`exec()`;
Only some rare cases that use it, e.g. `_namedtuples` or some template engines..
Noone else uses it.
No production code should include it.
I have also never used `eval()` so far.

Funnily, as _CPython_ 3.8.0.aplha0 has millions of lines in its source with only 120 uses of `eval()` inside;
You have simply added 8 uses of `eval()` into  _InstaPy_ codebase consisting of just 20553 lines of code, out of nothing.

You felt uncomfortable because you've stubbornly supported the flaws of your code.
Look, I run a good research before writing and get advice from professionals in the area.
And in particular, one of the ppl I took advice on usage of `eval()` is the 9th biggest contributor to _Python_.

>If you think been the biggest code contributor make you the only ruler here, than welcome.

I have never thought so, you have requested my review and I reviewed it as I ever do.
If you failed to be a 'professional' and handle my detailed review, it is your problem.
But it is obvious that for you quantity is important rather than quality since you've compared by the size of contribution.

>I became who I am today by listening to others and learning, been chief engineer at my company and having million dollar business at home. try to build things not destroy others.

If you were to listen to others, you'd also listen to my words rather than calling it uncomfortable. But yeah, for you, I have 'too' less programming experience but, note that, that doesn't make me not find the logical misbehaviours of your code.
Is it the way you review other your programmer's code in your organization?
Then I bet you are the chief of a farm rather than a software company.
Be a chief engineer or a chief programmer, that is what you can produce matters here.

WOW millions of bucks! Know what? - You made yourself worthless and as cheap as a piece of dollar, by that statement.
That's the difference between me and you.
I care about programming whilst you care about some bucks.

I told to Tim before, I am not too kind but I am very sincere.
And I've never meant to destroy others, rather I always defended myself.

>And please you should know better, once said your area has limited internet access, please fix your area of leaving it is in inconsistent state.

My country does not have an internet access problem. Last time I had internet access in phone but my laptop was not with me. But I see that you remember those details, very well.
Note that, it is you who talked about your 'real tools' having inconsistent internet connection.


I beg pardon from @timgrossmann and all community fellas who have followed this pitiful conversation if I have spoilt the community guidelines.
Feel free to delete this message, @timgrossmann, alongside my other messages you find 'intimidating'.


Now, @sionking, whether you leave InstaPy is real or not, that's debatable.

But I definitely find myself not well suited to this project.

I know that, because of me, counting this case, 4 [intermediate/advanced] programmers I can recall and their work have been declined in previous history.

And I feel obligated to leave cos I don't want any future potential contributor to abandon the project just because of **me**.


---
Thanks, @timgrossmann for your trust in me.
My python & general programming knowledge has been boosted by using _InstaPy_ that I can never forget.
Thank you very much, everybody, for all!

Cheers 😁
---

timgrossmann: @CharlesCCC Haha, same here. Always wondering about the code I wrote back then 😅 

@sionking After thinking about it, we might all be wrong in the approach we took with the failing/not failing debate. We have only very limited knowledge on how users use the tool. For users running it on a server using corncobs (since it's pretty weird and complex to run InstaPy 24/7 at the moment) it will definitely lead to problems (if several sessions get started RAM will fill up and running will not work anymore - so we have to implement checking if InstaPy is already running etc.) when enforcing the not fail logic you implemented here. If a session crashes, fine, they most likely won't even notice since they set it up and that's it.
For users running InstaPy on their local system it might be weird to see InstaPy crash if they walk upstairs with their device, quickly switching Wifis and then seeing the tool abort. 
Both situations, however, are just mere assumptions and enforcing one or the other is not the right approach with a tool that has this many ways of using it.
We definitely have to get some information on how users use it.

Regarding the not fail implementation, I still really like what you did, as mentioned in the very beginning, however I'd rather not enforce it on the users. Having this backed by a separate parameter and documentation telling the users that they can switch between "failure modes" nicely and understandably explained, I'm absolutely sold 🙌 (or even better the idle mode you mentioned 😉 )

I absolutely agree, correct behaviour is key. But having good and right code is also important, especially for an open source project that might attract more people to contribute. Who have to understand the code base before being able to add something.
This might be something cultural as well, here in Germany, people take code quality (at least for bigger projects) very seriously since maintainability is extremely important for long running projects. Even if eval works here, it still seems like it is not the right approach to choose (especially if we can avoid it and go with the regular way we used everywhere in the codebase).

I might be absolutely wrong, I'm not a python expert, but judging from what I know about eval in other languages it's an anti-pattern. 
I really respect your experience.

I'm sorry that he made you feel that way, having an emotional connection to code and work in combination with all of us not being native english speakers leads to a lot of discrepancies... I really don't think he means to be really rude or anything like that he's just sincere.
As mentioned before, I really don't like this kind of discussions and would heavily prefer something like @CharlesCCC did above. 

```
@CharlesCCC 
Instead of looping forever, I would recommend retry 3~5 times and then break the loop.
```

In the best case, this is enough to make a point and then we can continue from there on.

timgrossmann: Folks seriously, I really don't want to sounds harsh but I kind of get the feeling that this is getting too personal here...
Maybe you should leave it as is at the moment and avoid this thread for the next two days until Monday, then come back and read again where this discussion went.

I'm pretty sure this is where none of us wanted to end up with and neither accuses of someone living in an area that does not have internet coverage and lives in the past, nor emotionally discussing whether somebody is a good programmer.

Did this turn into some religious or cultural fight or what have I missed?
I'm serious, please come back on Monday and re-read this. 

Is this who you guys are and how your professional behaviour looks like? I know it's an open source project in your free time, but that is exactly the reason why we should stay together, and learn together and especially from each other... 

This is just a project, nobody will get rich from it or earn a nobel price for working on it.

sionking: @timgrossmann @uluQulu 
I am sorry if get any of you offended.
Have a nice week.

instapy/like_util.py
sionking: @timgrossmann 
Please tell me what is the prefer way instead of eval since @uluQulu suggested won't fit.
And still eval can be used for now, and changed later when refactoring.

instapy/util.py
timgrossmann: @sionking Thank you very much for this comment, I really appreciate it. 🙏

instapy/like_util.py
timgrossmann: @sionking I'll look over the PR now, sorry for the delay.

timgrossmann:
@sionking I'll write the comments down, let me know what you think 😊 

Regarding `get_links(browser, page, logger, media, element, elem_expression=None)`:
- Do we have any problem with passing in the xPath that basically is only the path used inside of the `elem_expression` variable? That way we could "simply" use a `browser.find_elements_by_tag_name(xPath)` were xPath is the passed argument

Regarding `wait_for_...` methods:
- Could we move them into the browser.util file to keep the separation of concern (since they look like "browser behaviour specific handling")
- and add another parameter, something like `idle_mode` which will be true to keep looping and, by default, false which will only loop 3 times? Something like `while idle_mode or i < 3:` should be good enough (please help me figure out if this makes sense)

Other than the comments above I like handling those problems / taking care of short outages by default. (I'd like to introduce a TCP style connection problem handling here, first sleep is 10sec, then 30sec, then 60sec if that is possible, let me know what you think about the delay times 👍)

---

In general, let's please remove the eval and instead stay with xPath in every case.
We can have queries like this:
` link_elems = main_elem.find_elements_by_tag_name('a')`
using xPath like this:
` link_elems = main_elem.find_elements_by_xpath('./a')`

Looking at performance it is a bit slower but in general since InstaPy is not build for high performance, the ms in difference won't make too much of a difference.

Please let me know your thoughts on the comments and what parts you might be able to refactor.
I'll try to find some time next week to refactor the remaining parts then.

sionking:
@timgrossmann 
I think it will be better to pass "XPATH"/ "CSS" / "ID" and to find element accordingly.
But still I don't see why not using eval and refactoring in the future.
I don't see why it is unsafe, beside you telling me that it is unsafe.
does it work ? yes. is it understandable ? yes.

sionking:
@timgrossmann 
Not looping forever will break the all point.
Have you ever worked with basic OS system? a good software should have some scheduling and task/event management and not only line-by-line. I think we should loop forever and later create an OS that will manage the running.

timgrossmann:
@sionking It's true, it does work and is kind of understandable. I mean it is my personal opinion but by looking at it, I ask myself, ok... Why are we using a string and then doing `eval` on that string rather than simply wiring the expression of that string right there instead of doing `eval`.

For me, the `eval` method introduces abstraction, but a wrong way of abstraction. We can just put everything into that method.

To explain why it is unsafe, consider a calculator that records the input of a user. Uses enter e.g. 1 + 2 + 3 and then hit the equal sign which will trigger `eval` on that expression. The solution will be 6 of course but if the user enters 
```python
import os
os.system('reboot')
```
The system the calculator is running on will reboot. It's just too abstract, there is no control behind it.
Of course this is completely different than a calculator and for those specific use case does not introduce any big problems, in my opinion it is still not the right way to go since users will see it and think, well this is quite nice we can abstract away a lot of things and then we will start to see this all over the code base.

Comparing that with xPath, which is a really nice abstraction. There is no query we can do with tag, css or id that can't be done with xPath, too.
In a speed critical application I'd consider not using it since it might introduce slower execution, but looking at the InstaPy code base and the amount of (wanted) sleeps and delays, that quite a good place for xPath without having any problem in case of performance.

What is the reason you want to keep the id, css and tag selectors?
Maybe I'm missing something.

Sorry for the long text 🙊

timgrossmann:
@sionking No , I haven't ever worked on OS systems. I'd actually like to at one point to get a better understanding of it and maybe what you're describing will make more sense to me.

I'd like to see what you're describing implemented, but let me give you an example for my concerns.
I've been running InstaPy on my DO server for quite a long time.
And one point I had system reboot as part of my quickstart file so that the sever would restart every now and then when InstaPy crashes or finishes because it would allocate more and more memory with every run which would lead to the tool being unable to execute at all after 2-3 times of encountering problems.

I don't want this behaviour back and I want the tool to finish in 1-2h then sleep 1-2h before starting again and again doing interactions.
@w4ngel Is working on features to make the daily limits more nicely distributed over the day so that it can run 24/7. Then might make sense to keep the tool up and running indefinitely. 
For now, when an error is encountered, I want my InstaPy session to crash (and do it nicely) and then just start again when the cronjob triggers it the next time.

Otherwise I would have to take into account InstaPy checking if there already is a session running on the server at the moment, we would have to take into account whether the session is somehow stuck etc. 

I've saw many people in the beginning that complained that the tool is "liking" 99 instead of 100 pictures and something like that. The problem here is that the best thing that can happen with this tool is that, if you know the code, everything makes sense (it's skipping post because of some set up restrictions etc). 
But from the outside it looks like the tool is non-deterministic. 
Instagram is not available? -> stop and try again later -> might lead to not being blocked
Instagram is not available? -> retry every x seconds/minutes until it works again and then continue  -> might lead to being blocked much earlier.

sionking:
@timgrossmann 
Let's take decisions here.

I will remove the wait for ... ( I will still use it in my branch), until a proper managed os will be suggested.

Regarding eval use, it been tested and working flawlessly, so I suggest to  merge it and later we can make it more modular, but if you prefer not, I will remove it also (but still use it in my branch).

timgrossmann:
@sionking sure, good to hear that 😊
Sorry for the delay.

I still think having this, optional, would be great. We just shouldn't make it on by default (or more like have it on by default to reload 3 times and then throw an error)

Also, please remove the eval.

Absolutely, feel free to use them in your branch. Maybe one day we will be happy to have both of them 👍 

I think the StaleElement problem will also be fixed using the decorator @breuerfelix wrote in the browser abstraction PR that is open. That way we can revert back all the changes to the get_links method.
So the idea of this PR should be to move the `wait_for_xxx` into the browser class making it available to the codebase and keeping them in your current PR.

Hope that works for you

instapy/like_util.py
```diff
@@ -404,26 +412,33 @@ def get_links_for_username(browser,
     # then do not navigate to it again
     web_address_navigator(browser, user_link)
 
-    if "Page Not Found" in browser.title:
+    if not is_page_available(browser, logger):
```
timgrossmann, line 415: We definitely want to keep the usage of this method 👌

instapy/like_util.py
```diff
@@ -404,26 +412,33 @@ def get_links_for_username(browser,
...
         logger.error(
             'Intagram error: The link you followed may be broken, or the '
             'page may have been removed...')
         return False
 
     # if private user, we can get links only if we following
-    following, follow_button = get_following_status(browser, 'profile',
+    following_state, follow_button = get_following_status(browser, 'profile',
```
timgrossmann, line 422: And this change with the section underneath handling the wait for and the is_private part

instapy/like_util.py
```diff
@@ -733,9 +748,20 @@ def get_tags(browser, url):
     return tags
 
 
-def get_links(browser, page, logger, media, element):
+def get_links(browser, page, logger, media, element, elem_expression=None):
```
timgrossmann, line 751: Revert this and later on just add the `retry()` decorator to this method

sionking: retry ?

sionking: ohh I see that in other PR.
I really don't understand what he did over there (I just guess in general)

timgrossmann: @sionking Sorry for the confusion. 
Basically what will happen after adding the `@retry()` decorator to a method is that if there should be any error, it will reload the same location and execute that method again. It will do this (default) 3 times and only then throw an actual error.
This will help us fight the StaleElement problem and most likely many others.

sionking: But it won't take effect here.
first we find the element by xpath later use the founds with find by tag.

timgrossmann: @sionking I'm sorry, I think I don't understand what you mean.
If there is a find by tag name, we can just replace it with xPath

sionking:
```
def get_links(browser, page, logger, media, elem_expression):
     element = eval(elem_expression)
     for in range(2):
         try:
             link_elems = element.find_elements_by_tag_name('a')
             break
         except StaleElementReferenceException:
             # notifiy
             logger.error("link_elems StaleElementReferenceException")
             # handle
             if elem_expression:
                 reload_webpage(browser)
                 element = eval(elem_expression)
```

@timgrossmann 
look at the line:
`element.find_elements_by_tag_name('a')`
this line will fail, and reloading will trigger this line again, but without featching the elemnet first `eval(elem_expression)` (which is not exist in this function) you will fail to find tag name.

breuerfelix:
> ```
> def get_links(browser, page, logger, media, elem_expression):
>      element = eval(elem_expression)
>      for in range(2):
>          try:
>              link_elems = element.find_elements_by_tag_name('a')
>              break
>          except StaleElementReferenceException:
>              # notifiy
>              logger.error("link_elems StaleElementReferenceException")
>              # handle
>              if elem_expression:
>                  reload_webpage(browser)
>                  element = eval(elem_expression)
> ```
> @timgrossmann
> look at the line:
> `element.find_elements_by_tag_name('a')`
> this line will fail, and reloading will trigger this line again, but without featching the elemnet first `eval(elem_expression)` (which is not exist in this function) you will fail to find tag name.

right for functions like these, i made this sweet decorator :) your function then will look like that:
```
@retry(2)
def get_links(browser, page, logger, media, elem_expression):
     element = eval(elem_expression)
     link_elems = element.find_elements_by_tag_name('a')
     return link_elems
```

what does this decorator ?
it will search for a variable which is of type browser in the args... it will save the current webpage (the one you are currently looking at)
if this function throws an exception (such as stale element exception) it will reload the page and execute it again, in total 2 times (since you passed 2 as a paramter)
this does exactly what you wanna do, but its abstract so you can use it for any function you want to :)

if you want that the decorator should not get the current page and instead use the page you provided you have to do it like that:
```
def my_funny_func():
    # dostuff
    page = 'blabla'

    @retry(2, page)
    def get_links(browser, logger, media, elem_expression):
         element = eval(elem_expression)
         link_elems = element.find_elements_by_tag_name('a')
         return link_elems

    my_elems = get_links(browser, logger, media, exprs)
    # do smth with my elems
```

btw @sionking thanks for providing this example since this is the PERFECT fit for the decorator ! @timgrossmann now you know where and how we can perfectly use it, and how clean it looks in the code :)

sionking:
@breuerfelix 
Thanks for good explanation, but you are looking at the code after I "fixed" it.

The original code is missing the element look up, as it given as parameter.
```
def get_links(browser, page, logger, media, element):
    link_elems = element.find_elements_by_tag_name('a')
    return link_elems
    ...
```

For this function reloading won't do any good since you need to find the element again.
@timgrossmann did you also get it ?

timgrossmann:
@sionking II'm sorry, maybe I'm just really dumb.
The retry method restarts the whole method again, true, meaning that with your example, in the words case we save half of the requests.
However we could get the same exact behaviour when making a sub-method which get's called in `get_links` which then has a `rety()`
Maybe we are discussing around one another points here.

sionking:
> @sionking II'm sorry, maybe I'm just really dumb.
> The retry method restarts the whole method again, true, meaning that with your example, in the words case we save half of the requests.
> However we could get the same exact behaviour when making a sub-method which get's called in `get_links` which then has a `rety()`
> Maybe we are discussing around one another points here.

so this sub method is the element fetching ? but this element is changing in a while loop, so you need to save which fetching is done last. too complicated I think.

breuerfelix:
added the util browser functions to my custom browser class :)

timgrossmann:
Changes have been merged into #3962 
Thank you very much for these proposals 😊

PR #3936, opened 2019-02-07
fixed get_active_users wrong behavior on videos

dmsfabiano:
**Old behavior:** 

1) Function was not taking care of videos correctly, given a video it will continue to the next iteration of the for loop without closing the current dialog box and opening the one for next image. This would cause for the function to get stuck in the same dialog box (which was a video) and never get any active users (`return = 0`).

2) After fixing [1] found another problem, the past implementation of for loops did not support the fact that a video was found. Even after adding the process to fix [1] (closing & opening the next dialog box), it will still iterate through the same list; therefore, there was no way to compare the actual number of posts requested by the user if a video was found.

**New Behavior:**

1) The function will always deliver the number of post comparisons requested by the `posts` parameter (unless `posts > total_posts`). This is done by not incrementing post count if a video is found.

2) Function properly closes the dialog box and opens the next one if video is found.


timgrossmann:
@dmsfabiano Thank you very much for this addition!

Could you explain what you mean by "not incrementing the post count if a video is found"?
Videos can be liked and commented, too. Shouldn't we treat them like regular posts? 🤔 
[Example](https://www.instagram.com/p/BVMSWmqlwd-/)

instapy/util.py
```diff
@@ -552,11 +555,28 @@ def get_active_users(browser, username, posts, boundary, logger):
                     likers_count = None
             try:
                 likes_button = browser.find_elements_by_xpath(
-                    "//button[contains(@class, '_8A5w5')]")[1]
+                    "//button[contains(@class, '_8A5w5')]")[0]
                 click_element(browser, likes_button)
-                sleep_actual(5)
```
uluQulu, line 557: Generally deleting the sleep lines added by others is cautious.
I think it was added by @sionking.

Well, do we need it?
Yes, a little bit of break is best before trying to fetch the usernames from the freshly opened dialog box.
But 5 is a little bit big for there, maybe 3 or 2?

instapy/util.py
```diff
@@ -552,11 +555,28 @@ def get_active_users(browser, username, posts, boundary, logger):
                     likers_count = None
             try:
                 likes_button = browser.find_elements_by_xpath(
-                    "//button[contains(@class, '_8A5w5')]")[1]
```
uluQulu, line 555: You have tested your custom case and CHANGED this part - the index.

But it is not solving better than prior.

There some cases and in one of them it returns 1 elements where your zero index wins.
Other case returns 3 elements where your index will fail cos the first element refers to unrelated location - 'Edit profile' whilst 2nd and 3rd elements wins.

What to do?

Check it, [-1] or [0].

dmsfabiano:
> @dmsfabiano Thank you very much for this addition!
> 
> Could you explain what you mean by "not incrementing the post count if a video is found"?
> Videos can be liked and commented, too. Shouldn't we treat them like regular posts? 🤔
> [Example](https://www.instagram.com/p/BVMSWmqlwd-/)

The function is meant to get the active users. To my limited Instagram knowledge, these are defined as the users that liked the last N posts. Even though you can like it, I didn't find a way in your example to get the likers, which is what we are interested on. Am I missing something to get the likers from a video?

instapy/util.py
dmsfabiano: I think 2 should be fine, I'll add this

instapy/util.py
dmsfabiano: I didn't think of the "Edit profile" case; good catch! I will add this. However. the index of [0] will definitely fail.

timgrossmann:
@dmsfabiano Double checked in, thank you for the explanation, you're right.
It seems like we can't even see the viewers 😅

instapy/util.py
uluQulu: I just saw my [original](https://gist.github.com/uluQulu/ea2211f9f818a4ca443de02f68a3f4b6) implementation and that sleep delay is 4 seconds in there.
image

Sleep has extra importance in this function cos of the smart scroller..

Let's make it 3 seconds 🤝🏼

dmsfabiano:
Please check #3936 feedback incorporated, fixed issue mentioned by @uluQulu  and also fixed another issue:  when the targeted account is followed by the account's user, a video will have a different length, now the function accounts for this and also empty case 👍

instapy/util.py
```diff
@@ -606,12 +637,7 @@ def get_active_users(browser, username, posts, boundary, logger):
                     sleep_actual(1.2)  # old value 5.6
                     sc_rolled += 1
 
-                """ Old method 1 """
-                # tmp_list = browser.find_elements_by_xpath(
-                #     "//a[contains(@class, 'FPmhX')]")
-
                 user_list = get_users_from_dialog(user_list, dialog)
-                # print("len(user_list): {}".format(len(user_list)))
```
uluQulu, line 614: This is my test line 😆
Thanks for deleting it xD

uluQulu (approved):
Good job, @dmsfabiano
Having lots of little counter variables makes difficult to me to diff the changes but finally could find time to check all cases.

In future that function is gonna take lots of refactors cos having bunch of handled blocks.
At that time, those little counters and their logic can be simplified.

You have done a good job to move `for` statement into `while` with added logic.
- Now, if users enter posts' count as 5 then it will always visit 5 posts and try to grab data [if there are at least 5 supported posts- now only photos or multiple-posts having video].
- Earlier, even in its **perfect** state implementation, it was to lose the iteration after encouters a video [yeah off that bug]. So with 5 post amount would visit 4 posts [as encountered a video]..

---

Those commented lines are mine cos last time I fixed it, I changed old grabbers and added working simpler alternative. But thought it is good to keep old methods, just it could be useful in future [their element's & finding them].

But I think that, we already have grabbed a good selenium knowledge, so it is good to delete those comments.

---
@timgrossmann, looks good to me 🎉

Cheers 😁
---

timgrossmann:
@uluQulu Thank you very much for reviewing! 😊 

@dmsfabiano Thank you very much for applying all the changes and adjusting you code.
I've added an entry to the changelog file.
Looking forward to your next PRs 😉 

Congrats to your first merged PR into InstaPy 😊

PR #3950, opened 2019-02-09
Proxy auth (Firefox)

converge:
Enable Proxy Authentication for Firefox.

I'm proposing this PR to test it a little more. I still need to write documentation for the feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,7 +211,8 @@ Sorry for the inconveniences
 * [Text Analytics](#text-analytics)
   *  [Yandex Translate API](#yandex-translate-api)
   *  [MeaningCloud Sentiment Analysis API](#meaningcloud-sentiment-analysis-api)
-* [Use a proxy](#use-a-proxy)
+* [Use a proxy (Chrome)](#use-a-proxy-chrome)
+* [Use a proxy (Firefox)](#use-a-proxy-firefox)
 * [Switching to Firefox](#switching-to-firefox)
 * [Emoji Support](#emoji-support)
 * [Clarifai ImageAPI](#clarifai-imageapi)
@@ -1906,16 +1907,18 @@ This project uses MeaningCloud™ (http://www.meaningcloud.com) for Text Analyti
 
 
 
-### Use a proxy
+### Use a proxy (Chrome)
 
 You can use InstaPy behind a proxy by specifying server address and port
 
+Simple proxy setup example:
 ```python
 session = InstaPy(username=insta_username, password=insta_password, proxy_address='8.8.8.8', proxy_port=8080)
 ```
 
-To use proxy with authentication you should firstly generate proxy chrome extension (works only with headless_browser=False unless using FF where it works with headless_browser=True).
+To use proxy with authentication you should firstly import proxy chrome extension to you configuration file (the one with your Instagram username and password).
 
+Proxy setup using authentication example:
 ```python
 from proxy_extension import create_proxy_extension
 
@@ -1923,6 +1926,27 @@ proxy = 'login:password@ip:port'
 proxy_chrome_extension = create_proxy_extension(proxy)
 
 session = InstaPy(username=insta_username, password=insta_password, proxy_chrome_extension=proxy_chrome_extension, nogui=True)
+```
+
+### Use a proxy (Firefox)
+
+You can use InstaPy behind a proxy by specifying server address, port and/or proxy authentication credentials. It works with and without ```headless_browser``` option.
+
+Simple proxy setup example:
+```python
+session = InstaPy(username=insta_username, password=insta_password, proxy_address='8.8.8.8', proxy_port=8080)
+
+```
+
+Proxy setup with authentication example:
+```python
+session = InstaPy(username=insta_username,
+                  password=insta_password,
+                  proxy_username='',
+                  proxy_password='',
+                  proxy_address='8.8.8.8',
+                  proxy_port=4444,
+                  use_firefox=True)
 ```
 
 ### Switching to Firefox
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1925,6 +1928,27 @@ proxy_chrome_extension = create_proxy_extension(proxy)
...
 ```
 
+### Use a proxy (Firefox)
+
+You can use InstaPy behind a proxy by specifying server address, port and/or proxy authentication credentials. It works with and without ```headless_browser``` option.
+
+Simple proxy setup example:
+```python
+session = InstaPy(username=insta_username, password=insta_password, proxy_address='8.8.8.8', proxy_port=8080)
+
```
uluQulu, line 1942: Also add `use_firefox` parameter or advice users to pass as a CLI flag?

uluQulu:
Very good change, @converge 💪🏼

I have tested and it works perfectly ☕

Actually, I liked the idea of passing proxy credentials- _username_ and _password_ that easily very much, 
```python
# inside quickstart script

session = InstaPy(proxy_username="abc-proxy", proxy_password="123-proxy", proxy_address="192.168
1.1", proxy_port=8080)
```
rather than bundling them before sending like it is for _Chrome_,
```python
# inside quickstart script

from proxy_extension import create_proxy_extension
proxy = "{}:{}@{}:{}".format(proxy_username, proxy_password, proxy_address, proxy_port)
proxy_chrome_extension = create_proxy_extension(proxy)
```
`proxy_chrome_extension` is just a string like this:
```dos
assets/chrome_extensions/proxy_auth_192.168.1.1:8080.zip
```

BTW while testing it, I found a big bug there, after switch to **workspaces**, I forgot to integrate `proxy_extension` module to use the **workspace** folders..

I will add support for it soon and thinks to move `proxy_extension` into the **instapy** package or make a new **extensions** package inside it and move there and so that it will also be added to the packaged distributions.
I am not a proxy user, that's why somehow missed that file 😐

---

Now the _Firefox_ webdriver will have support for proxy with authentication in both headless and graphical modes.
That will be great for ppl who wanted the headless mode and proxy authentication in _Chrome_, they can just switch to _Firefox_ until it would be supported in _Chrome_..

---

I think we have to remove the difficulty with _Chrome_ with the usage of 
```python
proxy_chrome_extension = create_proxy_extension(proxy)
```
It can be as it is for _Firefox_,
pass all of the parameters into the InstaPy constructor and in there, make a proxy extension file under the hood.
I can do it after this PR gets merged.

---

But another issue,
Now there will be 5 parameters in the InstaPy constructor regarding proxy,
```python
                 proxy_chrome_extension=None,
                 proxy_username=None,
                 proxy_password=None,
                 proxy_address=None,
                 proxy_port=None,
```
Okay, let's assume we will make a proxy extension file under the hood, then it is 4 parameters == still many;
Maybe in future we can make another function, e.g.,
```python
set_proxy(proxy_username, proxy_password, proxy_address, proxy_port)
```
that would run before instantiating the InstaPy and pre-configured proxy values could be stored in the `Settings` class, basically..

---

---
@timgrossmann looks good to me 

Cheers 😁
---

converge:
great review @uluQulu!

Reading your review, I just realized that this is becoming a big project, and our users will need a screen to manage it easily. Today we're growing non-stop, our users will get lost soon.

A screen/dashboard/gui/manager would help on this, I believe.

timgrossmann:
@uluQulu Thank you for the review!
Havin a `set_proxy` method will make this much nicer for sure! 
In there we can also handle whether we're using chrome or firefox and are able to throw an error if people try to combine headless chrome with authentication proxy. 🙌 
Will you move the proxy step to a separate method with the PR you've mentioned? 😊 

@converge So awesome, thank you very much for implementing this! @scriptworld-git is working on an extensible and dynamic GUI. I think ~end of march we will see the first version of that up and running with the first basic layout and features. From there on we can continue the work and apply user feedback to make it good/better.

README.md
timgrossmann: I've added it 😉

uluQulu:
Hey Tim,
>Will you move the proxy step to a separate method with the PR you've mentioned? 😊

Yes, I can do it since it's not a big work.

But rather than a method, it must be a standalone function cos it has to be started before instantiating _InstaPy_.
Why? - Cos, the _InstaPy_ instance will need the proxy data being ready to make a webdriver..

So, e.g.,
 - Adding a new **proxy_util.py** file directly into the **instapy** package;
 - To configure proxy data and then assign to the `Setting` class;
 - Also adding the **proxy_extension.py** (_used only for Chrome_) file's content to that new **proxy_util.py** file;

So that we **don't have to** make another package for that 1 function ~70 line code and distribute it as [in]/another package even it does not even has the type of 'extension'..

---

So in final,
adding it to the **\_\_init\_\_.py** of the **instapy** package,
```python
from .proxy_util import set_proxy
```
will make it easily importable into the _**quickstart** scripts_,
```python
from instapy import set_proxy

# add a proxy as your connection
set_proxy()

# get an InstaPy session!
session = InstaPy()

#...
```

---

Cheers 😁
---

timgrossmann:
@uluQulu Awesome 🙌 
Even though from looking at it, it would be sexy to see it somehow being passed into the InstaPy constructor 🤔
Looking at it like this seems to be completely separate and if we use the constructor, we don't have any session set up at that moment and InstaPy does not need to be internally wired to that proxy_util but we simply pass in this proxy element (that could also be changed to something else later on) and we have a more loosely coupled system.

Well maybe it's too much, but I'd like it more if we could "see" the interconnection between the two entities in the quickstart file 🙊

--- 

Hmm well, maybe not.
Let's see 🤔

PR #3952, opened 2019-02-09
Upated README.md

tE3m:
Updated the outdated `schedule` template.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2309,7 +2309,7 @@ insta_username=''
 insta_password=''
 
 #path to your workspace
-set_workspace=(path=None)
+set_workspace(path=None)
 
 def job():
   session = InstaPy(username=insta_username, password=insta_password)
--- a/README.md
+++ b/README.md
@@ -2312,23 +2312,20 @@ insta_password=''
 set_workspace=(path=None)
 
 def job():
-    try:
-        session = InstaPy(username=insta_username, password=insta_password)
-        with smart_run(session):
-          session.set_do_comment(enabled=True, percentage=20)
-          session.set_comments(['Well done!'])
-          session.set_do_follow(enabled=True, percentage=5, times=2)
-          session.like_by_tags(['love'], amount=100, media='Photo')
-    except:
-        import traceback
-        print(traceback.format_exc())
+  session = InstaPy(username=insta_username, password=insta_password)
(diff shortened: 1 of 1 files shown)


timgrossmann:
@TarekJ03 Nice! Good one.

I think we can also remove the try/catch block since the error handling is done inside the `smart_run` method. (correct me if I'm wrong @uluQulu 😉 )

Haha, awesome to see you also starting to contribute code changes, Tarek 🙌

tE3m:
I do what I can. I'll try adding features in the future, but for now, support is the most I can do. 😁

uluQulu:
Hey guys
congo on your first PR, @TarekJ03 🥳

@timgrossmann, you're right.

But the writer of those lines wanted to not break the code in any case.
That's why he/she catches it by `traceback.format_exc())` and prints it;
That prints the exception without breaking the code.

That style is useful for guys who run 24/7 scripts and want their code never stop even if it is not wise to do so with that **basic** handling cos they can stop **real** [else than **trivial**] problems and re-run code illogically..

You can see that style in Mehran's **quickstart** template, too - **follow_unfollow_and_send_telegram_msg.py**.

---

Good that you have removed it, @TarekJ03.
I think the official example of the '_schedule_' script should not use that **basic** style.
Only it should be included in some other example WHERE it could be called like a "_24/7 running script_".

---

Noting again, `smart_run` has its specific default error handling for general user:
- In case of _`NoSuchElementException`_, handle it helpfully providing data for debuggers;
- In case of _`KeyboardInterrupt`_, provide a clean and pretty exit (_we don't want to see the traceback here_);
- In case of other exception, **raise** it;


----
Looks good to me, @timgrossmann 

Cheers 😁
---

<sub>Could be great if spacing was 4.</sub>

uluQulu (approved):

tE3m:
Ohh, I even had "Indent using Tabs - Size: 4" configured, but I acidentially switched to "Indent using spaces"
🙃 Changed that now @uluQulu , thanks for the correction and approval 😁

timgrossmann:
@uluQulu Thanks for the review 😉 

Thanks for changing it, @TarekJ03! 
Congrats to your first PR into InstaPy (other than the incredible support you're providing of course 😉 ) 🙌

PR #3959, opened 2019-02-10
Implemented accepting pending follow requests for private account

Bersh:
Implemented `accept_follow_requests` function will accept all pending "follow" requests you may have in your private profile. 
If someone can test this on account with many(several pages) pending requests and tell if it works as expected I would appreciate it.


Bersh:
After checking account with many pending follow requests I found that there is pagination for that list. So I just implemented a maximum desired accepted accounts number as a parameter

instapy/instapy.py
```diff
@@ -4962,3 +4961,39 @@ def check_character_set(self, unistr):
...
+        web_address_navigator(self.browser, feed_link)
+
+        requests_to_confirm = self.browser.find_elements_by_xpath("//button[text()='Confirm']")
+
+        if len(requests_to_confirm) == 0:
+            self.logger.info("There are not follow requests in activity feed")
+
+        accepted = 0
+        for request in requests_to_confirm:
+            self.logger.info("Accepting {}".format(request.find_elements_by_xpath("./../../../../div[2]")[0]
```
timgrossmann: I think having such a specific xpath only for logging is not necessary.
This will probably lead to errors pretty soon, can we optimise this somehow? 😊

timgrossmann:
@Bersh Thank you very much for your PR.
Do you think we can replace that very specific xPath somehow?

instapy/instapy.py
Bersh: Indeed this xPath lokks bad. Unfortunately I didn't find a way to write it better :-( 
I have removed this log in the next commit

Bersh: see 6185445

Bersh:
I was wrong - pagination is present in pending requests list it is just done in the endless list manner. So I put back my page reload logic. Also checked on real account - works as expected

timgrossmann:
@Bersh Good to hear that it is working!
You've probably seen it, I've removed the instagram_endpoints.py from your PR.
We've moved that file to the research repository, I've already added your endpoint there. 👍 

Let's see if we can find another person to approve it.

One thing we still have to add is a bit of documentation in the README so users know that this method exists and how to use it.
Could you please add an entry to the README explaining your feature and adding an example in there (like with the other features)?

Bersh:
README updated

timgrossmann:
@Bersh Awesome, thank you very much!

While looking at the README I just realised that, even though it might be more misleading than simply using `accounts_limit`, we should rename that variable to `amount`.
Just to keep it consistent.

After that change we'll merge it into `dev`, 🙌

timgrossmann:
@Bersh Awesome, thank you very much for adjusting!

We'll release a new version on Monday that will include your new feature.
Congrats to your first merged PR to InstaPy 🙌 
Looking forward to the next ones 😋 😉

PR #3960, opened 2019-02-10
Enable users interact by the comments of their own profiles.

uluQulu:
Thanks to @TarekJ03 for highlighting this expected behaviour 🐬

### DIFF
--- a/instapy/util.py
+++ b/instapy/util.py
@@ -75,7 +75,7 @@ def is_private_profile(browser, logger, following=True):
 
     # double check with xpath that should work only when we not follwoing a
     # user
-    if is_private is True and not following:
+    if is_private and not following:
         logger.info("Is private account you're not following.")
         body_elem = browser.find_element_by_tag_name('body')
         is_private = body_elem.find_element_by_xpath(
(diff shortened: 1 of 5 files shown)


instapy/util.py
```diff
@@ -75,7 +75,7 @@ def is_private_profile(browser, logger, following=True):
 
     # double check with xpath that should work only when we not follwoing a
     # user
-    if is_private is True and not following:
+    if is_private and not following:
```
timgrossmann, line 78: I think this `is_private` checking should be kept as it was using `is True` :speak_no_evil:

uluQulu: Hey Tim
Please, read my https://github.com/timgrossmann/InstaPy/pull/3637#issuecomment-460226679 post having my final thoughts on comparing bool _values_.

Apart from [based on] that,
1. We should compare multi-type [Boolean] variables using the value equality comparison operator as `if cond == True`.
2. We must compare single-type Boolean variables as `if cond:`.

And in particular, `is_private` is a single-type variable.

timgrossmann:
@uluQulu Thank you for that change 👍 
Good to see that users realise this kind of weird behaviour 😄

PR #3962, opened 2019-02-10
Added Browser abstraction and Decorator which handles Selenium Browser exceptions by reloading the page

### LINKED ISSUE #3581: Browser class abstraction
#### Chrome/Firefox should have a separate class from instapy, and should handle executions/errors and must have a recovery method that WILL NOT fail the instapy flow.

Advantages of this feature:
- Decoupled error handling of web driver problems
- Better recovery capabilities

Possible ways of achieving this while keeping control:
- Creating a abstraction with a Browser class the handles errors and cares about recovery 

> [Bountysource this Issue](https://www.bountysource.com/issues/67497396-browser-class-abstraction)

--- 

###### Feel free to comment with adjustments and improvements

### LINKED ISSUE #3902: Bug fix: Stale element

breuerfelix:
According to the issue (fixes #3902) I want to introduce this sweet little decorator.  
I also added an abstraction for the web driver class as mentioned in (fixes #3581).  

# Browser abstraction

In the browser file there is a new class called custom_browser.  
This is the NEW browser class. Everytime we call a function anywhere in the code we do it via this Class since I changed the self.browser instance to the following class.

I already made a test Function for the function `get_element_by_xpath`. Everytime this function gets called, anywhere in the code, a line gets printed on the screen.

I made this approach since we don't have to change ANY code in any other file. From this time we have the Chromedriver in our hands and we can modify each function only at one place.  

If we need to do some custom error handling for the get element by xpath ? Just change this one function.

# Decorator

## How to use it:

Lets imagine you have a function which needs to click a certain button on the profile page. 
```
def click_and_search_button():
```

But this Button sometimes can't be found due to the stale element exception.
Now you can add the following decorator to it:
```
@retry()
def click_and_search_button():
```

## What does this do ?
* It saves the current browser URL into a variable
* I will try to execute the click and search button function
* if an exception occurs it will browse to the url it saved before (its like reloading)
* it will try to execute the function again
* after 3 times failing it will throw the exception

## Parameters
As you already noticed you have to use this Decorator not as usual.  
That is because it accepts Parameters:
```
@retry(5, 'https://instagram.com/certainpath')
def click_and_search_button():
```

This decorator is now going to try to execute this function 5 times and it won't browse to the url it had at the beginning of the function, it will always browse to the given url.

Both parameters are optional.

## Use Case

I am not that familiar with the InstaPy Source Code that's why I don't know to which functions I can attach this Decorator. I would start of with a function that often throws a stale element Exception.

But thats something I need help with :)

This is just **one** way of handling with stale element errors and maybe we could avoid some of them with this approach.


sionking:
seems like a good idea but I am not very familiar with decorators so far. I need to read about it some more before I can jump in.

breuerfelix:
that's okay @sionking but you should really do that since decorators are really awesome :))

although, keep in mind these are 2 DIFFERENT changes which does not affect each other.  
The custom browser instance can work without the decorator and vice versa. I just wanted to do both tweaks in one file, since they both affect the browser behaviour :)

timgrossmann (approved):

PR #3966, opened 2019-02-11
Fix selectors for comment likes

orlangure:
I was playing around with comment liking feature, and found out it didn't work.

There were two issues:
1. `commenter` calculation was broken. The existing code queried `img` tag, and after my change it is going to query the correct username `a` tag.
2. `button` element query in comment block was broken, it didn't return the correct button. With my change, the code will look for `span` with "Like" label, and use its parent (the actual like button).

This is my first PR here, so I'm aware that I probably miss something. Please guide me to getting this PR approved😼


timgrossmann:
@orlangure Hey, thank you for the fixes and your PR 😊 

Could you reason about why you chose to use css selectors instead of xPath? 🤔 
Both queries can also be done using xPath. Would be interesting to hear your points of view on that.

In general, I think we should stick to one approach and only add another one if we can't reach the goal with that other approach. But maybe there is a good reasoning for choosing css selectors.

orlangure:
@timgrossmann, no particular reason except familiarity. Changed to xpath.

timgrossmann:
@orlangure Good to know 😄 
One more question before merging this (I've double checked all the paths, thank you for your PR)

```python
span_like_elements = comment_line.find_elements_by_xpath(
                     "//span[@aria-label='Like']")
if not span_like_elements:
    # this is most likely a liked comment
    return True, "success"

# like the given comment
span_like = span_like_elements[0]
comment_like_button = span_like.find_element_by_xpath('..')
```

Why do you not use:
```python
span_like_element = comment_line.find_element_by_xpath(
                     "//span[@aria-label='Like']")
if not span_like_element:
    # this is most likely a liked comment
    return True, "success"

# like the given comment
comment_like_button = span_like_element.find_element_by_xpath('..')
```

Seems like we are not using any other index here

orlangure:
@timgrossmann `find_element_by_xpath` seems to raise an exception if no elements match the argument (already liked comment - no like button). So I used a query for multiple elements and tested for an empty list result (nothing found). I might be wrong though, to be honest I didn't read much about selenium sdk.

timgrossmann (approved):

timgrossmann:
@orlangure Thank you very much for this fix and all the comments to my questions 😊 

Congrats to your first PR merged into InstaPy 🙌 
Looking forward to your future ones 😋

orlangure:
@timgrossmann, thanks for a warm welcome😼

PR #4004, opened 2019-02-18
Temporarily turn off follow for like_by_tags interaction

traincodesleep:
By submitting this pull request I acknowledge that I have read and agree to the code of conduct for this project.

This Pull Request addresses the following issue:
When `like_by_tags` is used with `set_do_follow` enabled, the user is followed in the `like_by_tags` method as well as in the `like_by_users` method which is called by `like_by_tags` to interact with the user.

This also causes the `set_do_follow` percentage setting to be actually double what is specified since the chance that the user gets followed is calculated twice once in `like_by_tags` and again in `like_by_users`

This PR fixes the issue by implementing a snippet of code already available in `instapy.py` to temporarily turn off the following for the interaction and then reverting the setting after the interaction.

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1024,9 +1024,10 @@ class InstaPy:
                                       self.user_interact_percentage
                         # Do interactions if any
                         if do_interact and self.user_interact_amount > 0:
-                            original_do_follow = self.do_follow  # store the
+                            original_do_follow = self.do_follow
+                            # store the
                             # original value of `self.do_follow`
-                            self.do_follow = False 
+                            self.do_follow = False
                             # disable following
                             # temporarily cos the user is already followed
                             # above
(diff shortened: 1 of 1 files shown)


timgrossmann:
@emmanuelclarke Thank you very much for this fix. Good catch btw!
Could you change the way of adding comments to the (hopefully) general way it is done in the codebase, please?

```python
# comment
code block
```

Instead of 
```python
code block # comment
# comment
```

Thank you very much.
I'll add an entry to the changelog file and merge it once we've adjusted this small change.
I'll review your other PR soon.

Thank you very much and congrats to your first PR to InstaPy 🙌 

P.S. I've also changed the merge destination to our `dev` branch instead of `master`

traincodesleep:
@timgrossmann I made the necessary changes for the block that I added as well as some other places where I copied the code from that had the same comment formatting.

timgrossmann:
@emmanuelclarke Good to know that, thank you very much for letting me know. Seems like I have to take a closer look at that from now on 🤔 

Thank you very much for this fix and congrats to your first merged PR to InstaPy 😊 🙌

PR #4015, opened 2019-02-19
Update README.md

timgrossmann:


timgrossmann:
@mwargan What do you think about that README update 😉 ?

What is your opinion about the documentation part?
Where and how do we add it best to make it best seen by the user?

mwargan:
@timgrossmann Looks cool! I like how Stripe does it:https://github.com/stripe/stripe-php <-- I think its a good example. Putting it under a header `Documentation` in the readme should be enough to get the user to stop scrolling and see the finer link under said header.

timgrossmann:
God... this is a mess... So sorry 🙁

PR #4035, opened 2019-02-20
xpath update for Log In button

converge:
there is a new Log In button location. Should we add it on Changelog?


CharlesCCC:
Just about PR this and saw your change 👍

brocool:
#4036 solves it for you. Check it out the solution is simple and easy

timgrossmann (approved):

timgrossmann:
Love this community 😄 🙌 
Sorry for the delay...

@converge Please always add an entry to the changelog. Bug fixes like this should be published with a new "fix" version, so only last digit increased by one 👍

PR #4042, opened 2019-02-21
InfluxDB Integration - Test required

StanGirard:
# InfluxDB

**Disclaimer** : Still needs to be cleaned up. 

## **Testing still required**

This allows to log all follows, unfollows, comments, likes stats in an Influx DB for better analytics

### Installation

- Install [InfluxDB](https://docs.influxdata.com/influxdb/v1.7/introduction/installation/)
- Launch InfluxDB
- Create a user
- Create a DB
- Launch the session with the right arguments for your  InfluxDB database

```Python
session = InstaPy(username='', password='', headless_browser=False, disable_image_load=False, user_influx='root',
                 password_influx='root',
                 db_influx='test',
                 host_influx='localhost',
                 port_influx=8086)
```

### DIFF
--- a/instapy/comment_util.py
+++ b/instapy/comment_util.py
@@ -101,12 +101,13 @@ def comment_image(browser, username, comments, blacklist, logger, logfolder, cli
     if (client_influxDB is not None):
                 json_body = [
                 {
-                    "measurement": "Commented",
+                    "measurement": "Comment",
                     "tags": {
-                        "username": username,
+                        
                     },
                     "fields": {
-                        "comment": rand_comment,   
+                        "comment": rand_comment,
+                        "username": username   
                     }
                 }]
                 client_influxDB.write_points(json_body)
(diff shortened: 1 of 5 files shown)


sionking:
?

CharlesCCC:
Interesting, 
What is pros of InfluxDB over SQLite ? (besides analytics, i'm assuming you can also do analysis using SQLite)

StanGirard:
> Interesting,
> What is pros of InfluxDB over SQLite ? (besides analytics, i'm assuming you can also do analysis using SQLite)

It is a time series DB. It is made for advanced analytics. It is very fast and can handle hundreds of insert per second. It is mainly for the dashboard for the gui. SQL is still here unchanged 

I could go into details. But to keep it short it is a db made for analytics.

I got that using InfluxDB and Grafana.
capture d ecran 2019-02-21 a 20 27 08

PR #4059, opened 2019-02-23
Prevent division by zero in validate_username

orlangure:
TBH I'm not sure that this have ever caused issues, but I think it might be related to the bot sometimes following users that do not match `relationship_ratio`. At most (or all) such cases the bot followed accounts with zero followers/following numbers.

Anyway, I think it is better to defend against division by zero.

Oh, and I also removed some trailing white space.


timgrossmann:
@orlangure Good catch, thank you very much! 👍 

Congrats to your first merged PR (I've changed the destination to `dev` for now 😉 ) 🙌

PR #4064, opened 2019-02-24
changed connection method

StanGirard:
# Instapy

Changed the method for setting up.
The password is hardcoded in the settings.

### DIFF
--- a/instapy/settings.py
+++ b/instapy/settings.py
@@ -75,11 +75,11 @@ class Settings:
 
 
     #inluxdb Settings
-    host_influx = None
-    port_influx = None
-    user_influx = None
-    password_influx = None
-    db_influx  = None
+    host_influx = "localhost"
+    port_influx = 8086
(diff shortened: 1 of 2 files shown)


instapy/settings.py
```diff
@@ -75,11 +75,11 @@ class Settings:
 
 
     #inluxdb Settings
-    host_influx = None
-    port_influx = None
-    user_influx = None
-    password_influx = None
-    db_influx  = None
+    host_influx = "localhost"
```
breuerfelix, line 78: as already mentioned in discord, i would not make this static ! always pass this as param in instapy constructor, or pass an 'options' objects in the instapy constructor :)

breuerfelix (changes requested):

instapy/settings.py
breuerfelix: well we can just close this PR since i would leave it as it was before :) liked that !

Issue #4065, opened 2019-02-25
Use the InstaPy code instead of the installed by pip - Development purposes


joanroig:
## Current Behavior
I've cloned InstaPy and installed its dependencies, after that I've runned the quickstart.py file and I've noticed it compiled and runned the code from the /instapy folder.
If I run code from another folder, like /instapy/custom/quickstart.py the code is runned from the pip dependency, so I cannot test my changes on the instapy core.
## Expected Behavior
There should be an easy way to point the InstaPy core for making changes on it.
## Possible Solution (optional)
I've tried setting the workspace folder but it seems that only points the logs folder, the core runs from the pip dependency.
## InstaPy configuration
Regular clone with basic configurations.

timgrossmann:
@joanroig Awesome to hear you're planning to do some development on InstaPy 🙌 

I think what you are searching for is this:
```
pip install --help
...
  -e, --editable <path/url>   Install a project in editable mode (i.e. setuptools
                              "develop mode") from a local project path or a VCS url.
```

You might need to either deinstall instapy using `pip uninstall instapy` or use the `-I` flag to ignore the installed version when doing this to avoid `already satisfied`

Let me know if that helps 😊

breuerfelix:
ahahah nooo this is waaaay easier ! and @uluQulu already wrote that in the readme :)

clone the repo and cd into the repo.  
execute `pip3 install -e .`
it will install the current folder as a package but it will refer to this package everytime you use it anywhere on your computer :)

so you can make local changes, start the quickstart anywhere on your systems, and it will use your local changes. you do not have to update something, its always 'live'.

/edit actualy readme entry http://prntscr.com/mppuf4

joanroig:
Thanks! I've skipped the -e part in the docu

PR #4066, opened 2019-02-25
Solution for the 'Log in' text error due an A/B test

joanroig:
I've seen the login button text different in some occasions during past hour, so Instagram is probably doing an A/B test that can affect new logins without a previous cookie. The solution shall be as follow:
- Detect an alternative 'Log In' button if the regular 'Log in' is not found.

Needs to be tested and probably needs to be in a try catch statement, I currently cannot run my own code due my folder structure explained here https://github.com/timgrossmann/InstaPy/issues/4065

### DIFF
--- a/instapy/login_util.py
+++ b/instapy/login_util.py
@@ -202,11 +202,11 @@ def login_user(browser,
     # Check if the first div is 'Create an Account' or 'Log In'
     try:
         login_elem = browser.find_element_by_xpath(
-            "//article//a[text()='Log in']")
+            "//a[text()='Log in']")
     except NoSuchElementException:
         print("Login A/B test detected! Trying another string...")
         login_elem = browser.find_element_by_xpath(
-            "//article//a[text()='Log In']")
+            "//a[text()='Log In']")
 
     if login_elem is not None:
         try:
@@ -265,11 +265,11 @@ def login_user(browser,
 
     try:
         login_button = browser.find_element_by_xpath(
-            "//button//div[text()='Log in']")
+            "//div[text()='Log in']")
     except NoSuchElementException:
         print("Login A/B test detected! Trying another string...")
-        login_elem = browser.find_element_by_xpath(
-            "//article//a[text()='Log In']")
+        login_button = browser.find_element_by_xpath(
+            "//div[text()='Log In']")
 
     (ActionChains(browser)
      .move_to_element(login_button)
--- a/instapy/login_util.py
+++ b/instapy/login_util.py
@@ -203,7 +203,7 @@ def login_user(browser,
     try:
         login_elem = browser.find_element_by_xpath(
             "//article//a[text()='Log in']")
-    except:
+    except NoSuchElementException:
         print("Login A/B test detected! Trying another string...")
         login_elem = browser.find_element_by_xpath(
             "//article//a[text()='Log In']")
@@ -266,7 +266,7 @@ def login_user(browser,
     try:
         login_button = browser.find_element_by_xpath(
             "//button//div[text()='Log in']")
-    except:
+    except NoSuchElementException:
         print("Login A/B test detected! Trying another string...")
         login_elem = browser.find_element_by_xpath(
             "//article//a[text()='Log In']")
--- a/instapy/login_util.py
+++ b/instapy/login_util.py
@@ -200,10 +200,11 @@ def login_user(browser,
               "new cookie...".format(username))
 
     # Check if the first div is 'Create an Account' or 'Log In'
-    login_elem = browser.find_element_by_xpath(
-        "//article//a[text()='Log in']")
-
-    if login_elem is None:
+    try:
+        login_elem = browser.find_element_by_xpath(
+            "//article//a[text()='Log in']")
+    except:
+        print("Login A/B test detected! Trying another string...")
         login_elem = browser.find_element_by_xpath(
             "//article//a[text()='Log In']")
 
@@ -262,10 +263,11 @@ def login_user(browser,
     for i in range(2):
         update_activity()
 
-    login_button = browser.find_element_by_xpath(
-        "//button//div[text()='Log in']")
-
-    if login_button is None:
+    try:
+        login_button = browser.find_element_by_xpath(
+            "//button//div[text()='Log in']")
+    except:
+        print("Login A/B test detected! Trying another string...")
         login_elem = browser.find_element_by_xpath(
             "//article//a[text()='Log In']")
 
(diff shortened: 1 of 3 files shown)


joanroig:
Needs to be tested again, it appears only in some logins, and I cannot run my own code due https://github.com/timgrossmann/InstaPy/issues/4065

timgrossmann:
@joanroig Thank you very much for proposing this fix!
I've added a comment to your issue, let me know if that works for you and you can test your code 😊

breuerfelix:
@joanroig i also added a new comment to your issue ! to even better test your code :))

joanroig:
Thanks @timgrossmann and @breuerfelix, now I can test the code properly 👍 
I've forced the A/B test and I confirm it is working while creating the needed cookie, so it can be merged.

joanroig:
Thanks @timgrossmann and @breuerfelix, now I can test the code properly 👍
I've forced the A/B test and I confirm it is working when it creates the needed cookie and when it needs to be refreshed, so it can be merged.

bylezz:
Hello,

Can you explain a little more in detail how you fixed it? I get this:
     login_elem = browser.find_element_by_xpath (
NameError: name 'browser' is not defined

Thank you :)

joanroig:
> Hello,
> 
> Can you explain a little more in detail how you fixed it? I get this:
>      login_elem = browser.find_element_by_xpath (
> NameError: name 'browser' is not defined
> 
> Thank you :)

Yes, I´ve just catched the error throwed when the find_element_by_xpath doesn't find the expected element, and then I've launched the same code but for another string.
You need to get the changes and put them inside instapy/login_util.py

instapy/login_util.py
```diff
@@ -258,8 +263,13 @@ def login_user(browser,
     for i in range(2):
         update_activity()
 
-    login_button = browser.find_element_by_xpath(
-        "//button//div[text()='Log in']")
+    try:
+        login_button = browser.find_element_by_xpath(
+            "//button//div[text()='Log in']")
```
converge: ```//div``` and ```//button``` are redundant, the correct is:
```login_button = browser.find_element_by_xpath("//button/div[text()='Log in']")```

Comment (timgrossmann) [reply instapy/login_util.py]: Isn't that were we came from?
We merged one that introduces the `//` in order to avoid problems with the Log in div that is somewhere inside the button element.

File: instapy/login_util.py [hunk 2]
```diff
@@ -200,8 +200,13 @@ def login_user(browser,
...
     # Check if the first div is 'Create an Account' or 'Log In'
-    login_elem = browser.find_element_by_xpath(
-        "//article//a[text()='Log in']")
+    try:
+        login_elem = browser.find_element_by_xpath(
+            "//article//a[text()='Log in']")
+    except NoSuchElementException:
+        print("Login A/B test detected! Trying another string...")
+        login_elem = browser.find_element_by_xpath(
+            "//article//a[text()='Log In']")
```
CharlesCCC, line 210: I will also suggest add a 
```python 
finally:
    return False
```
to end the session gently with a error message if Instagram decide to change its element again.

converge: return True/False is a great idea, in this way we can start using Pytest for unit test.

instapy/login_util.py
converge: Maybe I made a mistake with the PR to fix the login button xpath, ```//``` will set the element as the root element for the next elements, two ```//``` will set the root element twice.

instapy/login_util.py
joanroig: Should I add it with another commit or you can add it as a reviewer? @CharlesCCC

timgrossmann: Well, unfortunately this does not work as expected...
When looking at this example:
```python
def test():
  try:
    print('test')
    return True
  finally:
    print('finally')
    return False

res=test()
print(res)
```

It outputs:
```
test
finally
False
```

So no matter what happens in the try or except block, the return value will always be `False`🙁

joanroig: https://stackoverflow.com/questions/19805654/python-try-finally-block-returns
Finally will be always executed, so I propose to return after the second chance if the result is empty again, but it will need another try catch block or so.

We can also prepare a list of strings to search, like "Login", "Log in", "Log In", etc, and encapsulate the string checker in a method; if the method succeeds it will return true and the execution can continue, otherwise will return false. I think it will help to have this code encapsulated if the project will include unit tests.

instapy/login_util.py
```diff
@@ -258,8 +263,13 @@ def login_user(browser,
...
         update_activity()
 
-    login_button = browser.find_element_by_xpath(
-        "//button//div[text()='Log in']")
+    try:
+        login_button = browser.find_element_by_xpath(
+            "//button//div[text()='Log in']")
+    except NoSuchElementException:
+        print("Login A/B test detected! Trying another string...")
+        login_elem = browser.find_element_by_xpath(
```
dimeloper: I think this should be `login_button`.

instapy/login_util.py
breuerfelix: @converge is right here, if we set `//` infront of the div, than `//button` got not effect at all :)

joanroig:
@timgrossmann I think this can be merged and then if anyone wants to improve the code then create another pull request, this one was made as a quick bugfix, so I think it should be available as soon as possible for everyone

timgrossmann:
@joanroig Absolutely! 
I'm just not sure if this actually fixes the the problem since as @converge and @breuerfelix pointed out, the `//` in front of both elements shouldn't fix it 🤔

breuerfelix:
> @joanroig Absolutely!
> I'm just not sure if this actually fixes the the problem since as @converge and @breuerfelix pointed out, the `//` in front of both elements shouldn't fix it 🤔

@timgrossmann well we can just change the second // to / but if fixes it since this times it tries to search for `Log in` and afterwards for `Log In` ... and this is something that we didnt do before :) so we got a fallback search implemented here.  
we can also just remove the `//button` from the search which would just look a little cleaner and will be a bit faster i guess, since it wont change the output

timgrossmann:
@joanroig Could you apply the changes mentioned by @breuerfelix, please? 😊 
After that we can merge this 👍

joanroig:
@timgrossmann @converge @breuerfelix I've retested all again, the last commit worked fine in all cases 😊

breuerfelix (approved):
simple and straight forward ! thank you @joanroig :)

timgrossmann:
@joanroig Awesome, thank you very much for the change 👍 
I'll merge and release this once I'm home.

timgrossmann (approved):

PR #4083, opened 2019-02-28
Expose `threaded_session`

ohduran:
Expose the `threaded_session` parameter on `close_browser()` when calling  `Instapy.end()` method so that asynchronous tasks (Celery) work fine by setting that parameter to `True`.


timgrossmann:
@ohduran Good one! 👍 
Thank you very much and congrats to your first merged PR with InstaPy 🙌

Issue #4096, opened 2019-03-03
Docker - Not writing/creating logs files and cookie


razki:
## Expected Behavior
I would expect when using the docker_quickstart.py and docker-compose.yml to see some log files written including a cookie.

## Current Behavior
Bot works fine, however it needs to create a cookie every time and there is no logs directory created

Herrox:
I may be wrong, but it looks like files (logs, cookie, blacklist...) are saved under `~/InstaPy/logs` instead of `/code/InstaPy/logs`

I will have a look on how to fix this.

Herrox:
@razki I just created the PR #4112 to fix this problem.
Waiting for the dev team to review the change and commiting it

PR #4112, opened 2019-03-05
Docker fix workspace issue + delete command parameters

Herrox:
Fix workspace variable as `/code` instead of `/root` in `docker-compose.yml` file: it was causing an issue (#4096) where files generated by `InstaPy` where written in a non-permanent storage on `/root` and erased each times the container was started again. Including `cookie`, `blacklist`, `general.log` and all the others files in `InstaPy/logs/_username_/`

Delete the `command` parameter in `docker-compose.yml` because this paramter is already specified directly in `Dockerfile`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,6 @@
-version: '3' 
+version: '3'
 services:
   web:
-    #command: ["./code/wait-for-selenium.sh", "http://selenium:4444/wd/hub", "--", "python", "code/docker_quickstart.py"]
     environment:
       - PYTHONUNBUFFERED=0
       - INSTAPY_WORKSPACE=/code
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,9 +1,10 @@
-version: '3'
+version: '3' 
 services:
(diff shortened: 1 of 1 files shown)


Herrox:
By the way, do you have guys any discord channel or another way to communicate about the new Docker project? It will be my pleasure to be handy on the matter, I already saw some improvements to the `Dockerfile` on `instapy-docker` repo.

Thanks.

timgrossmann:
@Herrox Thank you very much for this PR, yes we are moving the docker config out of the core repository into the instapy-docker repo.
This will happen as soon as the next minor release 0.3.0 will be released (probably Monday)

The discord server in general is linked in the readme, we don't have a dedicated docker channel though.
Maybe it makes sense to have though.

I'm sorry but we won't merge this PR into core anymore. 
I'll create a new PR that removes every file that is necessary for docker from core.

I'd love to see you work on the instapy-docker repo, though if you're interested.
I'll create a public channel on Discord so we can write there :+1: 

Thank you very much

Edit: I've added it, would love to see you there.

PR #4158, opened 2019-03-14
Add additional exception cath

CharlesCCC:
Currently the Instagram is having issue, my account is keep trying to login and can't find the "Login" element, instead of blow up with NoSuchElementException, add additional try catch along with return False to exit it gracefully. 

Also add gitignore to ignore all the *.db SQLite files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -98,7 +98,7 @@ ENV/
 .DS_Store
 
 # Storage
-instapy.db
+*.db
 logs/
 
 # Editors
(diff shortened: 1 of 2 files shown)


timgrossmann:
@CharlesCCC Does that happen all the time? (master or dev) 😊

CharlesCCC:
@timgrossmann master, It was happened quite often yesterday(which could be related to the Instagram/Facebook downtime). I had to add that logic just to keep my scheduled run going. otherwise, I had to go back and re-kick off the schedule every time when it blows up.

timgrossmann (approved):

PR #4170, opened 2019-03-15
Fix: (re) adding page_reload, after cookie load, on login_user()

bugre:
One of the last releases added a change to Language change on login page. Did also remove a reload_webpage() call in login_user() function, that is/was needed to correctly load the IG page after cookie had being added to the browser session.

This adds the import and reload again.


bugre:
Did force push the file, to remove a trailing whitespace on line 173, that got past :)

timgrossmann (approved):

PR #4173, opened 2019-03-15
Add sleep to wait for follower list to load

anna-sunberg:
Fixed "Failed to load desired amount of users" when trying to read long follower lists (users with hunders of thousands followers) that takes too long to load


timgrossmann (approved):

timgrossmann:
@anna-osipova Thank you very much for the addition, 5 secs are quite long but maybe this will help resolve some problems 👍

Issue #4039, opened 2019-02-21
Feature Request - Interact with likers (Without Follow)


s-h-a-d-o-w-pixel:
Hi,
I use the follow_likers and I was wondering why we don't have a like_likers or something.
I'm trying to give back likes to my own likers and that feature would be such a time saver.

ThatDudeIsReal:
I was about to post a Feature Request as well. it would make so much sense to interact with active users that like posts no matter if users like your post or users that liked a post from other accounts. these users

ThatDudeIsReal:
@timgrossmann is it possible to use "session.follow_likers" without following somehow similar to "session.set_do_follow(enabled=False)"

DmytroDiachuk:
Good idea

Issue #4013, opened 2019-02-19
Refactor like_by, follow_by, comment_by => interact_by


timgrossmann:
#### In order to make InstaPy easier to use we want to cut down on the amount of ways things can be done. One of the ways to start making it simpler is to remove the separate methods `like_by_...`, `comment_by_...` and  `follow_by_...` and introducing a common "interface" as `interact_by_...`.

Advantages of this feature:
- Clean up code base
- Less places to fix problems if elements change
- Deduplication of code 
- Simpler documentation
- Higher level of abstraction
- Unified interface for users (always `interact_by` followed by e.g. `tags`, `users` etc.) and then controlling interactions by using `set_do_like`, `set_do_comment` and `set_do_follow`

Possible ways of achieving:
- Use the already existing methods as a "base line" on how this can be achieved 

> [Back this Issue](https://opencollective.com/instapy)

--- 

###### Feel free to comment with adjustments and improvements

tE3m:
So would `set_do_follow` or other settings be global or only apply to a certain function? I think it'd be interesting to make settings callable from inside of one of those `interact_by` functions. What do you think?

timgrossmann:
@TarekJ03 Sorry I think I don't exactly understand what you're suggesting.

The idea, for now was to have:
- `interact_by_tags`
- `interact_by_feed`
- `interact_by_followers`
- etc.

with the "same" functionality we have at the moment, just combined into one method.
To avoid confusions like "what is the difference between follow_users_followers and interact_user_followers with do_follow".

timgrossmann:
@TarekJ03 do you mean having the follow like and comment percentage for each method separately instead of global? 🤔

tE3m:
@timgrossmann yes, exactly. Setting `set_do_follow`, `set_do_like` and `set_do_comment` for each method seperately.

timgrossmann:
@TarekJ03 Interesting approach 🤔 
It might make sense to do that, I think we have to consider that after analysing how people use InstaPy once the GUI is available.

PR #4177, opened 2019-03-16
New feature: like likers

orlangure:
This PR adds a new feature: `like_likers`.

We currently support `follow_likers`, which seems not enough as sometimes we might wish to only like our likers and not follow them (or maybe we already follow them). The code is very similar to `follow_likers`.

Related/might close https://github.com/timgrossmann/InstaPy/issues/4039


timgrossmann:
@orlangure Thank you very much for this feature!

I'm still very much into having everything as `interact_...` do you think you could refactor this to be a more generic version (so we could use `set_do_like`, `set_do_follow`, `set_do_comment`)?

If not, we maybe should extend the documentation a little bit.
It's not exactly clear to me what all the parameters do.

What is the example of 'not-a-user'?

```python
session.like_likers(['user', 'not-a-user'], photos_grab_amount=3,
                     like_likers_per_photo=3, likes_per_liker=3)
```

Thanks :v:

orlangure:
@timgrossmann, `not-a-user` is just a very bad choice of a username to put after `user` 😹

I'll look into refactoring and generalizing it eventually, not sure when I'll have time. If there is no time pressure, I'll continue using my own fork (I use this feature now) and keep this PR open indefinitely. Unless someone wants to step in and move things around.

timgrossmann:
@orlangure Ahh ok 😂 

Well I think we can merge this for now (after elaborating a bit more in the README and changing the example users 😛)
And then get back to it once we start having a bigger refactoring.

@orlangure Could you do the additions in the README, or should I do them? 😊

HCWcoder:
@orlangure thanks for this feature

orlangure:
@timgrossmann, on it.
@HCWcoder, glad you find it useful😼

orlangure:
I changed `like_likers` docs, let me know if there is something else that needs fixing before the great refactor.

dmsfabiano:
I did a very quick check, but from when I saw there are a couple things you should check. This could have way better functionality @timgrossmann 

1 - New function does not do checks for videos that don't display likes. What will the behavior be?
2 - Where is the check to see if user has likes activated and with what percentage?
3 - Why only `like_likers` and not `interact_likers` the fucntion is already implemented here so most of the work is done; is fairly easy to extend it using API calls to complete interacts (i.e. Like, commenting, following)

New features should be as complete as possible and using the tools we already have (API functions). Otherwise we will end up with a lot of single purpose functions that could be doing a lot more. 

Overall, great addition, I've had something like this in my toolset for a while, but let's make the official version as complete as possible  👍

ishandutta2007:
@timgrossmann 
I have been meaning to stir this discussion , maybe this is the right thread, we have way too many APIs to do the same or similar things. This kind of design is making it a hell to maintain the code.

How I see is there should be one and only one api which does this three steps:
1. Choose target user(by _tag_, or by _location_, or _by hashtag or from Followers/Following of a user_, from _a specific list_)
2. Choose the desired item of that user(_post_ or _profile_ or _comment_)
3. Interact with that item(_follow_ or _like_ or _comment_)

If we see these three steps separaly we may enhance either of one of these three steps iteratively than trying to create new workflows which are essentially new combination of a usecase of these three steps.
You can expose these three APIs separately or wrap with one. To start we can create a universal wrapper and refactor incrementally later as and when we get opportunity.

timgrossmann:
@dmsfabiano Thank you for the review, you've got very good points there. We should consider implementing the mentioned changes @orlangure :blush: 

@ishandutta2007 Well, absolutely. There is an issue open since Feb 19 (https://github.com/timgrossmann/InstaPy/issues/4013) and I already mentioned it several times on the discord channel. Would even be able to provide bounties for the refactoring through OpenCollective. The problem right now only is that I don't have the time to do refactors as big as those ones in my free time.

I'd love to see this happening but well yeah, we can't pay someone for a month to do that. Maybe at some point in the future.
For now having all of them separate with all the logic in place might motivate someone to merge together the logic and still PR an `interact_by` feature.

But I absolutely agree with you!

imjustin:
Has this been merged to the main branch? if I'd like to try this feature, what's the best way to go about that? (sorry i'm not super familiar with git...)

orlangure:
@imjustin not yet. You can do the following with git cli:
```
cd <instapy-dir>
git checkout -b like-likers
git pull origin pull/4177/head
```
This will move you to a new branch from master (latest version) and get the contents of this PR from github. You'll then be able  to use this feature, but only while being on that new branch. If you'd like to use the feature on `master`, just don't `git checkout`.

imjustin:
how does that work if i've installed instapy through the pip3 installer rather than cloning the repo?

orlangure:
I think you can install packages with `editable` option and then modify the files directly.
Or you could clone the repo to some other directory and run it from there a couple of times with this feature branch.

imjustin:
I follow what you're saying in theory, but i'm not sure i'm doing it right in practice haha. so what i've done is cloned your repo with the right interacty_by_likers branch into a separate folder, but when I put my new quickstart.py into that folder and run it from there, i think it's still using the existing installation of instapy. how do i force it to use the local instapy files rather than the system-installed version i previously had?

orlangure:
@imjustin Try looking into virtual environments in python, they can allow to isolate the scripts you are running from other scripts/projects that exist in your environment.

timgrossmann:
@orlangure Sorry for not merging this but I'm sure we can integrate this once we start to refactor our interaction methods into an `interact` based system.
Sorry 😞

stephane-archer:
Any news about this feature? @timgrossmann @orlangure

PR #4218, opened 2019-03-27
Added KEYS.ENTER to submit login data

CharlesCCC:
Currently running into issue when login into account. looks like instagram is changed their xpath again for login button.


timgrossmann:
@CharlesCCC Awesome, thank you very much for this update! 
Could you please make sure that the CI passes? :blush:

CharlesCCC:
@timgrossmann  I would love to, but TBH, I have no idea how to fix that, As I view the log,it failed to call ` tox` command ??? Is there anyway I can kick off the build again ?

instapy/login_util.py
```diff
@@ -259,9 +259,23 @@ def login_user(browser,
         login_button = browser.find_element_by_xpath(
             "//div[text()='Log in']")
```
converge: I believe the "Log in" way doesn't exist, am I wrong?

I'm just able to find "Log In" button.

.gitignore
```diff
@@ -104,6 +104,7 @@ logs/
 # Editors
 .idea
 .vscode
+.bak
```
converge, line 107: In this for Mysql Workbench backup file?

converge (changes requested):

instapy/login_util.py
CharlesCCC: You probably right, I was accumulated all the possible option from the previous changes.

.gitignore
CharlesCCC: actually, it was from Notepad++ file.  but it could be others as well.

hugio55: I am pretty new to this whole world. Where is this .gitignore file?

timgrossmann:
@CharlesCCC @converge which changes do you exactly want for this PR?
I think it's good to cover "every" possible combination here.

Please define so we can merge this 😊

CharlesCCC:
@timgrossmann I would say keep all the possible. There should be nothing to hurt. worst case is it take couple more second to login the user, but (it just one second every 3 months, since the cookie are cached).

converge:
It's all good for me, I just don't know how to deal with the "Log In" vs "Log in". It's hard to test it because it just catches "Log In" on my side, and I don't know if people catch "Log in".

mkhattar:
I changed "Log in" to "Log In", seems the text on the button changed. Please can someone else test it to confirm?

CharlesCCC:
@mkhattar @converge I have replaced "log in" with "Log In".

converge (approved):

orhan-drsn:
What about:
`from selenium.webdriver.common.keys import Keys`

`(ActionChains(browser)`
`        .move_to_element(input_password[0])`
        `.click()`
        `.send_keys(Keys.ENTER)`
        `.perform())`

Maybe this is the best way to don't have problems in future

CharlesCCC:
@orhan-drsn Seems a good idea, Have you tested it ? if yes, I will make the update to implement that . thank you

orhan-drsn:
@CharlesCCC yes it works for me.
It's like a real human would do it. After entering the passwort to push enter

CharlesCCC:
@orhan-drsn Thank you for confirming that, I made the change and also tested it locally, seems working really well.

converge:
@orhan-drsn great idea!

timgrossmann:
@CharlesCCC @orhan-drsn Thank you very much for this, guys! This will remove a lot of problems in the future 🙌

PR #4238, opened 2019-04-01
Added logging to telegram chat

da1z:


da1z:
should I fix it to work with 2.7?

timgrossmann:
@da1z Thank you very much for this! One thing we have to discuss is whether this actually should be part of the core repository.
There are already two repositories that work as extensions that can also enable logging to Telegram.

I think we should keep this separate from core, especially after @converge proposal to the core team.
Maybe we should publicly share the plan to lighten core with the community and create a project page and issues from it. What do you guys think? @converge @breuerfelix @CharlesCCC @w4ngel

breuerfelix:
i think we should add it to the core repo later on ! my idea:

one folder called core, one folder called extensions.
in the folder extensions every extension like this telegram api gets its own folder and it can be programmed in there.

But for this to happen we need to refactor some things in the core repo, hope this will happen soon :)

@da1z hoping to get this merged soon in the future because i really like the idea :)

timgrossmann:
@breuerfelix @da1z Feel free to work together to create the first step in building this distinction and give a guideline implementation of the future workings for extensions 🙏

PR #4239, opened 2019-04-01
Posibility to skip non bussiness accounts

da1z:


timgrossmann:
@da1z Awesome and simple addition! Thank you very much :blush: 
Could you please also add a short description into the README which describes this feature? :blush:

da1z:
@timgrossmann uodated readme

PR #4256, opened 2019-04-03
Qulu's proxy class

timgrossmann:
This feature was the last one for @uluQulu unfortunately.
Since he was not able to PR it himself, I'm now trying to take what he did and integrate it into our latest version of InstaPy...

> Documentation to be added


CharlesCCC:
uhhh, what's this different than the original "proxy" that was being used ? The change seems massive with multiple .py file been totally treat a overwritten(hard to tell what exactly was changed or how it should be used).

timgrossmann:
@CharlesCCC Yeah I fixed that now 😄 
Basically there is only one major addition. The proxy class.
This separates the usage of proxy from the actual InstaPy file.

In general, the user can now get a Proxy object and set it, then pass this proxy object to InstaPy.

timgrossmann:
One thing that was not implemented though is the authentication with credentials... 
We should discuss if this is worth merging and check back with @converge if he could adjust the proxy file to incorporate the authentication feature

breuerfelix:
10/10 merge it ! everything that's going to reduce our main file and abstract it to somewhere else is worth gold :)

timgrossmann:
@breuerfelix  haha, yeah this still needs more work tho.
This was created before @converge fixed the proxy auth for chrome.
So this would need to be integrated as well, together with testing for chrome and FF to make sure all 4, chrome proxy with/out auth and FF proxy with/out proxy works :smile: 

So might take a little bit more time :sweat_smile:

eracle:
I implemented a version that pass the chrome extension as a in-memory file:

https://github.com/timgrossmann/InstaPy/pull/4256/files#diff-6f081321fe2156f810826aa9de443c87R83

````````python
    pluginfile = io.BytesIO()

    with zipfile.ZipFile(pluginfile, 'w') as zp:
        zp.writestr("manifest.json", manifest_json)
        zp.writestr("background.js", background_js)

    return pluginfile.getvalue()

````````

https://github.com/timgrossmann/InstaPy/pull/4256/files#diff-6f081321fe2156f810826aa9de443c87R114

````````python
# chrome_proxy_extension from the previous code       

chrome_options.add_encoded_extension(base64.b64encode(chrome_proxy_extension).decode('ascii'))

````````

By the way is important also to include it on the set_selenium_remote_session function.

The idea behind it is to not have any flying files on the codebase.
If you like the idea I can prepare a PR.

breuerfelix:
is this an addition to this pr ? or a completely new PR with other changes ?

would be cool if we could merge this into this PR to have the things going together so we can merge it into master :)

eracle:
@breuerfelix I agree with you. Is an addition to this PR
Waiting for a thumb up regarding the idea of using an in-memory file for storing the chrome extension.

breuerfelix:
what are other choices ?

if the only way is to have a real file or to have a memory file... lets do the memory stuff for sure !

but right now, I don't really know why we need this at all :D I mean what does this in memory file exactly do ? :)

timgrossmann:
@eracle I like the idea 👍 

Would love to see an even improved version of the Proxy Class abstraction!
Please create a new PR for this change 😊

Issue #4279, opened 2019-04-07
InstaPy 0.7.0


breuerfelix:
# Ideas / Refactoring

I stumbled across multiple really good ideas for big refactoring.  
Before really doing this, we should have a place to discuss the road to 5.0.

You wanna discuss about an idea written here ? Quote the part and answer with your own thoughts ! Quoting is important since this post might change really often, so everybody still knows what the state on your comment was :)

## API ideas by @ishandutta2007 

I have been meaning to stir this discussion , maybe this is the right thread, we have way too many APIs to do the same or similar things. This kind of design is making it a hell to maintain the code.

How I see is there should be one and only one api which does this three steps:

Choose target user(by tag, or by location, or by hashtag or from Followers/Following of a user, from a specific list). 
Choose the desired item of that user(post or profile or comment)
Interact with that item(follow or like or comment). 
If we see these three steps separaly we may enhance either of one of these three steps iteratively than trying to create new workflows which are essentially new combination of a usecase of these three steps.  
You can expose these three APIs separately or wrap with one. To start we can create a universal wrapper and refactor incrementally later as and when we get opportunity.

## HTTP API by @converge 

would be awesome if you could wrap up your thoughts here

## Extension Management by @breuerfelix 

InstaPy package gets an additional folder named `extensions`. Every Extension (like Clarifai in this example) gets its own folder in this directory.  
Extensions will be programmed only in their folder. This will keep the main codebase clean.

__Interaction of Extensions with the core:__  
We create on Singleton called Event for example.  Lets assume Clarifai want to prevent the core loop from liking / commenting some pictures.  
There will be a function called `event.before_interaction(image)`. The core function will call this function before liking / commenting a picture and if it returns `true` it will like it and it will skip it if it returns `false`.  
The clarifai extension can then add its validator function for this event. `event.add_listener('before_interaction', self.validate_picture)`. `self.validate_picture` will be a function which checks wether the image is inappropriate or not. (returning true or false).  
`event.before_interaction(image)` is now calling ALL functions added as a listener to its own event and if any of those is returning false, it will return false. Only if all callback functions are returning true, the image will be liked.

The users can now just import their clarifai extensions in the main.py `from instapy.extensions import clarifai` and add these to the main instapy object `instapy.add_extension(clarifai)`.  
This function will then add all the Listeners to the event object.

That way the core package is able to interact with any amount of extensions without knowing them.  
The event singleton may grow in the future and will be the only gate between extensions and the core.

[...]

PR #4301, opened 2019-04-11
XPath Isolation

analyticsdept:
The goal of this is to remove reliance on any hardcoded XPath values by moving them into a separate file, minimizing risk to the code when edits are made and enabling simpler maintenance.

- Created new XPath reader function (read_xpath) to handle parsing of a JSON file containing the necessary XPaths; done this way to easily transition to a hosted XPaths file in the future

- Added an XPath JSON compiler util - maintain the XPaths in this Python file; when run it will overwrite the existing xpath.json file in the instapy directory. The JSON structure follows the pattern `function name: { xpath name: xpath value [, ...] } [, ...]`

- Updated all of the functions within InstaPy to use the read_xpath function instead of relying on hard-coded values


instapy/xpath.json
```diff
@@ -0,0 +1 @@
...op_elements": "//main/article/div[1]"}], "get_given_user_followers": [{"followers_link": "//a[@href=\"/{}/followers/\"]"}], "get_follow_requests": [{"view_more_button": "//button[text()='View More']", "list_of_users": "//section/div"}], "like_comment": [{"comment_like_button": "..", "span_like_elements": "//span[@aria-label='Like']", "comments_block": "//div/div/h3/../../../.."}], "get_source_link": [{"image": "//img[@class=\"FFVAD\"]", "video": "//video[@class=\"tWeCl\"]", "image_alt": "//img[@class=\"_8jZFn\"]"}], "get_given_user_following": [{"all_following": ["//a[contains", "(@href,'following')]/span"], "following_link": "//a[@href=\"/{}/following/\"]", "topCount_elements": "//span[contains(@class,'g47SY')]"}], "dialog_username_extractor": [{"person": "../../../*"}], "get_following_status": [{"follow_button_XP": "//button[text()='Following' or                                   text()='Requested' or                                   text()='Follow' or                                   text()='Follow Back' or                                   text()='Unblock']"}], "get_links_from_feed": [{"get_links": "//article/div[2]/div[2]/a"}], "get_cord_location": [{"json_text": "//body"}]}
```
ishandutta2007: My understanding from the discussion under #4279 was that we are to going to fetch this file over the network than hardcoding into codebase such that code instapy don't have to be updated for every xpath changes. @breuerfelix correct me if my understanding is wrong. @breuerfelix I agree with you on the part that this can be fetched from a CDN than a data store since it is static.

instapy/xpath.py
```diff
@@ -0,0 +1,9 @@
+import json
+import os
+def read_xpath(function_name, xpath_name):
```
breuerfelix, line 3: just import xpath object from xpath_compile here, and use it directly from memory :)

breuerfelix:
I really like this refactor !! finally instapy is getting split up part by part into a nice project :)

I got 2 suggestions:  
1) not binding xpaths to a function name, rather grouping them by the pages they are on.   example groups: `login_page`, `profile`, `story`, `post` so if we look at the groups, we know to which kind of page they belong on instapy. since xpaths are related to the instagram web page, for me it makes sense to Group them by these pages :)

let me know what you think ! (but this refactor could be also done in the future !)

2) we read in the json file for every xpath, I would delete the json file for now :) since you overwrite it on every start anyways, why don't just keep the xpaths in memory ? we can save a lot of disk reads :)  
just import the xpath object directly in the xpath.py file and we are done :) 

btw this work is awesome ! I can imagine this being a lot of work separating all these xpaths and it was really needed ! this will be so damn handy in the future :) thanks !!

instapy/xpath_compile.py
```diff
@@ -0,0 +1,231 @@
...
+    "comment_elem":"//button/span[@aria-label='Comment']"
+})
+
+xpath['unfollow'] = []
+xpath['unfollow'].append({
+    "following_link":'//section//ul//li[3]',
+    "find_dialog_box":"//div[text()='Following']/../../../following-sibling::div"
+})
+
+with open('xpath.json', 'w') as outfile:
```
breuerfelix: delete this

instapy/xpath.py
```diff
@@ -0,0 +1,9 @@
...
+    fn = os.path.join(os.path.dirname(__file__), 'xpath.json')
+    with open(fn) as json_file:
```
breuerfelix: delete that

instapy/xpath.py
breuerfelix, line 3: messed up this comment ... :D

instapy/xpath.json
breuerfelix: delete this file

breuerfelix (changes requested):

analyticsdept:
> 1. not binding xpaths to a function name, rather grouping them by the pages they are on.   example groups: `login_page`, `profile`, `story`, `post` so if we look at the groups, we know to which kind of page they belong on instapy. since xpaths are related to the instagram web page, for me it makes sense to Group them by these pages :)
> 
> let me know what you think ! (but this refactor could be also done in the future !)

I was a little hesitant to do this since some of the functions that used XPaths are out of context and to be honest I have no idea which screen they relate to. I'm happy to do this, but is there a place on here that we can document which function relates to which screen? For now I'm going to make all of the other changes and commit them and continue that work as an enhancement if that's cool?

> 1. we read in the json file for every xpath, I would delete the json file for now :) since you overwrite it on every start anyways, why don't just keep the xpaths in memory ? we can save a lot of disk reads :)
>    just import the xpath object directly in the xpath.py file and we are done :)

You're totally right about the disk reads - if we ever move to hosted/CDN XPaths this can be looked at again!

> btw this work is awesome ! I can imagine this being a lot of work separating all these xpaths and it was really needed ! this will be so damn handy in the future :) thanks !!

No problem at all! It definitely took a minute or two :P

instapy/xpath.py
analyticsdept: Cool, I've made this change - it'll be in my new commit

instapy/xpath.py
```diff
@@ -0,0 +1,6 @@
+from .xpath_compile import xpath
+
+def read_xpath(function_name, xpath_name):
+    for p in xpath[function_name]:
+        xpath_val = p[xpath_name]
```
breuerfelix: this would be easier:
return xpath[function_name][xpath_name]

instapy/xpath_compile.py
```diff
@@ -0,0 +1,226 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+xpath = {}
+
+xpath['bypass_suspicious_login'] = []
```
breuerfelix: since you only got one element in the array, and its one big key values store you can do it more easily:
xpath['bypass_suspicious_login'] = {
     "bypass_with_mobile_choice":"//label[@for='choice_0']",
     "bypass_with_mobile_button":"//label[@for='choice_0']",
     "choice":"//label[@for='choice_1']",
     "choice_no_such_element":"//label[@class='_q0nt5']",
     "choice_exception":"//label[@class='_q0nt5 _a7z3k']",
     "close_button":"[text()='Close']",
     "security_code_field":"//input[@id='security_code']",
     "send_security_code_button":"//button[text()='Send Security Code']",
     "submit_security_code_button":"//button[text()='Submit']",
     "this_was_me_button":"//button[@name='choice'][text()='This Was Me']",
     "wrong_login":["//p[text()='Please check the code we sent you and try ","again.']"]
 }

breuerfelix (commented):
thanks for the refactor! looks way clear now :)  
the most important thing is, that each xpath is resolved by the same function, because with that refactor its really easy in the future to refactor it to a cdn !

I am so sorry I just read over it one more time cause I wanted to propose the merge for dev branch and then I saw another thing...

I made comments in the code :) why do you made a list in that case ? maybe my idea with the nested key value store might be not that good because I don't think about some details you thought about in the past :D but until now I would suggest just doing this key value store cause we don't need the list anyways.

But btw of course we could keep the names as they are right now ! thats just something we could change in the future :D

instapy/xpath.py
analyticsdept: Cool, done - thanks for simplifying it :)

instapy/xpath_compile.py
analyticsdept: Way cleaner! Thanks for the suggestion - added to the new commit :)

analyticsdept:
> I made comments in the code :) why do you made a list in that case ? maybe my idea with the nested key value store might be not that good because I don't think about some details you thought about in the past :D but until now I would suggest just doing this key value store cause we don't need the list anyways.

A key/value store is cool - I just like neat groups so its easier to identify relationships between things in the code. That said, it's totally just a personal preference!

I think there's even more work that can be done from a cleanliness perspective - like having a predefined template for the objects... In a few cases there are alternate XPaths for NoSuchElement exceptions, and there's also no naming conventions for buttons, inputs, etc - so it may be a good idea to document a standard for how to deal with capturing those. In some cases XPaths are referenced inline and others are stored in local variables at the top of a function, so all of that could be cleaned up :)

```
function_name: {
element_name: {
xpath: "",
exception_xpath: ""
}
}
```

Thanks for all the feedback! Hopefully I've covered everything in my latest commit but let me know if I've missed something :)

breuerfelix (approved):

breuerfelix:
> A key/value store is cool - I just like neat groups so its easier to identify relationships between things in the code. That said, it's totally just a personal preference!

No I was talking about the array inside the first key value :P I love the idea of the grouped key value store ! really cool idea :)

thanks for the refactor, it looks really clean and neat now :P 

now its up to @timgrossmann to merge this one ! looking forward to it :)

timgrossmann:
@analyticsdept @breuerfelix Awesome guys, looks much cleaner now and I really like the structure of the xPaths! :raised_hands: 

In terms of code it's definitely ready to be merged.
We just need to make sure we test it before releasing it to the people.

Thank you very much!

timgrossmann:
Just gave it a quick test and seems to work fine! 
I'll do more testing before we merge the dev branch into master.

Thank you very much for isolating the xPaths from the codebase :blush:

analyticsdept:
@timgrossmann thanks to you and @breuerfelix for reviewing and the helpful comments... and you're totally welcome :) Hopefully it's going to save a few headaches in the future!

ishandutta2007 (commented):
?

instapy/xpath.json
breuerfelix: @ishandutta2007 you are totally right ! but this PR is the first step towards this :)
now we call a function for each xpath so we just have to change this function and we are done, it will be kinda easy now, this step HAS to be done no matter what.

So this PR was like a preparation for the CDN :) maybe I got time to finish it or got time to make a proper issue describing the behaviour in the next days :)

ishandutta2007: cool, just wanted to be sure I am on same page as I haven't gone through rest of the discussions in the CR. @breuerfelix I was discussing with @timgrossmann of a possible usecase for CDN for pods feature as well but for now will use DB.

analyticsdept: @ishandutta2007 yeah, once it's decided which CDN is right for this project, we can build the pull mechanism.

PR #4308, opened 2019-04-12
catch exception for each post separately

ishandutta2007:
Let's not punish the whole pods run for one bad data case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ The **goal** of this file is explaining to the users of our project the notable 
 _The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)_.
 
 
+## [0.4.2] - 2019-04-15
+### Fixed
+- Fail of whole pod run on exception
+
+
 ## [0.4.1] - 2019-04-06
 ### Added
 - Support for split database with -sdb flag to avoid SQLite lock up
--- a/instapy/__init__.py
+++ b/instapy/__init__.py
@@ -8,5 +8,5 @@ from .file_manager import get_workspace
 
 
 # __variables__ with double-quoted values will be available in setup.py
-__version__ = "0.4.1"
+__version__ = "0.4.2"
 
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -5228,6 +5228,7 @@ class InstaPy:
         web_address_navigator(self.browser, user_link)
         try:
             pod_post_ids = get_recent_posts_from_pods(topic, self.logger)
+            self.logger.info("Downloaded pod_post_ids : {}".format(pod_post_ids))
             sleep(2)
             post_link_elems = self.browser.find_elements_by_xpath("//a[contains(@href, '/p/')]")
             post_links = []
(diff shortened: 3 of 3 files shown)


timgrossmann:
@ishandutta2007 Good idea! Thank you very much.
I think you're right about PRing it to the master, it seems like it is a bug fix :+1: 

Let's merge it together with another one of the PRs.

Also, do you think it would be possible to explain why commenting is necessary, in a short section, to the community. I hear a lot of people complaining and struggling since they get all the same comments at their posts since noone changes the comments :sweat_smile:

timgrossmann (approved):

ishandutta2007:
Commenting is actually more important in driving instagrams trending algorithm than liking.
Wheather commenting is necessary or not should not be a question. If we disable commentiing altogether it would hardly have any impact on instagrams algorithm. The issues here are not about volume, everyone loves volume, it's about relevance. 

The solutions needed are:
1. more comments, niche specific comments and split users into pod topics. 200 comments for each niche,  split into about 5 niche(not more). We are already closing on 100 posts a day in general within 2 weeks of adding it to templates, It won't take more than 2-3 months to reach the same volume for each of the topics. I am researching from time to time seeing the useage data what those first 5 topics should be.

2. We can expose the percent of posts it's engaging with that it should comment on(curently it's hardcoded to 80%). But I belive the number shouldn't be allowed to configure between 0-100% but something between 40-80%.

timgrossmann:
@ishandutta2007 True, but who is going to create the comments? 
Considering the amount of posts in there, all getting the same 10 comments over and over again will probably trigger Instagram to block us... Especially if more and more people are using the pods.

We need huge lists of default comments for the lists. 
I think we should disable them for now until we have the comment lists ready so it's not as spammy as it is now

ishandutta2007:
Humans intuitively don't have millions of pickup lines to choose from. If you look back at all your lines from the birth of internet till today I am sure you wont be able to list more than a couple of 100 complements that you ever wrote on any social media. Just do this exercise, check a post with say 10k comments you will see lots of similar comments without any automation tool whatsoever. Few hundreds per niche will be more than enough.

I think @breuerfelix had an idea for entering all form of data via UI by different user and download that list once a month from CDN or so and refresh. That will probably be the way for comments as well since comments also falls under data.

> probably trigger Instagram to block us

We are doing far riskier things in instapy than this, afaik instagram doesn't do blocking by comment analysis much.

>I think we should disable them for now

As far as disabling is concerned, the simplest way to disable is by removing the line `session.join_pods()` from your template. Since likes don't count much as a form of engagement it's as good as not using pods at all.

ishandutta2007:
Hey the simplest quick solution I can think of to avoid all users using the same comments from template is by 

1. Forcing them to enter new comments as soon as they start the script via commandline. once he types in these comments, it will be stored locally and won't ask you to enter for future runs. commandline is important as it would prevent their temptation to copy paste from somewhere.

2. We'll not only save this entered comments-list locally but also upload it to instapy pods server that way the server comments list will be 10k comments when 1k users join pods and so on. Even if we dedupe from 10k comments we will still be by my estimation over 1k comments. This way we ensure comments list and users list grow at the same rate.

3. comments used at any instant will be union of 
-  sample of comments downloaded(say random 100 out of 1k), 
-  hardcoded comments in template(10 as we already have)(also we may relax this limit check here as we will put the limit on commandline) and 
-  his own entered comments via commandline(is going to be 10 as well). 

I will change the code over the weekend as per this. Let's know if you have any concerns.

PR #4334, opened 2019-04-18
Remove redundant assignment by correting logic

matbrgz:
Assigning a variable to itself is useless and very likely indicates an error in the code.

Assigning a variable to itself is redundant and often an indication of a mistake in the code.


timgrossmann:
@MatheusRV Absolutely true! Thank you very much for the fix :+1:

PR #4335, opened 2019-04-18
Comparison using is when operands support __eq__

matbrgz:
Comparison using 'is' when equivalence is not the same as identity

When you compare two values using the is or is not operator, it is the object identities of the two values that is tested rather than their equality. If the class of either of the values in the comparison redefines equality then the is operator may return False even though the objects compare as equal. Equality is defined by the __eq__ or, in Python2, __cmp__ method. To compare two objects for equality, use the == or != operator instead.

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -1697,10 +1697,10 @@ class InstaPy:
         not_valid_users = 0
 
         # if smart hashtag is enabled
-        if use_smart_hashtags is True and self.smart_hashtags is not []:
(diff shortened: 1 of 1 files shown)


breuerfelix:
I changed base to dev, PR approved by me, you are totally right, here want to test for equality and not for identity !

timgrossmann:
@MatheusRV Absolutely true! Thank you for PRing this 😊 

@breuerfelix Thank you Felix for approving this!

PR #4336, opened 2019-04-18
Variable defined multiple times

matbrgz:
Assignment to a variable occurs multiple times without any intermediate use of that variable

Multiple assignments to a single variable without an intervening usage makes the first assignment redundant. Its value is lost.


CharlesCCC:
I don't think it is redundant, I believe it used for user who need DB location override.

timgrossmann:
@MatheusRV @CharlesCCC Checking the scope of the variable and taking into consideration that the address variable is reassigned before being used for any checks indicates that it actually is redundant.

Thank you very much for the update of quality for this project!

PR #4344, opened 2019-04-22
 Fix 'Failed to get comments'

dbknv:
'Not commenting! Failed to get comments' count!' sometimes happens with message 'Message: javascript error: Cannot read property 'count' of undefined'


timgrossmann:
@sterid Nice fixes! Please give us a bit to review and double check the functionality before we merge this.
Thank you for contributing to InstaPy!

timgrossmann:
@sterid What does the change of the media type fix?
Mainly, what is the usage of Carousel in the code?

Also, can we safely remove `' '` from the media type list?

dbknv:
I was trying to fix broken media post type filter for **like_by_tags** and **like_by_locations**. It doesn't work for years for me. 

I left possible filter value the same: Photo, Video and All. Also I think carousel is the same as photo post. Therefore when media=Photo it can be Photo or Carousel. 

Now **get_links** checks DIV on each post. And it can be one of 'Photo', 'Carousel' or 'Video' type. 
image

> Also, can we safely remove ' ' from the media type list?

Did you mean **get_links_for_tag | get_links_for_location** ? I think so. I don't how it works in original code because tag **a** always returns ' '

timgrossmann:
@sterid Hey, sorry for the long delay...
Could you please rebase to the latest version of InstaPy dev 🙈

timgrossmann:
#4404

PR #4404, opened 2019-05-08
Fix get comments count and media type filter

dbknv:
This is rebasing codebase from [4344](https://github.com/timgrossmann/InstaPy/pull/4344) to DEV. 

**Get comment count**
'Not commenting! Failed to get comments' count!' sometimes happens with message 'Message: javascript error: Cannot read property 'count' of undefined'. To fix it I added fallback JS.

**Media Type Filter**
I was trying to fix broken media post type filter for _like_by_tags_ and _like_by_locations_. It doesn't work for years for me. 

I left possible filter value the same: Photo, Video and All. In my opinion carousel is the same as photo post. Therefore when 

- media=Photo it can be Photo or Carousel
- media=Video it only can be Video
- media=All it can be Photo, Carousel or Video

To fix broken behaviour I changed **get_links** to check DIV on each post. 
image
It can be one of 'Photo', 'Carousel' or 'Video' type.


timgrossmann:
@sterid Just tested it locally. Seems to still work fine.
Thank you very much for adjusting!

dion-ricky:
I still get error failed to get comments count. Anyone can help?

image

PR #4412, opened 2019-05-09
End threaded session

hanagantig:
Threaded session ending without exception `ValueError: signal only works in main thread`


timgrossmann:
@hanagantig Good catch of this edge case.
Would you be able to also adjust the Documentation file and update add a small section about this use case there, please? 😊

timgrossmann:
@hanagantig Perfect, thank you very much! 
Really appreciate it

hanagantig:
You are welcome! I'm glad to be helpful.

PR #4440, opened 2019-05-16
Redis cache for likes

krazybean:


timgrossmann:
@krazybean Even though I really like the idea of having a key value db in the project for like tracking, I think we should implement what you've did with the already implemented DBs of the GUI branch and not add even more databases.

Thank you very much for contributing to InstaPy though!
We really appreciate it!

PR #4515, opened 2019-06-05
Added import Settings

InfraRedApple:
The `from instapy import Settings` is required when you are using Settings methods.


timgrossmann (approved):

timgrossmann:
@InfraRedApple Absolutely right, thank you very much for updating the docs to avoid confusion!

Issue #4517, opened 2019-06-05
session.follow_user_following(...) erroring: '''Message: invalid argument: 'value' must be a string'''


ZKing1000:
## Problem Code:
```session.follow_user_following(['sasha_rude', 'sashapcb', 'space_trashy', 'elon'], amount=10, randomize=True, sleep_delay=1)```

## Error:
> Traceback (most recent call last):
>   File "instagram.py", line 39, in <module>
>     session.follow_user_following(['sasha_rude', 'sashapcb', 'space_trashy', 'elon'], amount=10, randomize=True, sleep_delay=1)
>   File "/home/holden/.local/lib/python2.7/site-packages/instapy/instapy.py", line 3360, in follow_user_following
>     self.logfolder)
>   File "/home/holden/.local/lib/python2.7/site-packages/instapy/unfollow_util.py", line 1108, in get_given_user_following
    browser.find_element_by_xpath(read_xpath(get_given_user_following.__name__,"all_following")).text)
  File "/home/holden/.local/lib/python2.7/site-packages/selenium/webdriver/remote/webdriver.py", line 394, in find_element_by_xpath
    return self.find_element(by=By.XPATH, value=xpath)
  File "/home/holden/.local/lib/python2.7/site-packages/selenium/webdriver/remote/webdriver.py", line 978, in find_element
    'value': value})['value']
  File "/home/holden/.local/lib/python2.7/site-packages/selenium/webdriver/remote/webdriver.py", line 321, in execute
    self.error_handler.check_response(response)
  File "/home/holden/.local/lib/python2.7/site-packages/selenium/webdriver/remote/errorhandler.py", line 243, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.InvalidArgumentException: Message: invalid argument: 'value' must be a string
  (Session info: chrome=75.0.3770.80)
  (Driver info: chromedriver=2.46.628388 (4a34a70827ac54148e092aafb70504c4ea7ae926),platform=Linux 4.15.0-43-generic x86_64)

## Fix (~/.local/lib/python2.7/site-packages/instapy/unfollow_util.py):

>        ###### OLD ####
>        #allfollowing = format_number(
>        #    browser.find_element_by_xpath(
>        #        read_xpath(
>        #            get_given_user_following.__name__,"all_following")
>        #        )
>        #    .text
>        #)
>        #### END OLD ####

```
        ## NEW ##
        allfollowing = format_number(
[...]

PR #4518, opened 2019-06-06
Fix for session.follow_user_following() broken behavior #4517

ZKing1000:
In reference to issue #4517


InfraRedApple:
I implemented this change, and ran InstaPy and the error still was thrown.

timgrossmann (approved):

timgrossmann:
@ZKing1000 Thank you very much for this fix!
In the issue, I saw that the problem seems to appear at other places, too.

Would you be able to also PR them for the other places? 🙈 

Thanks :v:

ZKing1000:
Oh hey, just saw this. Sure I'll see about it. 😊

PR #4520, opened 2019-06-07
Fix for get_relationship_counts broken behavior

InfraRedApple:
This error is thrown when `get_relationship_counts` is called. Removing the `.text` fixes the error.


ZKing1000:
I think our issues may be related... #4517

lucicocosta:
i removed .text  another error occurred....

timgrossmann:
@InfraRedApple Doesn't this also need a conversion to `str`?

InfraRedApple:
Yes, it would require a str() conversion.

PR #4577, opened 2019-06-24
Extension of the current "Restricting Likes" Features

AKwoKWH:
This is a extension of the current "Restricting Likes"
https://github.com/timgrossmann/InstaPy/blob/master/DOCUMENTATION.md#restricting-likes

An addition option is created:
```
words starting with * will match the full caption of the image
(e. g. *cat matches "These are my lovely cat." even there are no hashtag in the caption)
```

### DIFF
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -629,7 +629,9 @@ def check_link(browser, post_link, dont_like, mandatory_words,
     dont_like_regex = []
 
     for dont_likes in dont_like:
-        if dont_likes.startswith("#"):
+        if dont_likes.startswith("*"):
+            dont_like_regex.append(dont_likes.replace("*",""))
+        elif dont_likes.startswith("#"):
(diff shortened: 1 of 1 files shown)


timgrossmann:
@AKwoKWH Could you please add a short section about this new feature to the DOCUMENTATION.md? 😊

For now, I've changed the base branch to develop since this is not only a fix but also some additional functionality.
We'll merge this into dev first and then with the next release into master.

Thank you very much for this addition!

breuerfelix:
@AKwoKWH sorry for mentioning the documentation again but it would be a pleasure for us to finally merge this pull request, so all here left is a small entry in the documentation file :)

It would be a shame if this pull request will be open for months only because the docs are left :D the earlier we merge this, the less problems we have while merging :)

dmfaria:
This way, you can even use a regular expression after *. It would be nice to have this documented.

MedRedha (approved):

Issue #4540, opened 2019-06-12
session.set_action_delays do not work


Tachenz:
Hi guys... recently got banned on Instagram because - 
session.set_action_delays(enabled=True,
                           like=90) setting started to get completely ignored.

Before it would rest for 90 seconds after liking. Now - it just going non stop... Did that get broken somehow?

Running latest version of instapy. Been working fine untill about 10 days ago.

pneumokoniose:
Same here

Tachenz:
For now I am just using sleep delay in like util py again, but that is obviously not ideal - as it will be the same for all accounts running from the same installation

billidani7:
Same here on a raspberry pi 3+

Svmurvj:
I can confirm that 

Here is how I set that: 
**session.set_action_delays(enabled=True, like=30, follow=30, random_range=(80, 140))**

So follow delays should be:

- min 30s * 80% = 24s
- max 30s * 140% = 42s

Here is peace from log:

INFO [2019-06-18 11:52:30]   Ongoing Follow [25/42]: now following 'x'...
INFO [2019-06-18 11:52:32]   User: 'x'  |> followers: 62  |> following: 306  |> relationship ratio: 0.2
INFO [2019-06-18 11:52:40]   --> Followed 'b'x''!
INFO [2019-06-18 11:52:40]   --> x added to blacklist for z campaign (action: followed)
INFO [2019-06-18 11:52:46]   Follow per user: 24  |  Total Follow: 24

INFO [2019-06-18 11:52:46]   Ongoing Follow [26/42]: now following 'y'...
INFO [2019-06-18 11:52:51]   User: 'y'  |> followers: 192  |> following: 367  |> relationship ratio: 0.52
INFO [2019-06-18 11:52:58]   --> Followed 'b'y''!
INFO [2019-06-18 11:52:58]   --> y added to blacklist for z campaign (action: followed)
INFO [2019-06-18 11:53:06]   Follow per user: 25  |  Total Follow: 25

I see the time difference between    Followed 'b'x''!    and    Followed 'b'y''!     is 18s, depend on setting it should be >= 24s. 

Is my measurement correct in that case depending on log ? Even when I set this to 50s the result was similar. 

Can anybody check that and confirm ?

dpascoa:
Mine is not working too. Something happened with action_delay

octavius-1:
My action_delay is not working either, any updates? I'm massively following people with no wait times between actions

[...]

PR #4606, opened 2019-07-02
Fix get_action_delay method always returning default values #4540

MichalGallovic:
Hey, this should fix `get_action_delay` bug #4540, returning default values even after updating values with `set_action_delays`.

Not sure whether to set merge into master or dev.

EDIT:

Hey this was just a one-liner change.

On master there is a bug (#4540) in `get_action_delay` method. It checks if `config[action]` is `int,float` and if yes it returns default value - so even if we change action delays with `set_action_delays` default value is always returned. I reversed the logic to `is not True` which fixes the bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ The **goal** of this file is explaining to the users of our project the notable 
 _The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)_.
 
 
+## [0.5.4] - 2019-07-03
+### Fixed
+- `get_action_delay` method always returning default values #4540
+
+
 ## [0.5.3] - 2019-07-02
 ### Fixed
 - Argument Being Interpreted as Sequence in `bypass_suspicious_login`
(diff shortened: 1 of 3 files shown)


Svmurvj:
Create a Pull request (to the dev branch), reference the issue number and explain what you have changed/deleted/added to the project in depth and then someone will review your work and may suggest changes or ask questions

MichalGallovic:
@Svmurvj Changed from `master` to `dev`. Also explained in more detail. Issue has already been referenced in the title and first comment. Thanks!

h2jun:
In my case, I solved it with type method, not isinstance method.
`isinstance(config[action], (int, float))` -> `type(config[action]) not in [int, float]`
`all((isinstance(i, (type(None), int, float)) for i in` -> `all((type(i) in [type(None), int, float] for i in`
`any(not isinstance(i, type(None)) for i in config["random_range"])):` -> `any(type(i) is not None for i in config["random_range"])):`

good luck.

timgrossmann:
@MichalGallovic Thank you very much for the fix and really good catch!

Congrats to your first merged PR to InstaPy btw. 🙌

timgrossmann (approved):

MichalGallovic:
Thanks Tim! 😁

Issue #4619, opened 2019-07-04
Set action delay fix results in new issue


Tachenz:
![2DA672F3-6531-4067-B739-8967E76C279B](https://user-images.githubusercontent.com/36073534/60675288-eca25500-9e73-11e9-9a54-e90d42c35e4c.jpeg)

Issue #4615, opened 2019-07-04
Bug at launch 


Micktoon:
## Expected Behavior
The bot should launch the chrome window and start working 

## Current Behavior
However, this one doesn't launch anything and indicates the following error code, but from time to time it will still work.

```
Traceback (most recent call last):
  File "quickstart.py", line 26, in <module>
    with smart_run(session):
  File "/usr/lib/python3.7/contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "/home/pi/.local/lib/python3.7/site-packages/instapy/util.py", line 1679, in smart_run
    session.login()
  File "/home/pi/.local/lib/python3.7/site-packages/instapy/instapy.py", line 407, in login
    self.bypass_with_mobile):
  File "/home/pi/.local/lib/python3.7/site-packages/instapy/login_util.py", line 173, in login_user
    reload_webpage(browser)
  File "/home/pi/.local/lib/python3.7/site-packages/instapy/util.py", line 1706, in reload_webpage
    browser.execute_script("location.reload()")
  File "/home/pi/.local/lib/python3.7/site-packages/selenium/webdriver/remote/webdriver.py", line 636, in execute_script
    'args': converted_args})['value']
  File "/home/pi/.local/lib/python3.7/site-packages/selenium/webdriver/remote/webdriver.py", line 321, in execute
    self.error_handler.check_response(response)
  File "/home/pi/.local/lib/python3.7/site-packages/selenium/webdriver/remote/errorhandler.py", line 242, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.WebDriverException: Message: unknown error: cannot determine loading status
from unknown error: unknown sessionId
  (Session info: chrome=74.0.3729.157)
  (Driver info: chromedriver=2.36 (a81166ad79e68fbfe7cf5ba243192d6412e26b37),platform=Linux 4.19.50-v7+ armv7l)
```

## InstaPy configuration
Instapy is install in Raspbery Pi B+

Thx,

Micktoon:
I have a problem with likes, the bot does 1 and crashes with error 

```
Traceback (most recent call last):
  File "quick.py", line 59, in <module>
    amount=random.randint(50, 100), interact=True)
  File "/home/pi/.local/lib/python3.7/site-packages/instapy/instapy.py", line 1811, in like_by_tags
    liked_img)
[...]

Issue #4610, opened 2019-07-03
KeyError: 'like'


JustAskMarina:
Hi, this has started happening today every time I run Instapy and it hits at the first like 

`Traceback (most recent call last):
  File "C:\Users\###\Telegram-InstaPy-Scheduling\scripts.py", line 208, in ###_script
    interact=True)
  File "C:\Users\###\AppData\Local\Programs\Python\Python37-32\lib\site-packages\instapy\instapy.py", line 1811, in like_by_tags
    liked_img)
  File "C:\Users\###\AppData\Local\Programs\Python\Python37-32\lib\site-packages\instapy\like_util.py", line 696, in like_image
    naply = get_action_delay("like")
  File "C:\Users\###\AppData\Local\Programs\Python\Python37-32\lib\site-packages\instapy\util.py", line 1771, in get_action_delay
    custom_delay = config[action]
KeyError: 'like'`

I have 3 setups of the bot on different machines, only one has this problem and I can't really understand why.
Win10, Chrome 73, instapy-chromedriver and every other dependency updated.

hijak:
im also getting this error

```
Traceback (most recent call last):
  File "insta.py", line 17, in <module>
    bot.like_by_tags(['XX', 'XX', 'XX', 'XX', 'XX', 'XX', 'XX'], amount=100)
  File "/usr/local/lib/python3.7/site-packages/instapy/instapy.py", line 1811, in like_by_tags
    liked_img)
  File "/usr/local/lib/python3.7/site-packages/instapy/like_util.py", line 696, in like_image
    naply = get_action_delay("like")
  File "/usr/local/lib/python3.7/site-packages/instapy/util.py", line 1771, in get_action_delay
    custom_delay = config[action]
KeyError: 'like'

```

saahil1292:
I got the fix of this. Open the file "like_util.py" and go to the line where "get_action_delay("like") and just just change from "like" to "Like". Hope it helps :)

skaeslin:
I have the same Problem. Sadly your fix did not work for me. (or nor for long)

I run it on a raspbnerry pi 3 an everything is up to date (did a new install today)

```
OOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOO
INFO [2019-07-03 23:25:28] [sk]  Session started!
oooooooooooooooooooooooooooooooooooooooooooooooooooooo


...................................................................
[...]

PR #4620, opened 2019-07-04
Fix missing check for action and uninitialized action_delays in settings

MichalGallovic:
Hey,

yesterday's fix of `get_action_delays` method https://github.com/timgrossmann/InstaPy/pull/4606/commits/b71563f7381e48c2bd8cccac1a861f720ddbb452 caused another bug along the line #4610 #4619 #4622 #4610 #4615

`Settings.action_delays` is not initialized properly. As suggested by @lvkins I'm adding another check to see if we have any actions set in `Settings.action_delays`.

Maybe we should initialize `Settings.action_delays` in `__init__` so we have some sensible defaults ?

### DIFF
--- a/instapy/util.py
+++ b/instapy/util.py
@@ -1765,7 +1765,7 @@ def get_action_delay(action):
             action not in config or
             config["enabled"] is not True or
             config[action] is None or
-            isinstance(config[action], (int, float))) is not True:
+            isinstance(config[action], (int, float)) is not True):
         return defaults[action]
 
     else:
--- a/instapy/util.py
+++ b/instapy/util.py
@@ -1762,6 +1762,7 @@ def get_action_delay(action):
     config = Settings.action_delays
 
     if (not config or
+            action not in config or
             config["enabled"] is not True or
             config[action] is None or
             isinstance(config[action], (int, float))) is not True:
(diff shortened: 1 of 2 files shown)


instapy/util.py
```diff
@@ -1762,6 +1762,7 @@ def get_action_delay(action):
     config = Settings.action_delays
 
     if (not config or
+            action not in config or
             config["enabled"] is not True or
             config[action] is None or
             isinstance(config[action], (int, float))) is not True:
```
lvkins: Should be `isinstance(config[action], (int, float)) is not True):`  otherwise, bug is still there because of bad parenthesis.

MichalGallovic: Thanks!

MichalGallovic:
Added `test_action_delays.py` to make sure, `get_action_delays` returns at least default values, when nothing is set.

Wanted to take this further and also test that `set_action_delays` changes values etc, but wasn't really sure about mocking instapy, selenium etc...

MichalGallovic:
Those `assert` warnings seem to be false positives.

breuerfelix:
thanks for this fix ! I experienced some people in discord having this issue :)

MichalGallovic:
@breuerfelix thx for merge and single quotes :) Yes there is quite a few comments in #4610 as well.

PR #4670, opened 2019-07-16
Feature/stories

christophe-duc:
Adding functionalities to InstaPy to be able to watch stories, either on a specific tag or from users.

PR #4673, opened 2019-07-17
Feature/stories v2

christophe-duc:
- added code to parse and understand the JSON story request
- fix various bugs (non existent page, or private page reported by @converge) 
- fix documentation (removed the word try reported by @converge )

passed testing:
  - interact OK
  - story_by_tags OK
  - story_by_users OK

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -1232,7 +1232,7 @@ session.set_do_story(enabled = True, percentage = 70)
 
 ### Watch stories by Tags
 
-Will watch up to 50 stories published with specified tags.
+Will watch up to 20 stories published with specified tags.
 
 ```python
 session.story_by_tags(['tag1', 'tag2'])
@@ -1240,7 +1240,7 @@ session.story_by_tags(['tag1', 'tag2'])
 
 ### Watch stories from users
 
-Take a list of users and try to watch their stories.
+Take a list of users and watch their stories.
 
 ```python
 session.story_by_users(['user1', 'user2'])
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -1227,8 +1227,12 @@ session.remove_follow_requests(amount=200, sleep_delay=600)
 Will add story watching while interacting with users
 
 ```python
-session.set_do_story(enabled = True, percentage = 70)
+session.set_do_story(enabled = True, percentage = 70, simulate=True)
 ```
+
+ `simulate`:
+ If set to `True` InstaPy will simulate watching the stories (you won't see it in the browser), we just send commands to Instagram saying we have watched the stories.
+ If set to `False` Instapy will perform the exact same action as a human user (clicking on stories, waiting until watching finishes, etc...)
 
 ### Watch stories by Tags
 
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -1227,7 +1227,7 @@ session.remove_follow_requests(amount=200, sleep_delay=600)
 Will add story watching while interacting with users
 
 ```python
-session.set_do_story(enabled = True, percentage = 70, simulate=True)
+session.set_do_story(enabled = True, percentage = 70, simulate = True)
 ```
 
  `simulate`:
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -198,6 +198,8 @@ class InstaPy:
         self.not_valid_users = 0
         self.video_played = 0
         self.already_Visited = 0
+        self.stories_watched = 0
+        self.reels_watched = 0
 
         self.follow_times = 1
         self.share_times = 1
@@ -4769,6 +4771,7 @@ class InstaPy:
                 "\t|> REPLIED to {} comments\n"
                 "\t|> INAPPROPRIATE images: {}\n"
                 "\t|> NOT VALID users: {}\n"
+                "\t|> WATCHED {} story(ies)  |  WATCHED {} reel(s)\n"
                 "\n{}\n{}"
                 .format(self.liked_img,
                         self.already_liked,
@@ -4780,6 +4783,8 @@ class InstaPy:
                         self.replied_to_comments,
                         self.inap_img,
                         self.not_valid_users,
+                        self.stories_watched,
+                        self.reels_watched,
                         owner_relationship_info,
                         run_time_msg))
         else:
@@ -5458,17 +5463,20 @@ class InstaPy:
                     break
 
                 # inform user whats happening
-                self.logger.info('Loading stories view...')
-                self.logger.info('Tag [{}/{}]'.format(index + 1, len(tags)))
+                if len(tags)> 1:
+                    self.logger.info('Tag [{}/{}]'.format(index + 1, len(tags)))
                 self.logger.info('Loading stories with Tag --> {}'.format(tag.encode('utf-8')))
 
                 try:
-                    watch_story(self.browser,
+                    reels=watch_story(self.browser,
                                 tag,
                                 self.logger, "tag")
                 except NoSuchElementException:
                     self.logger.info('No stories skipping this tag')
                     continue
+                if reels > 0:
+                    self.stories_watched += 1
+                    self.reels_watched += reels
 
     def story_by_users(self, users: list = None):
         """ Watch stories for specific user(s)"""
@@ -5485,12 +5493,15 @@ class InstaPy:
                     break
 
                 # inform user whats happening
-                self.logger.info('Loading stories view...')
-                self.logger.info('User [{}/{}]'.format(index + 1, len(users)))
+                if len(users) >1:
+                    self.logger.info('User [{}/{}]'.format(index + 1, len(users)))
                 self.logger.info('Loading stories with User --> {}'.format(user.encode('utf-8')))
 
                 try:
-                    watch_story(self.browser, user, self.logger, "user")
+                    reels=watch_story(self.browser, user, self.logger, "user")
                 except NoSuchElementException:
                     self.logger.info('No stories skipping this user')
                     continue
+                if reels > 0:
+                    self.stories_watched += 1
+                    self.reels_watched += reels
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -2217,7 +2217,6 @@ class InstaPy:
         followed = 0
         already_followed = 0
         not_valid_users = 0
-        story_watched = 0
 
         self.quotient_breach = False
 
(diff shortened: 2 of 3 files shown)


breuerfelix:
could you fix the travis bug maybe ? I don't know 100% what's wrong there but I think you first have to combine the args and then pass it as one list to the function :) correct me if I am wrong !

christophe-duc:
> could you fix the travis bug maybe ? I don't know 100% what's wrong there but I think you first have to combine the args and then pass it as one list to the function :) correct me if I am wrong !

corrected and re-tested

instapy/story_util.py
```diff
@@ -1,29 +1,136 @@
...
+                            "&variables={{\"reel_ids\":[\"{}\"],\"tag_names\":[],\"location_ids\":[],".format(elem_id) + \
+                            "\"highlight_reel_ids\":[],\"precomposed_overlay\":false,\"show_story_viewer_list\":true," + \
+                            "\"story_viewer_fetch_count\":50,\"story_viewer_cursor\":\"\"," + \
+                            "\"stories_video_dash_manifest\":false}"
+    cookies = browser.get_cookies()
+
+    s = requests.Session()
+
+    for cookie in cookies:
+        if (cookie['name'] == "urlgen") or (cookie['name'] == "rur"):
```
breuerfelix: just to avoid duplicate code (cause I hate that :D)

what about:

```python
all_args = {
                 'name': cookie['name'],
                 'value': cookie['value'],
                 'domain': cookie['domain'],
                 'secure': cookie['secure'],
                 'rest': {'HttpOnly': cookie['httpOnly']},
                 'path': cookie['path']
             }
if not (cookie['name'] == "urlgen") and not (cookie['name'] == "rur"):
             all_args['expires'] = cookie['expiry']
         
         s.cookies.set(**all_args)
```

I hope I negotiated the if statement correctly :D

breuerfelix:
I like the changes :) but I think @converge is more in the topic and lets wait for his review !! :)

Just one question, is it easy to add how many max stories to watch ? maybe per user and per session ?

its now hardcoded to 20 but it could be just a param right ? :) and the default value of this param will be 20. so nothing big will change.

timgrossmann:
@christophe-duc @breuerfelix One thing we could also take into account is leveraging the API more.
There are services that make use of an API endpoint called "mark all stories watched" which takes a list of stories enabling services to "view" a million stories a day. 

E.g. masslooking.net

instapy/story_util.py
christophe-duc: yes, i'll make the change. thanks!

christophe-duc:
> I like the changes :) but I think @converge is more in the topic and lets wait for his review !! :)
> 
> Just one question, is it easy to add how many max stories to watch ? maybe per user and per session ?
> 
> its now hardcoded to 20 but it could be just a param right ? :) and the default value of this param will be 20. so nothing big will change.

Here 20 is the max amount of reels_item I have seen being able to receive in the JSON response of the graphql query, for the moment I am not able to watch just one or to stop after a certain amount has been viewed as there is a lot of JS events involved and the code is quite obfuscated.

In the traces, I took, there are additional reporting that is done to instagram (when one reel was seen apparently), but I haven't played with it yet. (possibly being able to report something watched even if we haven't, could help with the other comment from @timgrossmann ) 

on @timgrossmann comment:
I have never played with instagram API before, but couple of comments:
  - the million story thing is obviously currently a big buzz as it seems instagram has forgot to control that, but I would expect a quick reaction from them.
  - the way it is integrated right now, is difficult, but it uses what's available to everybody when they uses the app/browser normally. So it is probably closer to what a human would do and might throw their bot detection algorithm off a little bit more. From the API, yes engagement is going to be great, but I currently don't see how that's going to help our bot sessions.

Let me know what you guys, think.

breuerfelix:
so basically, every change requested would mean to put a lot of work into this feature. so I guess we could say thats version 1, we should merge this PR and release it with 1.5.6 :))

instapy/story_util.py
```diff
@@ -1,29 +1,128 @@
...
 from .util import click_element
 from .util import web_address_navigator
+from .util import update_activity
 from .xpath import read_xpath
 
+import requests
 
-def watch_story(browser, elem, logger, action_type):
+
+def get_story_data(browser, elem: str, action_type: str, logger) -> dict:
```
converge: are we using strong typing elsewhere on InstaPy? if not, it would be nice to follow the style (weak typing) the majority of the code uses.

instapy/story_util.py
```diff
@@ -1,29 +1,128 @@
...
+    """not used fort the moment, more coding needed to understand this"""
+
+    # if things change in the future, modify here:
+    query_hash = "cda12de4f7fd3719c0569ce03589f4c4"
+
+    if action_type == "tag":
+        # pretty easy here, we just have to fill tag_names with the tag we want
+        graphql_query_url = "https://www.instagram.com/graphql/query/?query_hash" + \
```
converge: it's "cosmetic" as some people say, but I believe it's important to make it more readable, does it make sense?

```python
# this is just a quick example and doesn't work
graphql_query_url = (
            'https://www.instagram.com/graphql/query/?query_hash={}'
            '&variables={"reel_ids":[],"tag_names":["{}"],"location_ids":[],'
            'highlight_reel_ids": [], "precomposed_overlay": false, '
            'show_story_viewer_list": true, "story_viewer_fetch_count": 50, '
            'story_viewer_cursor": "", "stories_video_dash_manifest": false}'
            .format(query_hash, elem)
        )
```

instapy/story_util.py
```diff
@@ -1,29 +1,128 @@
...
+            return {'status': 'not_ok', 'reels_cnt': 0}
+        graphql_query_url = "https://www.instagram.com/graphql/query/?query_hash" + \
+                            "={}".format(query_hash) + \
+                            "&variables={{\"reel_ids\":[\"{}\"],\"tag_names\":[],\"location_ids\":[],".format(elem_id) + \
+                            "\"highlight_reel_ids\":[],\"precomposed_overlay\":false,\"show_story_viewer_list\":true," + \
+                            "\"story_viewer_fetch_count\":50,\"story_viewer_cursor\":\"\"," + \
+                            "\"stories_video_dash_manifest\":false}"
+    cookies = browser.get_cookies()
+
+    s = requests.Session()
```
converge: sorry for more cosmetic things, but I need to point it also, `session` instead of `s` will give more sense for what the code is doing.

converge (changes requested):
just these small changes and it's all good for me. and just as a note for future code, just a little more documentation, and it will be perfect. great work! 🤘 🤘 🤘

instapy/story_util.py
christophe-duc: the strong typing was introduced by @breuerfelix in the first merged branch of stories.

see here: [https://github.com/timgrossmann/InstaPy/pull/4670/commits/2c43359979b3ab9b19742e5575fa9d5e821c63ad](url)

That's why I re-used it, I though it was a new rule. Let me know what you guys decide and i'll update the function prototypes.

instapy/story_util.py
christophe-duc: fixed and rewrote the code to be simpler.

instapy/story_util.py
christophe-duc: no problem, fixed.

instapy/story_util.py
breuerfelix: well typing is not present in instapy at all but I have to get into typing for the gui (at least for the api functions in instapy) so I think that if you are able to write types, you always should because the compiler doesn't care if you give a string or an int (even if you said it should be string) the code won't break BUT as a programmer the code is way easier to understand :)

Sometimes its not possible (or hard possible) for example with the browser object. So yeah don't use types here if its to difficult, but in general I would say, always try to use types :)

what do you think about it ?

christophe-duc:
@timgrossmann : following your mass story comment, I have tried to evolve the code. I am happy to report that I am fully capable now to completely simulate the control plane of watching stories. It means I am able to make instagram believe you watched the stories, when you actually didn't perform the action on the web page.

Code is ready, I'll create 2 modes for watching stories, a real mode that uses the web interface as a user would do and a simulated mode that just send the commands to instagram.

Let me know if I shall include this into the PR.

instapy/story_util.py
converge: I would prefer to keep with the old InstaPy code style (weak typing). I think about new contributors facing two code style, can be confusing for them.

instapy/story_util.py
```diff
@@ -1,45 +1,171 @@
...
+    query_hash = "cda12de4f7fd3719c0569ce03589f4c4"
+    elem_id = ""
+
+
+    if action_type == "user":
+        try:
+            reel_id = browser.execute_script(
+                "return window._sharedData.entry_data."
+                "ProfilePage[0].graphql.user.id")
+            #correct formating for elem_id
```
breuerfelix: no whitespace after hashtag in comments :) since we wanna add that rule to the linter, better start correcting it right away :D

there are couple of these comments, just a small fix

instapy/story_util.py
```diff
@@ -1,45 +1,171 @@
...
+        # check how many reels we got
+        media_cnt = len(response['data']['reels_media'])
+
+        if media_cnt == 0:
+            # then nothing to watch, we received no stories
+            return {'status': 'ok', 'reels_cnt': 0}
+        else:
+            # we got content
+            # check if there is something new to watch otherwise we just return 0
+            seen=0
```
breuerfelix: whitespace also for things like this :) its just prettify :D

DOCUMENTATION.md
```diff
@@ -1227,20 +1227,24 @@ session.remove_follow_requests(amount=200, sleep_delay=600)
 Will add story watching while interacting with users
 
 ```python
-session.set_do_story(enabled = True, percentage = 70)
+session.set_do_story(enabled = True, percentage = 70, simulate=True)
```
breuerfelix: simulate = True (whitespace)

instapy/instapy.py
```diff
@@ -518,12 +521,20 @@ def set_do_like(self, enabled=False, percentage=0):
 
         return self
 
-    def set_do_story(self, enabled: bool = False, percentage: int = 0):
+    def set_do_story(self, enabled: bool = False, percentage: int = 0, simulate: bool=False):
```
breuerfelix: bool = False (whitespace)

instapy/story_util.py
```diff
@@ -1,45 +1,171 @@
...
 from .util import click_element
 from .util import web_address_navigator
+from .util import update_activity
 from .xpath import read_xpath
 
+import requests
 
-def watch_story(browser, elem, logger, action_type):
+
+def get_story_data(browser, elem: str, action_type: str, logger,simulate: bool=False) -> dict:
```
breuerfelix: logger, simulate: bool = False (3x whitespace)

breuerfelix (changes requested):
yep I am the human linter :D

maybe today I will take some time and specify the linter and also specify that the linter fixes whitespace itself (I always do that in my javascript projects) cause this is just a pain in the ass.

sorry that you have to do it manually right now :D

btw you are a beast ! really awesome feature programmed in this short amount of time. I wanna get it merged by today !

breuerfelix (approved):
thanks ! human linter is happy now :D

PR #4683, opened 2019-07-19
ability to choose a random user agent, latest chrome per default

breuerfelix:
This is a really basic implementation but it is better than nothing.

We still have to discuss wether it is okay to always have a random user agent enabled or per default just use the latest chrome agent.

Let the discussion start!

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -1756,6 +1756,12 @@ session.login()
 session.end(threaded_session=True)
 ```
 
+### Get a random UserAgent
+
+```python
+session = InstaPy(..., random_user_agent = True, ...)
+```
+
 ---
 
 <br /> 
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -69,6 +69,7 @@
   - [Switching to Firefox](#switching-to-firefox)
   - [Use a proxy (Firefox)](#use-a-proxy-firefox)
   - [Running in threads](#running-in-threads)
+  - [Random UserAgent](#get-a-random-useragent)
   
  <br />
 
--- a/instapy/browser.py
+++ b/instapy/browser.py
@@ -33,7 +33,8 @@ def set_selenium_local_session(proxy_address,
                                browser_profile_path,
                                disable_image_load,
                                page_delay,
-                               logger):
+                               logger,
+                               random_user_agent):
     """Starts local session for a selenium server.
     Default case scenario."""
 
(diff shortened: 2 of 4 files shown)


timgrossmann (approved):

converge:
that's a good point to discuss. As you pointed, use this is better than nothing for now, but I personally prefer a definitive solution.

My idea is:
1. check the OS
2. check browser
3. set user agent based on step 1 and 2

does it make sense?

breuerfelix:
2) check browser -> already implemented :)

1) check OS

thats maybe not optimal because alot of people are running the bot on ubuntu server oder raspberry pi for example. And its not possible to surf instapy with a raspbian lite or ubuntu server os :D 

But I also don't know how many information from the OS gets into the user agent, if its just smth like `Linux` for both ubuntu and raspbian for example, that would totally be fine. We should check that.

thanks for the review :)

ghost:
basically we can use this to make instagram think we are using it from our phone? if thats the case, all i have to do is add my phone chrome browser agent to browser.py? i saw that you added one for the iphone, im using an android. do you know where i can get that for my phone?

breuerfelix:
@dandy27 well you have to look through the web to figure this one out :)

But with this feature, it will always use the latest one :)

ghost:
> @dandy27 well you have to look through the web to figure this one out :)
> 
> But with this feature, it will always use the latest one :)

i see, so i was right?this is something very nice, sorry im new to programming. i will add this to my files right away. i want instagram to think that im always in my phone, not a browser since they had been acting stupid lately with whoever use browser

ghost:
> @dandy27 well you have to look through the web to figure this one out :)
> 
> But with this feature, it will always use the latest one :)

2019-07-21T15:01:39.440041700Z 
2019-07-21 15:01:39 (1.70 MB/s) - written to stdout [160/160]
2019-07-21T15:01:39.440109200Z 
Selenium is up - executing command
  File "code/docker_quickstart.py", line 1, in <module>
    from instapy import InstaPy
  File "/code/instapy/__init__.py", line 3, in <module>
    from .instapy import InstaPy
  File "/code/instapy/instapy.py", line 79, in <module>
    from .browser import set_selenium_local_session
  File "/code/instapy/browser.py", line 10, in <module>
    from fake_useragent import UserAgent
ModuleNotFoundError: No module named 'fake_useragent'



im sorry to bother you, i dont even know if this is the place. but after adding your mod, i get this error. my python is very limit and i cant find a way to fix it. trying to use it using docker

breuerfelix:
in the pr you see that I added a new requirement to this package.

so you have to go and pip install -r requirements.txt again :)

ghost:
> in the pr you see that I added a new requirement to this package.
> 
> so you have to go and pip install -r requirements.txt again :)

i know, and i already have it. so i dont understand why isnt working

Issue #4556, opened 2019-06-20
Interacting with same users


ghost:
## Expected Behavior
Skip users that bot has already interacted with.
## Current Behavior
When interacting with user followers, it keeps repeating people.
## InstaPy configuration
session = InstaPy(username="", password="", headless_browser=True)
       with smart_run(session):
                        session.set_simulation(enabled=False)
                        session.set_user_interact(amount=3, randomize=True, percentage=100, media='Photo')
                        session.set_do_like(enabled=True, percentage=75)
                        session.interact_user_followers(['user....'])

PR #4691, opened 2019-07-22
Added new feature: do not interact with same users [updated]

ghost:
## Reason for changes
Bot used to interact with same users over and over again when dealing with:

- Interact with someone else's followers

 - Interact with users that someone else is following

## What has been changed
- Added option to store users bot has already interacted with in database
- Added check for repetitive users, based on database entries
- Added instruction how to use new feature in documentation
- Updated to new version of InstaPy to avoid huge diffs

## Problem reference
#4556


breuerfelix:
really cool feature !  
I changed the default value to false so existing templates wont behave different :)

breuerfelix:
@mateuszpokojowczyk well I just read over the code and ... you wrote the function, thats pretty cool but, you never use it haha :D

so right now nothing will happen if you set it to true :)

ghost:
> @mateuszpokojowczyk well I just read over the code and ... you wrote the function, thats pretty cool but, you never use it haha :D
> 
> so right now nothing will happen if you set it to true :)

Well, gotta check it out, maybe made mistake while updating.
It was all working before update.

PR #4692, opened 2019-07-22
flake8 enhancement + linter fix

breuerfelix:
changes:
- extract flake8 config
- only run one flake command in tests
- fixed some lines too long errors
- fixed a forgotten import
- fixed a logger error
- fixed deprecated regex

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -8,10 +8,11 @@ select =
 	F821,
 	F822,
 	F823,
-	E501,
-	E225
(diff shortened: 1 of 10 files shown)


converge:
fixing style day is a happy day 🎉 🎉 🎉 
I had a quick overview, sounds like there is none breaking changing, right? (I want to approve it without testing every method)

*testing -> a subject for a nearby future

PR #4698, opened 2019-07-23
Speed up login process and Connection check feature

converge:
- speed up the login process by decreasing the implicit wait to 5 seconds
- new approach for connection verification, it will check if we can navigate, and then check Instagram.com website status
Captura de Tela 2019-07-23 às 10 51 17

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -398,6 +398,21 @@ class InstaPy:
 
     def login(self):
         """Used to login the user either with the username and password"""
+        # InstaPy uses page_delay speed to implicit wait for elements,
+        # here we're decreasing it to 3 seconds instead of the default 25 seconds
+        # to speed up the login process.
+        #
+        # In short: default page_delay speed took 25 seconds trying to locate every
+        # element, now it's taking 5 seconds.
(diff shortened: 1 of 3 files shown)


instapy/instapy.py
```diff
@@ -398,6 +398,15 @@ def set_selenium_remote_session(self, selenium_url='',
 
     def login(self):
         """Used to login the user either with the username and password"""
+        # InstaPy uses page_delay speed to implicit wait for elements,
```
breuerfelix: why not put that in the doc string ? :) would be awesome since then we also see it in the gui ;D

instapy/instapy.py
```diff
@@ -413,8 +422,9 @@ def login(self):
                             self.logger)
 
             self.aborting = True
```
breuerfelix, line 426: what do you think about a `return self` here and remove the else block ? I mean just use normal indention on the else block. If nothing else is after the else anyways, then I like less indention as possible :D

converge: all good for me @breuerfelix , thank you!

let's merge it 🤘 🤘 🤘

breuerfelix (approved):

PR #4701, opened 2019-07-23
fix useragent

breuerfelix:


instapy/browser.py
```diff
@@ -43,19 +43,23 @@ def set_selenium_local_session(proxy_address,
...
     # define the custom user agent
-    fb_agent = (
+    user_agent = (
         'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 '
         '(KHTML, like Gecko) Chrome/74.0.3729.169 Safari/537.36'
     )
-    ua = UserAgent(cache = False, fallback = fb_agent)
+    
+    try:
+        ua = UserAgent(cache = False, fallback = fb_agent)
```
converge: little mistake, it should be `user_agent` not `fb_agent`

instapy/browser.py
```diff
@@ -167,6 +169,9 @@ def set_selenium_local_session(proxy_address,
                 Settings.chromedriver_location)
             return browser, err_msg
 
+        # only work in chrome
```
converge: I did test, it also works in Firefox

converge (changes requested):
I see that there is a new parameter random_user_agent, what do you think about to make it less complex?

at the moment InstaPy is quite complex, we have been adding a lot of options, I think we should start to make it less flexible, it opens space for more bugs, and maybe no user will use this option

instapy/browser.py
breuerfelix: dang I shouldn't have done it quick at work :D thanks for mentioning :)

instapy/browser.py
breuerfelix: this is really strange because Joe also had errors here on raspberry pi ... I will test it in deep next days, so I put I back until then :)

breuerfelix:
fixed changes !

converge (approved):
🤘

PR #4715, opened 2019-07-26
Rewrite of stories

christophe-duc:
This is a better version of stories:
  - works in simulate = True or simulate = False, even if no video playback is available
  - fixed the user agent that was missing in the graphql queries

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -1236,7 +1236,7 @@ session.set_do_story(enabled = True, percentage = 70, simulate = True)
  If set to `True` InstaPy will simulate watching the stories (you won't see it in the browser), we just send commands to Instagram saying we have watched the stories.
  If set to `False` Instapy will perform the exact same action as a human user (clicking on stories, waiting until watching finishes, etc...)
 
-  Please note: `simulate = False` is the safest settings as it fully simulates all instagram interactions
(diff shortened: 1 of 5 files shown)


DOCUMENTATION.md
```diff
@@ -1236,8 +1236,7 @@ session.set_do_story(enabled = True, percentage = 70, simulate = True)
  If set to `True` InstaPy will simulate watching the stories (you won't see it in the browser), we just send commands to Instagram saying we have watched the stories.
  If set to `False` Instapy will perform the exact same action as a human user (clicking on stories, waiting until watching finishes, etc...)
 
-  Please note: on very small setups or if running on platform that cannot see videos, you will get stuck trying to watch the video. 
-  Recommended mode for those setups is `simulate=True`
+  Please note: `simulate = False` is the safest settings as it fully simulates all instagram interactions
```
timgrossmann: Does this make sense? 
Disabling the simulation should actually not simulate anything instead of `fully simulates all` 🤔

christophe-duc: yes, bad explanation, i've resubmitted a change

timgrossmann: `disables all additional, simulated interactions` would be best 😄 
Thanks

PR #4734, opened 2019-07-29
fix_xpath_likers_from_photo

christophe-duc:
Screen Shot 2019-07-29 at 19 15 28
fix xpath when there is now 2 links on a post: Liked by xxx and 17,234 others

we check first if the 2nd link is present (others) if not we fallback to the old way

### DIFF
--- a/instapy/commenters_util.py
+++ b/instapy/commenters_util.py
@@ -296,16 +296,22 @@ def likers_from_photo(browser, amount=20):
     """ Get the list of users from the 'Likes' dialog of a photo """
 
     liked_counter_button = "//div/article/div[2]/section[2]/div/div/a"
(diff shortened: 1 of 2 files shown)


instapy/commenters_util.py
```diff
@@ -296,16 +296,22 @@ def likers_from_photo(browser, amount=20):
     """ Get the list of users from the 'Likes' dialog of a photo """
 
     liked_counter_button = "//div/article/div[2]/section[2]/div/div/a"
+    second_counter_button = "//div/article/div[2]/section[2]/div/div/button"
```
breuerfelix: could you put that into the xpath compile file ? please ? :)

christophe-duc: sure, no problem. done. thanks!

PR #4803, opened 2019-08-08
Update login_util.py

nano2dev:
using raw_input raises less errors when using instapy programmatically.


converge (changes requested):
we're using python3, there is no raw_input.

PR #4814, opened 2019-08-10
Simple update browser.py and time_util.py for more randomness

jm-willy:
Instead of `time.sleep()`, use the random` sleep()` from `.time_util` for browser.py file
 and randomize `sleep_percentage` and `custom_percentage` from `.time_util`


converge:
what if we delete the custom sleep and start using?
```python
from time import sleep
from random import randint
sleep(randint(3, 6))
```

jm-willy:
@converge  I think `time_util.sleep` is more human-like than `time.sleep` because `time_util.sleep` gives values according to a Gaussian distribution/bell curve, which is very present in nature, including single human behavior and single human performance(such as click, scroll speed), also single machine performance. Maybe a Pareto distribution is more natural for performance of a large number of objects/humans(see wealth distribution for example), but probably not for single things. I believe this can help to fool Insta's AI to think InstaPy is a human user. Please correct my if got something wrong

breuerfelix:
didn't even know about this way of setting a random sleep delay ! really cool, will definitely check this out in detail soon !

Issue #4827, opened 2019-08-12
Bypass suspicious login failing on Dev branch


alxhoff:
@converge this is testing the Dev branch as of commit d5660dbfb8c8e7b58ebf9b81c51ad83cc9319130

## Expected Behavior
Logs in and runs
## Current Behavior
Security code is sent to the first option (email) from the list of two (email and mobile) and then instapy logs in, only to return to the log in page. At which time I usually get another security code via SMS and then instapy exits without any error messages.

## Possible Solution (optional)
Would be nice if the security code could be chosen to be sent from mobile as I am experiencing large delays in receiving the emails. For me the box is something like `//input[@id='choice_0']`

converge:
It would be good to have the option to choose how the security code should be sent. Unfortunately, I won't be able to do that, but it's an easy implementation.

alxhoff:
This is not the real problem though. The problem is that the browser returns to the login screen and instapy exits

PR #4842, opened 2019-08-13
Allows SMS to be set as the default security code authentication method

alxhoff:
As mentioned in  #4827


converge:
hey @alxhoff , I tried a new approach for this based on your solution, what're your thoughts? https://github.com/timgrossmann/InstaPy/pull/4850

converge:
thanks @alxhoff , I did use your PR as base for #4850, just merged it. I see that there is a new PR, let's keep the discussion.

Issue #1577, opened 2018-03-10
Follow loop error sample larger than population


cormo1990:
Hello fam,
during the follow user followers action, i found that error and the bor stopped immediatly.
can you help me please?

"ERROR - follow loop error sample larger than population"

thanks in advance

tanmald:
I too am having this error. But it only happens after following the followers of the first user on the list.

WARNING [2018-03-13 10:33:00] [xxx]  "user2" -> Less users to follow than requested.
ERROR [2018-03-13 10:33:00] [xxx]  follow loop error sample larger than population

ekumar96:
I am having the exact same issue as you tanmald, but it will not happen is terminal is active/in the front/foreground on my Desktop. Pretty perplexing.

britotiagos:
I got this error when using the smat_hashtag, what I did to overcome it was changing the 
sort='random' 
to
sort = 'top'

PR #4846, opened 2019-08-13
fix set_smart_hashtags with sort='random'

NotAdrian:
When using the functionality set_smart_hashtags with words which are not in English or do not generate more than 9 results for that one hashtag the program will crash with this error:

  File "playing_around_with_quota_supervisor.py", line 58, in <module>
    limit=10, sort='random', log_tags=True)
  File "/home/pi/.local/lib/python3.7/site-packages/instapy/instapy.py", line 703, in set_smart_hashtags
    random_tags = random.sample(data['results'], limit)
  File "/usr/lib/python3.7/random.py", line 321, in sample
    raise ValueError("Sample larger than population or is negative")
ValueError: Sample larger than population or is negative
`

The newly added secure handling will fix this issue for everyone who wants to use the set_smart_hashtags function with sort set to random and doesn't or can't work with the sort function set to 'top' which was suggested as a workaround in this issue: #1577

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -700,8 +700,10 @@ class InstaPy:
                         self.smart_hashtags.append(item["tag"])
 
                 elif sort == "random":
-                    if len(data['results']) < limit:
-                        random_tags = random.sample(data['results'], len(data['results']))
+                    if len(data["results"]) < limit:
+                        random_tags = random.sample(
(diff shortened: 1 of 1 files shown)


breuerfelix:
thank you for this small bugfix ! that never happened to me but its definitely a better solution than before :)

PR #4848, opened 2019-08-13
quickfix To Actually Ignore Top Posts when requested

NotAdrian:
Due to a design change at Instagram relating to their structure when searching for pictures by tags the "top_posts" arrays are not being filled anymore (in the whole program probably). Since they are simply returning 0 elements and the top post elements are now being automatically found through the normal image search there wasn't a big error crisis for the users but the like ratio should have gone down drastically for some people since the top post are now being counted as if they were actual posts which could potentially be liked. This would lead to people starting to raise their base values because there are 9 pictures per tag which are definitely not going to get liked.

The quick fix currently works like this: I'm raising the inserted base Amount by 9 to delete the first nine objects in the Array at the end of the function if skip_top_posts is set to true.

### DIFF
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -302,7 +302,7 @@ def get_links_for_tag(browser, tag, amount, skip_top_posts, randomize, media, lo
             )
             possible_posts = None
 
-    if skip_top_posts: 
+    if skip_top_posts:
         amount = amount + 9
 
     logger.info(
@@ -399,7 +399,7 @@ def get_links_for_tag(browser, tag, amount, skip_top_posts, randomize, media, lo
 
     sleep(4)
 
-    if skip_top_posts: 
+    if skip_top_posts:
         del links[0:9]
 
     if randomize is True:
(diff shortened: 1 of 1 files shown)


jm-willy:
I have checked it and the structure of the html have not changed, the current xpath: "//main/article/div[1]" highlights the first 9 pictures and "//main/article/div[2]" the rest, I have not checked the function though. What do you think is the problem?

christophe-duc (approved):

NotAdrian:
@jm-willy you are right I took another look and the paths are still the same. 
The problem was this line right here and the wrong usage of the here produced value in the function further down in the code:
`        possible_posts = (
            possible_posts if not skip_top_posts else possible_posts - len(top_posts)
        )`
possible_posts was set correctly with this but they only cut the array(/the available link List) shorter without skipping the first nine entries!!! Which lead to this malfunction of still checking the Top Posts.
I have removed this part of the program now because it wasn't integrated right in the ongoing code 
What I added in my first commit fixes this issue in a simple way and I haven't run into an error since I've been using it that way.

breuerfelix:
thanks for this fix ! cool that you figured out why this function didnt work :))

PR #4850, opened 2019-08-14
@Alxhoff: bypass suspicious attempt (by phone or by email)

converge:
hey @alxhoff, thanks for your PR, I made some changes and would like to ask you to try this approach, what do you think?


alxhoff (approved):
Works for me, problem solved :)

PR #4855, opened 2019-08-14
Proceeds to log in a second time when returned to login screen after security code validation, unsure of cause

alxhoff:
Solution to the problem I outlined in  #4827

Instagram returns to the login screen after successfully completing the first security code check


converge:
is this really necessary? this flow doesnt make much sense for me.

alxhoff:
I can't seem to get my script to not get looped back to the login page. No idea why it is happening to
 me. Otherwise I can't get the bot to work.

PR #4858, opened 2019-08-14
Fix user is inaccessible issue

converge:


SeRealMF:
Do i have to update it manually? Or just do ./update.sh in the quickstart folder?

PR #4873, opened 2019-08-15
remove follow and comments from pods

converge:


converge:
https://github.com/converge/instapy-pods-server -> working with restrictions (able to publish 5 posts per day).

notes;
- now we need to update InstaPy to use the new function.
- update server address

PR #4886, opened 2019-08-16
Update login_util.py

isrmor:
I updated to be able to use a proxy with IP and DNS, just adding socket.gethostbyname(proxy_address) call in line 192. 

If proxy_address is IP, it will return the IP, and if proxy_address is a domain, it will return the IP.

### DIFF
--- a/instapy/login_util.py
+++ b/instapy/login_util.py
@@ -189,7 +189,10 @@ def check_browser(browser, logfolder, logger, proxy_address):
         browser.get("view-source:https://api.myip.com/")
         pre = browser.find_element_by_tag_name("pre").text
         current_ip_info = json.loads(pre)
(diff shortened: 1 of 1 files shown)


breuerfelix:
so you are checking if the proxy is an actual IP address ? that also means that people cant use a proxy which is a domain right ?

isrmor:
I am checking that actual ip (using a proxy), is the same that is behind the domain name. To make sure that i am running via the domain proxy. If proxy is ip and not domain, it will do the same.

Before that change, i was trying to use a domain as a proxy and i cant use it.

breuerfelix:
ahh okay now I understand! thanks alot for this small change :) hope this is gonna help some people :)

PR #4892, opened 2019-08-17
Xpath, script and logical fixes of the new update

brnskn:
After changing of user agent(iphone x), instagram web page was changed. 

Fixed xpaths and logics that make follow_commenters function worked. 
Updated the checking of private user logic.
Fixed followers count xpath bug.
Updated the graphql script to get the ProfilePage information.

About the new commit:
I realised that, on some Instagram account, the old "window._sharedData.entry_data.ProfilePage[0]" variable is still working and the new one is not working. That's why I added a backup plan. It checks if "window.__additionalData" is defined, if not it tries to get info with "window._sharedData.entry_data.ProfilePage[0]".

### DIFF
--- a/instapy/commenters_util.py
+++ b/instapy/commenters_util.py
@@ -39,12 +39,14 @@ def remove_duplicates_preserving_order(seq):
     seen_add = seen.add
     return [x for x in seq if not (x in seen or seen_add(x))]
 
+
 def check_exists_by_tag_name(element, tag_name):
     try:
         element.find_element_by_tag_name(tag_name)
     except NoSuchElementException:
         return False
     return True
+
 
 def extract_post_info(browser):
     """Get the information from the current post"""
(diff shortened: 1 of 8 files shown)


instapy/commenters_util.py
```diff
@@ -19,6 +19,7 @@
 from .util import get_users_from_dialog
 from .util import progress_tracker
 from .util import close_dialog_box
+from .util import get_current_url
```
converge: Selenium already provide this function, it's called `browser.current_url`

instapy/commenters_util.py
```diff
@@ -39,74 +40,64 @@ def remove_duplicates_preserving_order(seq):
     return [x for x in seq if not (x in seen or seen_add(x))]
 
 
+def check_exists_by_tag_name(element, tag_name):
```
converge: I believe this is unnecessary, `.find_element_by_xpath` for example will already give us the results we need to (success or error), this would add more code for us to care about. At the moment we're trying to decrease complexity.

instapy/commenters_util.py
```diff
@@ -120,8 +111,8 @@ def extract_information(browser, username, daysold, max_pic):
         # we don't need to scroll more than is max number of posts we want
         # to extract
 
-    except Exception:
-        print("\nError: Couldn't get user profile. Moving on..")
+    except Exception as e:
+        print("\nError: Couldn't get user profile. Moving on.." + str(e))
```
converge, line 106: that's perfect, we're adding more info to the user.

converge (changes requested):
I believe these small changes will make it better and ready to push to `dev`.

instapy/commenters_util.py
brnskn: Also, we will be knowing more information about the error from the log.

breuerfelix:
@converge since you started this review, would be cool if you could resolve your changes :)) (when vacation time is over :P)

PR #4903, opened 2019-08-19
Improve testing

Jamim:
Hello,

I believe it would be nice to have tests for the upcoming Python versions.

These changes also:
 - add more precise language classifiers at `setup.py`
 - remove `tox-venv` since it causes failure on CI for Python 3.5
 - remove [deprecated](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration) `sudo` option from `.travis.yml`

Regards!


Jamim:
Hello @converge,
Can you please review/merge this PR?
Thank you!

tox.ini
```diff
@@ -26,5 +25,5 @@ skip_install = True
 [testenv:black]
 basepython = python3.7
 deps = black
-commands = black -t py33 --check .
+commands = black -t py37 --check .
```
breuerfelix: this needs to be 33 because we have to target 33 since we support versions which are lower than 3.7 :)

Jamim: Hello @breuerfelix,

As I see at `setup.py`, InstaPy [doesn't actually support](https://github.com/timgrossmann/InstaPy/blob/master/setup.py#L88) Python 3.3.
```python
    python_requires=">=3, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
```

Should I change the target to `py34`?

breuerfelix: actually thats true... I don't know why we don't have 3.3 :D but yeah change it to 3.4 :) that would be cool !

breuerfelix:
really cool changes ! I didnt know that this is possible :D always happy to learn new stuff :)

tox.ini
Jamim: It'd done.

Issue #4778, opened 2019-08-05
Installation fails due to conflicting jsonschema  version


NeolithEra:
Hi, users are unable to run **_InstaPy_** due to dependency conflict with _**jsonschema**_ package. 
As shown in the following full dependency graph of **_InstaPy_**, **_InstaPy_** requires _**jsonschema>=2.6.0**_，while **_clarifai 2.6.2_** requires _**jsonschema >=2.5,<3**_.

According to pip’s “first found wins” installation strategy, _**jsonschema 3.0.2**_ is the actually installed version. 
However, _**jsonschema 3.0.2**_ does not satisfy _**jsonschema >=2.5,<3**_.

### Dependency tree
```
InstaPy-legacy
| +-certifi(version range:>=2018.10.15)
| +-chardet(version range:>=3.0.4)
| +-clarifai(version range:>=2.4.1)
| | +-configparser(version range:>=3.5,<4)
| | +-future(version range:>=0.15,<2)
| | +-googleapis-common-protos(version range:>=1.5.0,<2)
| | +-grpcio(version range:>=1.13.0,<2)
| | | +-coverage(version range:>=4.0)
| | | +-cython(version range:>=0.29.8)
| | | +-enum34(version range:>=1.0.4)
| | | +-protobuf(version range:>=3.5.0.post1)
| | | +-six(version range:>=1.10)
| | | +-wheel(version range:>=0.29)
| | +-jsonschema(version range:>=2.5,<3)
| | +-protobuf(version range:>=3.6,<4)
| | +-requests(version range:>=2.13,<3)
| | | +-certifi(version range:>=2017.4.17)
| | | +-chardet(version range:<3.1.0,>=3.0.2)
| | | +-idna(version range:>=2.5,<2.9)
| | | +-urllib3(version range:<1.26,>=1.21.1)
| +-configparser(version range:>=3.5.0)
| +-easyprocess(version range:>=0.2.3)
| +-emoji(version range:>=0.5.1)
| +-future(version range:>=0.17.1)
| +-googleapis-common-protos(version range:>=1.5.5)
| +-grpcio(version range:>=1.16.1)
| | +-coverage(version range:>=4.0)
| | +-cython(version range:>=0.29.8)
| | +-enum34(version range:>=1.0.4)
| | +-protobuf(version range:>=3.5.0.post1)
| | +-six(version range:>=1.10)
| | +-wheel(version range:>=0.29)
| +-idna(version range:>=2.7)
[...]

PR #4932, opened 2019-08-23
Fix dependency conflict for issue

NeolithEra:
#4778


breuerfelix:
thanks for this small but needed fix !

PR #4971, opened 2019-08-30
Refactor and appium dev

christophe-duc:
This branch is intended to be a collaboration space for refactoring the code and be able to integrate appium webdriver into instapy.

If you would like to contribute to this effort, please contact me and I'll give you access to the branch on my fork, so you can commit.

Ideally we want to have well organized classes that can inherit when needed.


CharlesCCC:
uhhh interesting idea, so you goal is to migrate from web based automation to Android+ appium automation ?

christophe-duc:
> uhhh interesting idea, so you goal is to migrate from web based automation to Android+ appium automation ?

Ideally we would like to support both and have instapy capable of doing either selenium or appium depending on user choice.

christophe-duc:
The new structure is for the moment like this:

```
  +-- instapy
    + instapy.py (main class)
    +-- drivers
        + appium_webdriver.py
        + selemium_webdriver.py
        +-- appium_endpoints
            + appium_user.py
            + appium_post.py
            + appium_comment.py
            + xpath_compile.py
        +-- selenium_endpoints
            + selenium_user.py
            + selenium_post.py
            + selenium_comment.py
            + xpath_compile.py
    +-- common
        + logger.py
        + quota_management.py
        +-- model
            + user.py
            + post.py
            + comment.py
            + common.py
    +-- engine
        + legacy_engine.py
        + fsm_engine.py
```

the class inherits like this:
Common->User->AppiumUser

in the engine, we should only be working with a WebDriver (abstract class of AppiumWebDriver and SeleniumWebDriver), User, Post, Comment classes

CharlesCCC:
hi @christophe-duc , i just setup appium locally with your test script. The setup seems pretty straight forward , and the script is able to login to Instagram without problem. 

But right after login, it error out with the following:

```
  File "test_appium_android.py", line 58, in <module>
    profile[0].click()
IndexError: list index out of range
```

christophe-duc:
> hi @christophe-duc , i just setup appium locally with your test script. The setup seems pretty straight forward , and the script is able to login to Instagram without problem.
> 
> But right after login, it error out with the following:
> 
> ```
> Traceback (most recent call last):
>   File "test_appium_android.py", line 58, in <module>
>     profile[0].click()
> IndexError: list index out of range
> ```

The code is still in very heavy development at the moment.

CharlesCCC:
@christophe-duc Yep, I can see that, still great work tho, I believe this is moving to a really good direction. 

I was looking at your code. How do you plain to re-use the code/logic from InstaPy ? because the test script seems using totally different "element" for click and etc.

christophe-duc:
yes, the test script is for us to experiment with low level Appium functions, it does not reflect how we will in the future call the different functions.

There is an abstraction layer which we are trying to code at the moment.

my above post describing the new structure (in particular the legacy_engine class) will have the main role of mapping the current instapy functions and calling the abstracted classes (User, Post, Comment)

Are you interested in contributing? If so, join us on discord, and post on the #dev channel.

CharlesCCC:
I'm more than happy to help out, I'm on the InstaPy discord , @christophe-duc is that you "chrisrohitbrendan" ?

timgrossmann:
@christophe-duc Thanks for all the work you've put into this.
Sad to see it not being delivered to the community. Hopefully the new tool will also attract this many open-source enthusiasts.

All the best 😊

breuerfelix:
btw @christophe-duc since you basically have rewritten the whole lib, why dont you also create a new lib / packages and decouple it from instapy ? would like to see how this works compared to my approach in the long term !

Issue #5018, opened 2019-09-06
Follow by tag failing, with "Unavailable Page" error


Lusitaniae:
## Expected Behavior
Follow by tag is currently failing

## Current Behavior
All profiles being checked by the follow by tag return the same error, seen below:

```
WARNING [2019-09-06 10:25:16] [username]  Unavailable Page: b'https://www.instagram.com/p/B2DfiG3BBwf/'
INFO [2019-09-06 10:25:16] [username]  --> User not followed: Unavailable Page
```
## InstaPy configuration
I'm currently running latest commit on `dev`

Issue seems to have been introduced in 
https://github.com/timgrossmann/InstaPy/commit/c0d5f6c0dee312b31e8e5b2c3eac5fee3a2e039e

Rolling back to 
https://github.com/timgrossmann/InstaPy/commit/5b150da4026eb752dc841f7e722b57e1a6e44a29

Seems to solve it

Lusitaniae:
Linked PR solved it!

Issue #5017, opened 2019-09-05
"Couldnt get if user is Private"


GreatMalboro:
## Expected Behavior

To work after I downloaded the newest update. Still no rush here really tho.

## Current Behavior

Only pulling Private accounts.

ERROR [2019-09-05 15:28:08] [greatmalboro]  ~cannot get if user is private
INFO [2019-09-05 15:28:08] [greatmalboro]  ---> Sorry, couldn't get if user is private

## InstaPy configuration

I happen to just throw the geckodriver into the same place that is holding the quickstart.

Issue #5013, opened 2019-09-05
WARNING [2019-09-05 18:07:23] [My.Insta]  Unavailable Page: b'https://www.instagram.com/p/B2CRf3CgQnI/'


Tr1pke:
## Expected Behavior
working like it worked before
## Current Behavior
WARNING [2019-09-05 18:06:49] [My.Insta.Account]  Unavailable Page: b'https://www.instagram.com/p/B2CRsf5grJl/'
INFO [2019-09-05 18:06:49] [My.Insta.Account]  --> Image not liked: b'Unavailable Page'
INFO [2019-09-05 18:06:49] [My.Insta.Account]  Like# [8/59]
INFO [2019-09-05 18:06:49] [My.Insta.Account]  https://www.instagram.com/p/B2CRrdEIwYB/
WARNING [2019-09-05 18:06:52] [My.Insta.Account]  Unavailable Page: b'https://www.instagram.com/p/B2CRrdEIwYB/'
INFO [2019-09-05 18:06:52] [My.Insta.Account]  --> Image not liked: b'Unavailable Page'
INFO [2019-09-05 18:06:52] [My.Insta.Account]  Like# [9/59]
INFO [2019-09-05 18:06:52] [My.Insta.Account]  https://www.instagram.com/p/B2CRppVgvQ0/
WARNING [2019-09-05 18:06:56] [My.Insta.Account]  Unavailable Page: b'https://www.instagram.com/p/B2CRppVgvQ0/'
INFO [2019-09-05 18:06:56] [My.Insta.Account]  --> Image not liked: b'Unavailable Page'
INFO [2019-09-05 18:06:56] [My.Insta.Account]  Like# [10/59]
INFO [2019-09-05 18:06:56] [My.Insta.Account]  https://www.instagram.com/p/B2CRnHIgEd-/
WARNING [2019-09-05 18:07:00] [My.Insta.Account]  Unavailable Page: b'https://www.instagram.com/p/B2CRnHIgEd-/'
INFO [2019-09-05 18:07:00] [My.Insta.Account]  --> Image not liked: b'Unavailable Page'
INFO [2019-09-05 18:07:00] [My.Insta.Account]  Like# [11/59]
INFO [2019-09-05 18:07:00] [My.Insta.Account]  https://www.instagram.com/p/B2CRleMItjQ/
[...]

Issue #5006, opened 2019-09-04
: The element reference of <article class="FyNDV"> is stale; either the element is no longer attached to the DOM, it is not in the current frame context, or the document has been refreshed


superryeti:
## Expected Behavior
interact_user_followers was supposed to get links from the users.
## Current Behavior

```
  File "instabot.py", line 320, in <module>
    competitors, amount=amount, randomize=True)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/instapy/instapy.py", line 3185, in interact_user_followers
    self.user_interact_media,
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/instapy/instapy.py", line 2471, in interact_by_users
    media,
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/instapy/like_util.py", line 490, in get_links_for_username
    links = links + get_links(browser, person, logger, media, main_elem)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/instapy/like_util.py", line 829, in get_links
    link_elems = element.find_elements_by_xpath('//a[starts-with(@href, "/p/")]')
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/selenium-3.141.0-py3.7.egg/selenium/webdriver/remote/webelement.py", line 381, in find_elements_by_xpath
    return self.find_elements(by=By.XPATH, value=xpath)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/selenium-3.141.0-py3.7.egg/selenium/webdriver/remote/webelement.py", line 685, in find_elements
    {"using": by, "value": value})['value']
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/selenium-3.141.0-py3.7.egg/selenium/webdriver/remote/webelement.py", line 633, in _execute
    return self._parent.execute(command, params)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/selenium-3.141.0-py3.7.egg/selenium/webdriver/remote/webdriver.py", line 321, in execute
    self.error_handler.check_response(response)
[...]

PR #5028, opened 2019-09-07
Bug fixes about StaleElementReferenceException and graphql scripts

brnskn:
Fixes: #5018 #5017 #5013 #5006


breuerfelix:
awesome work mate ! I hope they will end up in master soon !

PR #5044, opened 2019-09-10
Fix error `TypeError: '<' not supported between instances of 'int' and 'NoneType'

Immelstorn:
Error 
TypeError: '<' not supported between instances of 'int' and 'NoneType'

is raised if set     
min_likes=None in set_delimit_liking


christophe-duc:
this is also fixed by #5046

breuerfelix:
thanks for this small fix :)

PR #5047, opened 2019-09-11
Outdated variable name

hmellor:
When min and max where changed to minimum and maximum about a month ago this one was left behind.


hmellor:
Turns out this was a duplicate

PR #5060, opened 2019-09-12
Add docstring and small refactor for interact_user_followers()

aalvrz:
First PR here :tada:

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -3,7 +3,6 @@
 import time
 from datetime import datetime, timedelta
 from math import ceil
-from typing import List
 import random
 from sys import platform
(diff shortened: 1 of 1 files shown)


breuerfelix:
Hello BigChief !

really cool :) that's a kind of PR we need more from ! it will make the codebase more readable for other people :)

the problem is, we currently only support the normal list type and not the advanced one in the Gui. I will try to fix this asap but until then, I will convert the type back to the normal list one :)

aalvrz:
@breuerfelix Thanks for the explanation! :+1: 

Please do let me know (mention) when support for `typing` has been added in the GUI.

Issue #4784, opened 2019-08-06
AttributeError: 'list' object has no attribute 'lower'


fgisslen:
I get this error:
Anyone know what's wrong?

"AttributeError: 'list' object has no attribute 'lower'"

Traceback (most recent call last):
  File "/home/x/InstaPy/bellman2.py", line 119, in <module>
    session.interact_user_followers(random_targets, amount=random.randint(10, 20), randomize=True)
  File "/home/x/InstaPy/instapy/instapy.py", line 2940, in interact_user_followers
    validation, details = self.validate_user_call(person)
  File "/home/x/InstaPy/instapy/instapy.py", line 1248, in validate_user_call 
    self.logfolder)
  File "/home/x/InstaPy/instapy/util.py", line 382, in validate_username
    if bio_keyword.lower() in profile_bio.lower():
AttributeError: 'list' object has no attribute 'lower'

fgisslen:
@uluQulu

uluQulu:
Hey @fgisslen 

It is a bug in **util.py**:

![image](https://user-images.githubusercontent.com/33269249/62539248-d702bf80-b865-11e9-9b79-8cf453e3e6be.png)

1. Add a new line:
```python
profile_bio = getUserData("graphql.user.biography", browser)
profile_bio = [pb.lower() for pb in profile_bio]   # ADD THIS NEW LINE
```
2. Change this line:
```python
if bio_keyword.lower() in profile_bio.lower():
```
becomes
```python
if bio_keyword.lower() in profile_bio:
```

Chill 😎

fgisslen:
@uluQulu thx m8

fgisslen:
@uluQulu actually I still get after editing:


Traceback (most recent call last):
  File "x.py", line 119, in <module>
    session.interact_user_followers(random_targets, amount=random.randint(10, 20), randomize=True)
  File "/home/x/InstaPy/instapy/instapy.py", line 2940, in interact_user_followers
    validation, details = self.validate_user_call(person)
  File "/home/x/InstaPy/instapy/instapy.py", line 1248, in validate_user_call
    self.logfolder)
  File "/home/x/InstaPy/instapy/util.py", line 383, in validate_username
    if bio_keyword.lower() in profile_bio:
AttributeError: 'list' object has no attribute 'lower'

rudyvanbebu:
Thanks all for revising this, tested after editing. Exiting script / session with the following traceback log

Traceback (most recent call last):
[...]

PR #5066, opened 2019-09-13
InstapyFollowed and nonFollowers parameters, and issue 4784

FabrizioConrado:
Managing InstapyFollowed and nonFollowers parameters = True in unfollow()

When passing both parameters InstapyFollowed and nonFollowers as True to unfollow(), the second is ignored as the condition InstapyFollowed comes first. The suggestion below would ignore InstapyFollowed and base the following actions on nonFollowers (which is itself based on the InstapyFollowed list minus users that followed back at any time).

Add in unfollow_util.py, line 295:

if nonFollowers is True: InstapyFollowed = False # nonFollowers is already based on InstapyFollowed minus Followers

The second one is to swap the order of **self.skip_bio_keyword** and  **self.dont_skip_business_categories** on validate_user_call() as pointed out by @uluQulu here: https://github.com/timgrossmann/InstaPy/issues/4784

### DIFF
--- a/instapy/unfollow_util.py
+++ b/instapy/unfollow_util.py
@@ -292,6 +292,8 @@ def unfollow(
 
     if customList is True or InstapyFollowed is True or nonFollowers is True:
 
+        if nonFollowers is True: InstapyFollowed = False  # nonFollowers is already based on InstapyFollowed minus Followers
+
         if customList is True:
             logger.info("Unfollowing from the list of pre-defined usernames\n")
             unfollow_list = customList_data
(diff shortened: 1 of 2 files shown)


instapy/unfollow_util.py
```diff
@@ -292,6 +292,8 @@ def unfollow(
 
     if customList is True or InstapyFollowed is True or nonFollowers is True:
 
+        if nonFollowers is True: InstapyFollowed = False  # nonFollowers is already based on InstapyFollowed minus Followers
```
aalvrz, line 295: It is preferable that this is separated in two lines.

PR #5069, opened 2019-09-13
Refactor smart_run exceptions handling

aalvrz:
Is there any reason for using `isinstance` to check for which exception? If so please do let me know.


breuerfelix:
nope you did it the proper way !
thanks :)

PR #5071, opened 2019-09-13
Added load_following_data

meinert:
Added load_following_data method to load local stored following data

Similar to load_followers_data

### DIFF
--- a/instapy/relationship_tools.py
+++ b/instapy/relationship_tools.py
@@ -1255,3 +1255,159 @@ def load_followers_data(username, compare_by, compare_track, logger, logfolder):
 
     # return that file to be compared
     return followers_data, selected_filename
+
+def load_following_data(username, compare_by, compare_track, logger, logfolder):
+    """ Write grabbed `following` data into local storage """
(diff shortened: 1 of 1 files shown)


breuerfelix:
you need to run the black formatter over this code :)

btw this is a cool feature but right now its code that is never ever used anywhere.

where would you use this function ? cause adding functions that are never called is really confusing :D at least if they are not on the instapy class :)

meinert:
Could you please point me the direction of the "black formatter"?

I am using the method in my run scripts. To save server_calls I try to load following_data from local storage, if it has been saved current month.

breuerfelix:
black formatter means you need to install the formatter with pip3 install black
and then run it on the codebase with `black -t 34 .` so it will fix all Linting errors :)

well if you use it in your run scripts could you please add this function to the API in the instapy class, just a function in the instapy class which calls this function (and with python typings on the instapy function) and also put some lines in the readme so people know how to use it ? cause otherwise this function is just tangling around in the codebase and no one knows that its there :)

PR #5072, opened 2019-09-13
Update unfollow()  and  validate_user_call()

FabrizioConrado:
unfollow() - Manage **InstapyFollowers** and **nonFollowers** parameters when both are True

validate_user_call() - swap order of **self.skip_bio_keyword**  with   **self.dont_skip_business_categories**


breuerfelix:
thanks again !! :)

PR #5077, opened 2019-09-15
IMAGE SIZE IMPROVEMENTS AND ALIGNMENT

pathakmayank:
Have improved the major INSTAGRAM image size with some alignments at the end and have also improved UserId and Password for better understanding of the user.


breuerfelix:
thanks for these beautify fixes :)

PR #5078, opened 2019-09-15
Random seed generator

jm-willy:
I included this function in `util.py` to get random and unique seeds for each user.

```
path_str = sys.path[0] + os.get_exec_path()[0] + sys.path[-1] + os.get_exec_path()[-1] + str(os.getpid())


def new_seed():
    random.seed(str(random.randint(0, 1_000_000)) + str(os.times()) + str(datetime.datetime.now()) + path_str)
```

`new_seed()` function is called from the three most commonly used instapy functions each time those functions themselves are called: `sleep()`, `update_activity()` and `read_xpath()`, to renew the seed constantly

### DIFF
--- a/instapy/time_util.py
+++ b/instapy/time_util.py
@@ -6,16 +6,15 @@ from random import uniform
 from random import randint
 from random import seed
 from sys import path
-from os import get_exec_path()
-from os import getpid()
-from os import times()
+from os import get_exec_path
+from os import getpid
+from os import times
 
 # Amount of variance to be introduced
 # i.e. random time will be in the range: TIME +/- STDEV %
 STDEV = 0.5
 sleep_percentage = 1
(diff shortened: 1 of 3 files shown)


breuerfelix:
could you run the black formatter over the code please ? :) `black -t py34 .` and commit the changes :)

thanks for this PR, I hope this seeding won't take too much cpu power :D:D

jm-willy:
@breuerfelix Modified it, I think is lighter on the processor now but equally effective, I'm new and don't know what's black but this is PEP-8 compliant and python 3.7

christophe-duc:
> @breuerfelix Modified it, I think is lighter on the processor now but equally effective, I'm new and don't know what's black but this is PEP-8 compliant and python 3.7

black is just a formatter, you can install it with `pip install black` and run it on the file you modified and it will reformat everything according to the standard used in instapy.

thanks!

PR #5148, opened 2019-10-14
Update for util.py

riccardovalente:
Fix for error at startup when `want_check_browser` was introduced


breuerfelix:
thanks alot ! :)

PR #5443, opened 2020-04-28
Insert target users, hashtags etc from text file - Issue #5396

IreneArapogiorgi:
Hello, I was assigned to implement the ability to parse text files containing target users, hashtags, comments etc in order to use it with functions which accept lists as given parameters. This will be a part of a university assignment of mine as I also mentioned on issue #5396.

**Functions added:**
file_handling in util.py
target_list in instapy.py

_PS: At first, I implemented file_handling function inside instapy.py (with only a few changes) & that also worked, but I had to change that implementation because I wasn't able to do some unit testing._

**Unit testing:**
test_file_handling.py in tests folder
invalid_target_list.txt in tests/resources folder
valid_target_list.txt in tests/resources folder

**Example of using new feature in quickstart.py**
```
""" Quickstart script for InstaPy usage """

# imports
from instapy import InstaPy
from instapy import smart_run
from instapy import set_workspace

# set workspace folder at desired location (default is at your home folder)
set_workspace(path=None)

# login credentials
insta_username = "your instagram username"
insta_password = "your instagram password"

# get an InstaPy session
session = InstaPy(username=insta_username,
                  password=insta_password,
                  headless_browser=False)

with smart_run(session):

    # like posts based on hashtags
    hashtags = session.target_list("C:\\Users\\..........\\hashtags.txt")
    session.like_by_tags(hashtags, amount=9)
```
If there are any changes/improvements needed, please let me know so that I can work on them.

I will be thrilled if you accept my contribution in the end, as this is also a grading factor for my assignment.

Issue #5396, opened 2020-04-06
Insert target users, hashtags etc from text file


IreneArapogiorgi:
Hello, I would like to contribute to InstaPy as part of a university course of mine, called Software Engineering in Practice, in which we are supposed to contribute to an open source project of our choice.

### Expected Behavior
Amongst other feature ideas, I thought it would be useful for most users to have the ability to import a text file with their target users, hashtags etc. This implementation will especially benefit businesses using InstaPy as they could store users/hashtags etc in already-made target lists as text files and parse each one of them depending on their campaigns.
### Current Behavior
For all existing functions accepting lists as given parameters, simple user is obliged to edit those lists inside the script, which complicates code comprehension & requires time.
### Possible Solution
Implement the option of reading a text file for each function which accepts lists as given parameters.

_I will be thrilled if you accept my upcoming contribution & I hope we'll have a great collaboration._
_PS: Any advice is more than welcomed._

Miladinho:
This software is heavily dependent on the changes Facebook makes to Instagram and therefore is not itself something that would be that useful for you to speak about as a good project reference when you go into job interviews. Furthermore, I think it depends heavily on the use of selenium, so unless you want to go into a role like SDET or QA engineering, there is not much framework specific knowledge to be gained. I recommend using your time for this school project to work on something that can develop specific skills you want to use in a job that you desire.

IreneArapogiorgi:
@Miladinho I really appreciate your advice. Thank you for responding so fast.

As this is going to be my first contribution to an open source project, I chose InstaPy for being well-written in Python (in which I'm already experienced) and also because I have personally been involved in social media marketing before (and especially Instagram campaigns) and therefore I'm quite familiar with InstaPy's theme and functionality. My goal is actually to manage comprehending other people's code, follow up with their standards and try to make even a small contribution. And then I shall chase bigger projects.

I'm keeping your advice and I already have a list of alternatives, however I would like to give this project a chance.

breuerfelix:
> ... I chose InstaPy for being well-written in Python ...

i don't know if i can sign this :D in germany we say 'Too many cooks spoil the broth.'. But yeah i guess that's the same for every big OS project :)

We already talked about your project. There is just one tip i got for you. Don't change the existing functions. Just program new functions which take a path to a list as a param, parse the list and pass the parsed list to the already existing function :) you can then also pass all other params to the existing function with `*args` and `**kwargs`.  Just some suggestions that comes to my mind!

[...]

PR #5490, opened 2020-05-13
Fix text file handling for comments

IreneArapogiorgi:
Hello, I found out a mistake I made while implementing the `file_handling` function mentioned on issue #5396 and submitted on pull request #5443. 

I was removing all spaces found before, after and in-between target words and also transforming all letters to lowercase, which would transform a possible comment from `This is a comment` to `thisisacomment`.

I fixed that and updated the unit tests included.

### DIFF
--- a/instapy/util.py
+++ b/instapy/util.py
@@ -2408,9 +2408,8 @@ def file_handling(file):
             # extract file's lines in list
             for line in f.readlines():
                 if line != "\n":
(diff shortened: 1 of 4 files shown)


breuerfelix:
Sorry for the delay! Good that you are reminding me at Discord :)

Thanks for the fix. Merged !

breuerfelix (approved):

PR #5591, opened 2020-06-25
Update URL

specdrake:
The URL `https://www.instagram.com/p` should be replaced by `https://www.instagram.com/`

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -4705,8 +4705,8 @@ class InstaPy:
                 # cos this method has just one iterator
                 break
 
-            if "https://www.instagram.com/p/" not in url:
-                url = "https://www.instagram.com/p/" + url
+            if "https://www.instagram.com/" not in url:
+                url = "https://www.instagram.com/" + url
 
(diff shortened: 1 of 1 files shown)


timgrossmann:
@specdrake `/p/` is necessary when viewing a post, why does it have to be removed here?

specdrake:
I was talking in the page context as Instagram pages don't have `/p`. I think there is something wrong with the `check_link` function in `like_util.py`. Please take a look there.

timgrossmann:
@specdrake Looks like they might start to remove that `/p/` section of the URL. However when I open a post on Instagram I see the `/p/` path in the URL

specdrake:
Yeah, posts do have a `/p` section in the URL. My use case is : "Interacting on a given profile" and it's URL doesn't contain the `/p` part, although removing it doesn't fix the errors. I think Instagram must have changed something in the "Window" object as I can't find the `data` property at the given path. So, the `check_link` function may have become outdated. I am not experienced with JavaScript, so I couldn't change the DOM manipulation for troubleshooting.

PR #5638, opened 2020-07-17
Fixing some bugs

Mohanbarman:
Fixing invalid like element issue and crashing when trying to comment on disable comment post.

### DIFF
--- a/instapy/comment_util.py
+++ b/instapy/comment_util.py
@@ -84,11 +84,11 @@ def comment_image(browser, username, comments, blacklist, logger, logfolder):
             )
             return False, "commenting disabled"
 
-        if not comment_elem:
-            logger.warning(
-                "--> Comment Action Likely Failed!" "\t~comment Element was not found"
-            )
-            return False, "commenting disabled"
+    if not comment_elem:
+        logger.warning(
+            "--> Comment Action Likely Failed!" "\t~comment Element was not found"
+        )
+        return False, "commenting disabled"
 
     # wait, to avoid crash
     sleep(3)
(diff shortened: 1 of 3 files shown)


timgrossmann:
@Mohanbarman Thank you for this PR

Could you please rebase this PR on the new master branch to update it?

Also, can we abstract the xPaths even more? 
Only using `//*[@aria-label='Comment']` etc.?

LeydenJar:
Thanks for the fixes, that works!!

I was trying to replace the xpaths with the ones you can get with the inspectors in chrome and firefox, but none of them worked, wich I think is weird because they've always done the job before when I've used selenium. Can you guys tell me why this happens?

timgrossmann (approved):

timgrossmann:
@LeydenJar Could you please check this PR again? Thanks 😊

instapy/comment_util.py
```diff
@@ -68,7 +75,21 @@ def comment_image(browser, username, comments, blacklist, logger, logfolder):
...
+    comment_elem = open_comment_section(browser, logger)
+
+    if comment_elem:
+        if type(comment_elem) == tuple and not comment_elem[0]:
+            logger.warning(
+                "--> Comment Action Likely Failed!" "\t~comment Element was not found"
+            )
+            return False, "commenting disabled"
+
+        if not comment_elem:
```
breuerfelix: that does not really make sense here. you first check for if comment_elem and in the if clause you check for if not comment_elem... this will NEVER happen :D

breuerfelix (dismissed):

sskender:
This seems to be working. Will it be merged?

timgrossmann:
@sskender Can you please test again?

garyfu5353:
Hi Tim 
And thanks for years of Bot thanks to InstaPy 

From today the Bot can write a comment but can't publish it like if it can't press the button..

Have you an issue ?

PR #5659, opened 2020-07-25
fix smart location url

timgrossmann:


timgrossmann:
@oliyan Can you please second this update?

PR #5684, opened 2020-08-02
Fix for obtaining user id.

Jwdev-wr:
This will also save on server calls as the last one would failed and reload the page everytime.


timgrossmann:
@Jwdev-wr Good spot that this was no working anymore 👍 
I've adjusted the graphql path to the actual user_id in a simpler format.

Could you check back if it works correctly, please? 😊

timgrossmann (approved):

PR #5709, opened 2020-08-07
Bio keyword bug fix

owoki:
When skip words were added, there was an obligation to add the mandatory words. I solved this by adding the if.

### DIFF
--- a/instapy/instapy.py
+++ b/instapy/instapy.py
@@ -5840,3 +5840,34 @@ class InstaPy:
             return []
 
         return target_list
+    def send_message(self, usernames: list, message):
+        self.selectors = {
+                "home_to_login_button": ".WquS1 a",
+                "username_field": "username",
+                "password_field": "password",
+                "button_login": "._0mzm-",
+                "search_user": "queryBox",
+                "select_user": "._0mzm-",
+                "textarea": "textarea",
+                "send": "button"
+                }
+        if not isinstance(usernames, list):
+            usernames = [usernames]
+        
+        for username in usernames:
+            self.logger.info('Send message {} to {}'.format(message, username))
+            self.browser.get('https://www.instagram.com/direct/new/')
+            self.browser.find_element_by_name(self.selectors['search_user']).send_keys(username)
+            sleep(1)    
+            # Select user
+            elements = self.browser.find_element_by_xpath("//button[contains(@class,'dCJp8')]").click()
+            sleep(0.5)
+            self.browser.find_element_by_xpath("//div[@class='rIacr']").click()
+            
+            # Go to page
+            
+            sleep(1)
+            self.browser.find_elements_by_xpath("*//textarea")[0].send_keys(message)    
+            sleep(2)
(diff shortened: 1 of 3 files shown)


instapy/util.py
```diff
@@ -435,24 +435,26 @@ def validate_username(
...
-            return False, "---> Sorry, couldn't get get user bio " "account active\n"
-        for bio_keyword in skip_bio_keyword:
-            if bio_keyword.lower() in profile_bio:
-                return (
-                    False,
-                    "{} has a bio keyword of {}, by default skip\n".format(
-                        username, bio_keyword
-                    ),
-                )
+        if len(skip_bio_keyword) > 0:
```
timgrossmann, line 438: This is already checked in the line above with:

```python
if len(skip_bio_keyword) > 0 or len(mandatory_bio_keywords) > 0:
```

instapy/xpath_compile.py
```diff
@@ -185,6 +185,6 @@
 }
 
 xpath["likers_from_photo"] = {
-    "liked_counter_button": "//div/article/div[2]/section[2]/div/div/a",
+    "liked_counter_button": "//a[contains(@href,'liked_by/')]",
```
timgrossmann, line 188: Does not seem to be working when tested in Chrome

instapy/instapy.py
```diff
@@ -5840,3 +5840,34 @@ def target_list(self, file):
             return []
 
         return target_list
+    def send_message(self, usernames: list, message):
+        self.selectors = {
```
timgrossmann, line 5844: We should move that to the xPaths file that holds all the references to the elements 😊

instapy/util.py
owoki: When you write the words you don't want, the value of len is greater than 0. This if works because it is an OR connector. But mandatory words pass the profile even if the empty array comes. This piece of code checks this. The person may want to use "skip" and not use "mandatory" one.

instapy/xpath_compile.py
owoki: It didn't work in Firefox, so it was added and worked.

instapy/instapy.py
owoki: thx but "sleep" not defined so maybe instagram gonna block in multiple dms

instapy/util.py
marioguima: > But mandatory words pass the profile even if the empty array comes.

it happened to me when using the mandatory_bio_keywords list

the solution for recursive code is to use the python all function

Ps. I don't know how to share my code here on git

but the solution was like this
---------------------------------------------

def evaluate_mandatory_words(text, mandatory_words_list):
    return all(x.lower() in text.lower() for x in mandatory_words_list)

---------------------------------------------
or without calling the function
in line 452

if not all(x.lower() in (username + " " + profile_bio).lower() for x in mandatory_bio_keywords):
            return False, "Mandatory bio keywords not found"

Issue #5640, opened 2020-07-17
Follow likers - Couldn't find liked counter button


alinakhay:
## Expected Behavior

Expected to load the users who liked the post in order to follow them 
## Current Behavior

Getting the "Couldn't find liked counter button. May be a video. Moving on.." message on posts that are definitely not a video and have over 100 likes.

## Possible Solution (optional)

I presume instagram changed the code so that the bot can't really find the part of the page where it supposed to get likers from.

## InstaPy configuration

```
import random
from instapy import InstaPy
from instapy import smart_run

session = InstaPy(username='', password='', headless_browser=False) --or True does not really matter - same behaviour

with smart_run(session):
    
    my_users = ['fashionage', 'lifestylecatcher', 'i__am_fashion']
    
    session.set_do_follow(enabled=False, percentage=30, times=1) #True
    session.set_do_comment(enabled=True, percentage=95)
    session.set_comments([  'Nice shot! @{}', 
                            'I love your profile! @{}', 'etc...'
                             ],
                         media='Photo')
                         
    session.set_do_like(True, percentage=80)
    session.set_delimit_liking(enabled=True, max_likes=70, min_likes=0)

    session.set_relationship_bounds(enabled=True,
                                    potency_ratio=None,
                                    delimit_by_numbers=True,
                                    min_followers=10,
                                    min_following=90)


    session.set_user_interact(amount=1, percentage=80,randomize=True, media='Photo')
    session.follow_likers(my_users, photos_grab_amount = 2, follow_likers_per_photo = 11, randomize=True, sleep_delay=100, interact=True)

    session.join_pods()
```

andrenpt:
I am having the exact same problem.

avacaondata:
Same issue here!

andrenpt:
Fix for this:  

`"liked_counter_button": "//div/article/div[3]/section[2]/div/div[2]/a",`
💯

tanmald:
> Fix for this:
> 
> `"liked_counter_button": "//div/article/div[3]/section[2]/div/div[2]/a",`
> 💯

didn't work for me.

mkhazan:
Same issue, even after updating to 6.10.

szabizs:
The problem seems to come from instagram's resolution, if you do not resize the window, it will work, since elements will change based on screen resolution.

Also I've updated my xpath

```
xpath["likers_from_photo"] = {
[...]

PR #5744, opened 2020-08-20
Follow likers - Couldn't find liked counter button #5640

jordanstarrk:
Updated correct xpath for "likers_from_photo" used in session.follow_likers


ppxscal:
I copied the fixes but the program returns an empty list for the like usernames for a photo:

Few likes, not guaranteed you don't follow these likers already.
Got photo likers: ['']

Tried tracing the bug to commenters_util.py, but I can't figure out why it's not working.

Issue #5803, opened 2020-09-26
Keyword randomize doesn't work


WetkinTeodohr:
This is the line of code that gives me an error. I didn't put the location for privacy purposes. 
session.like_by_locations(['????'], amount=100, randomize=True)

#5050  Expected Behavior
I want it to randomize the posts it takes a look at.

## Current Behavior
It gives me back this error.
TypeError: like_by_locations() got an unexpected keyword argument 'randomize'

 I've been searching for why it does this but i can't come up with anything because everyone just uses the 'randomize' like this. 

Help would be much appreciated.

elulcao:
> This is the line of code that gives me an error. I didn't put the location for privacy purposes.
> session.like_by_locations(['????'], amount=100, randomize=True)
> 
> #5050 Expected Behavior
> I want it to randomize the posts it takes a look at.
> 
> ## Current Behavior
> It gives me back this error.
> TypeError: like_by_locations() got an unexpected keyword argument 'randomize'
> 
> I've been searching for why it does this but i can't come up with anything because everyone just uses the 'randomize' like this.
> 
> Help would be much appreciated.

def like_by_locations() doesn't have randomize flag, I'm reviewing if a code change is needed
thanks

```python
    def like_by_locations(
        self,
        locations: list = None,
        amount: int = 50,
        media: str = None,
        skip_top_posts: bool = True,
    ):
```

PR #5809, opened 2020-10-01
Accept follow requests assert

elulcao:
Problem:
- `accept_follow_requests()` asserts if user requested to accept more than 10 followers. This is because an xpath cannot be reached, lets said they have 11 follower but only 10 are visible since User requested to accept 20  the 11 follower will be tried to be accepted but before that an assert will be thrown.
- Adde to `like_by_locations()` the `randomize` flag, #5803 
- There are some messages for referenced variables before assignment.

Solution: 
- Call `scroll_down()` after follower accepted, with this new def the window is scrolled down after accepting a follower.
- Added `like_by_locations()` with `randomize` flag
nit changes:
- Sorted imports, to make code readable
- Defined local variables ( ie: 'unfollow_state') to fix message for referenced before assignment
- Added *.code-workspace in .gitignore for vscode project (macOS) 
 

Note:
```python
(venv) ➜  InstaPy git:(accept_follow_requests_assert) pycodestyle --max-line-lengt=200 instapy
instapy/util.py:1772:9: E731 do not assign a lambda expression, use a def
instapy/util.py:2256:50: E203 whitespace before ':'
(venv) ➜  InstaPy git:(accept_follow_requests_assert)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,6 +105,7 @@ logs/
 # Editors
 .idea
 .vscode
+*.code-workspace
 .bak
 
 # pytest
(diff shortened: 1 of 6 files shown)


timgrossmann (approved):

timgrossmann:
@elulcao Awesome PR, Daniel.
Thank you very much for this!

PR #5843, opened 2020-10-23
Add method to check if selected user is verified (verified badge)

shifenis:
Add a simple method that returns `true` or `false` depending on whether the user is verified or not.

Simple usage:

```python
from instapy import relationship_tools

session = InstaPy(username="username",password="psw")
relationship_tools.is_user_verified(session.browser, username="username1", session.logger)
```

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -78,6 +78,7 @@
  <br />
 
 - **[Relationship tools](#relationship-tools)**
+  - [Is selected user is verified](#is-user-verified)
   - [Grab Followers of a user](#grab-followers-of-a-user)
   - [Grab Following of a user](#grab-following-of-a-user)
   - [Pick Unfollowers of a user](#pick-unfollowers-of-a-user)
@@ -1906,6 +1907,16 @@ session = InstaPy(username=insta_username,
 <br />
 
 ## Relationship tools
+
+### Is user verified
+###### Returns `true` or `false` depending if the the user is verified or not
+```python
+isUserVerified = session.isUserVerified(username="username1")
+```
+### Parameters:
+`username`: A desired username to grab its followers
+
+
 ### Grab Followers of a user
 ###### Gets and returns `followers` of the given user in desired amount, also can save locally
 ```python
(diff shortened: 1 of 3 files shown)


instapy/relationship_tools.py
```diff
@@ -852,6 +852,21 @@ def get_mutual_following(
 
     return mutual_following
 
+def isUserVerified(browser, username, logger):
```
elulcao: nit:
what about "is_user_verified( )"
Adding pydoc for arguments, 
logger.info should offer more information:
ie:
```python
logger.info(
            "Please enter a username to check the verified badge  ~leaving out of an invalid value"
)
```

instapy/instapy.py
```diff
@@ -4218,6 +4219,17 @@ def set_blacklist(self, enabled: bool, campaign: str):
         except Exception:
             self.logger.info("Campaign {} first run".format(campaign))
 
+    def isUserVerified(self, username: str = None):
```
elulcao: Already defined in relationship_tools.py

DOCUMENTATION.md
```diff
@@ -78,6 +78,7 @@
  <br />
 
 - **[Relationship tools](#relationship-tools)**
+  - [Is selected user is verified](#is-user-verified)
```
elulcao: nit:
Check if selected user has the verified blue badge](#is-user-verified)

elulcao (changes requested):

instapy/instapy.py
shifenis: Do you want to expose it just from `relationship_tools.py` instead of passing through `instapy.py`?

elulcao: Up to you :)
What I think is that the definition could be made in one place, and be called from other places as many times as necessary. Nested calls from the same definition sounds a bit fancy.

timgrossmann:
@shifenis Thank you very much for this addition!
Could you please format the changes you did using `black` once 😄 

@elulcao Thank you for doing the review!

timgrossmann:
@shifenis I think you need to format using py34 instead so the linting step for older python versions also passen 😊 

just do `black -t py34` instead

sorry for the inconveniences

shifenis:
@timgrossmann No problem  😄

elulcao (commented):
README.md file has been removed, 
it seems that you need to updated your branch to the latest one.

PR #5470, opened 2020-05-09
Check if user's account is private - Issue #5451

IreneArapogiorgi:
Hello, I was assigned to implement checking whether target user's account is private or not in order to proceed into finding his followers/followings. This will be a part of a university assignment of mine as I also mentioned on issue #5451.

About my implementation, I used [is_private_profile()](https://github.com/timgrossmann/InstaPy/blob/master/instapy/util.py#L65) to check if target user's account is private or not. In addition to that, I created a new file `follow_util.py` in which I moved the implementation of [get_following_status()](https://github.com/timgrossmann/InstaPy/blob/master/instapy/unfollow_util.py#L147) and then I imported it inside both unfollow_util.py and like_util.py, where it was already needed and used, and then inside relationship_tools.py in order to check whether the user is following the targeted private account, so in this case we can proceed into finding his followers and followings.

**Files added:**
follow_util.py

**Files changed:**
unfollow_util.py
relationship_tools.py
like_util.py
instapy.py
util.py

If there are any changes/improvements needed, please let me know so that I can work on them.
Also, any advice about unit testing my implementation is welcomed.

PR #5855, opened 2020-10-29
Check if user's account is private - Supplementary PR

IreneArapogiorgi:
This PR is supplementary to PR #5470. Due to several conflicts between InstaPy's current master branch and my own branch, which I was not able to resolve, I created a new one having the same code contribution that I've already analyzed in pull request #5470. A detailed description of the contribution and examples of usage are provided in that pull request.


instapy/follow_util.py
```diff
@@ -0,0 +1,110 @@
+""" Module which handles the follow features """
+
+from .util import web_address_navigator
```
elulcao, line 3: nit: Using the comment for the imports would be nice, ie: check instapy.py
```
# import built-in & third-party modules
# import InstaPy modules
# import exceptions
```

instapy/follow_util.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # check if the page is available
+    valid_page = is_page_available(browser, logger)
+    if not valid_page:
+        logger.warning(user_inaccessible_msg)
+        person_new = verify_username_by_id(
+            browser, username, person, None, logger, logfolder
+        )
+        if person_new:
+            web_address_navigator(browser, ig_homepage + person_new)
```
elulcao, line 42: ig_homepage`  is possibly unbound,
ig_homepage = "https://www.instagram.com/" needs to be defined before.

instapy/follow_util.py
```diff
@@ -0,0 +1,110 @@
...
+    followed"""
+
+    # try to find the user by ID
+    if person_id is None:
+        person_id = load_user_id(username, person, logger, logfolder)
+
+    if person_id and person_id not in [None, "unknown", "undefined"]:
+        # get the [new] username of the user from the stored user ID
+        person_new = get_username_from_id(browser, person_id, logger)
+        if person_new:
```
elulcao: one if-else statement can be removed, it only have to check if `if person_new != person:`
if `person_new` is None the report that `The user with the ID of '{}' is unreachable".format(person))`

ie:
```
if person_new != person:
...
else:
```

elulcao (changes requested):
Sharing some ideas,
thanks for considering my request.

instapy/follow_util.py
IreneArapogiorgi: @elulcao @timgrossmann Do you think we could address these two changes in a new Issue & PR, as they are not connected directly with the main purpose of this PR? And, in addition to that, so that we can check seperately that they work properly? Right now, in my branch, these two methods are the exact same as in upstream/master's unfollow_util.py file.

elulcao: sounds good to me, 
happy coding :)

IreneArapogiorgi: @elulcao Could you explain to me why the statement `if person_new:` should be removed? It seems to me that both of the if-else statements are needed.

elulcao: Hello @IreneArapogiorgi 
here is what I have in mind, use it like reference and
let me know your thoughts

```python
def verify_username_by_id(browser, username, person, logger, logfolder):
    """Check if the given user has changed username after the time of
    followed"""

    # try to find the user by ID
    # Please remove my comments
    # Checking again for the ID is much better than checking if the arg is None
    person_id = load_user_id(username, person, logger, logfolder)

    # If person_id is None, inform to InstaPy user that record does not exist 
    # and return,
    if person_id not in [None, "unknown", "undefined"]:
        # get the [new] username of the user from the stored user ID
        person_new = get_username_from_id(browser, person_id, logger)

        # Please remove my comments
        # Now we can check if person_new is not the same value as person,
        # if person_new is None then just inform to the InstaPy user that 
        # record does not exist.
        if person_new is not None and person_new != person:
            logger.info(
                "User '{}' has changed username and now is called '{}' :S".format(
                    person, person_new
                    )
            )
            return person_new
        
    # Check who call this def, since will receive a None value.
    logger.info("User '{}' doesn't exist in local records".format(person))
    return None
```

IreneArapogiorgi: @elulcao I totally agree with your perspective! Just updated the function and I think this PR is ready for merge @timgrossmann.

elulcao: @IreneArapogiorgi 
thanks!

elulcao (approved):
Thanks @IreneArapogiorgi

timgrossmann (approved):

timgrossmann:
@IreneArapogiorgi Looks good, thank you for the PR!

@elulcao Perfect, thanks for checking!

elulcao:
Hello @IreneArapogiorgi 

The txn was reverted, some issues were reported like #5914
it seems that extra argument `self_username` was causing a 
problem with current value `username` in some methods. 
Most of the problems comes from `unfollow_users` and 
cascade calls to other methods.

If you have a chance to take a look again to the txn it would 
be nice.

IreneArapogiorgi:
@elulcao Thank you for reaching out to me. I think I know how to fix this. I'll be working on it and will keep you posted.

PR #5858, opened 2020-10-31
fixed xpath selector for likers from photo

### LINKED ISSUE #5924: Can't follow likers
Within session.follow_likers(), the link to the correct photo is returned but the liked counter button cannot be found.

I thought this meant my xpath was incorrect so I changed it to "liked_counter_button": "//a[contains(@href,'liked_by/')]", but I'm still getting the same error.

My version of instapy is up-to-date. Is there something I'm missing? Is my xpath incorrect? Thanks for the help

markkdev:
Small fix to update the xpath for the liker button on photos.

### DIFF
--- a/instapy/xpath_compile.py
+++ b/instapy/xpath_compile.py
@@ -191,6 +191,6 @@ xpath["watch_story"] = {
 }
 
 xpath["likers_from_photo"] = {
-    "liked_counter_button": "//div/article/div[2]/section[2]/div/div/a",
-    "second_counter_button": "//div/article/div[2]/section[2]/div/div/button",
+    "liked_counter_button": "//div/article/div[3]/section[2]/div/div/a",
+    "second_counter_button": "//div/article/div[3]/section[2]/div/div/button",
 }


instapy/xpath_compile.py
```diff
@@ -191,6 +191,6 @@
 }
 
 xpath["likers_from_photo"] = {
-    "liked_counter_button": "//div/article/div[2]/section[2]/div/div/a",
-    "second_counter_button": "//div/article/div[2]/section[2]/div/div/button",
-}
+    "liked_counter_button": "//div/article/div[3]/section[2]/div/div/a",
```
elulcao, line 194: is `liked_counter_button` the same xpath like `//a[contains(@href,'liked_by')]/span`
do you think that using a generic could works here?

Second button could be `//a[contains(@href,'liked_by')]`

I believe that IG could change the xpaths levels very easily, so it would be a bit more stable to use generic definitions.

elulcao (changes requested):

instapy/xpath_compile.py
elulcao: Travis is not happy 😢 
Could you check your fix with `black -t py34` 👨‍💻

elulcao:
For the record:

Regarding this xpath,
there is another issue for videos, at the moment the number of views in a video
does not behave like the number of likes in a photo. Clicking on the number of
views only shows the same number in a floating window.
In logs we should see something like `Couldn't find liked counter button. May be a video.`
this is normal and nothing to be fixed by now.

I will push some changes here, to inform the changes.  InstaPy user will see the new messages.

* instapy.py: Added logger to get_photo_urls_from_profile(), users_liked(), extract_information()
* relationship_tools.py: Fixed return in get_followers()
* commenters_util.py: Converted print() to logger(), Some print() were redirected to STDOUT instead of the general.log
* browser.py: Updated Window size to Apple iPhone XS Max: 414, 896, Window size for some pics or videos is not enough.

elulcao:
Closing this PR,
changes were handled with #5929
👍

PR #5868, opened 2020-11-03
Add IGTV media type

epiphian:
To be able to like posts with IGTV


instapy/constants.py
```diff
@@ -2,4 +2,5 @@
 MEDIA_PHOTO = "Photo"
 MEDIA_CAROUSEL = "Carousel"
 MEDIA_VIDEO = "Video"
-MEDIA_ALL_TYPES = (MEDIA_PHOTO, MEDIA_CAROUSEL, MEDIA_VIDEO)
+MEDIA_IGTV = "IGTV"
```
elulcao, line 5: IGTV media depends of a xpath value, I checked quickly and it seems to me that they use the class `_5wCQW`, but currently `_5wCQW` is not used in code.
could you check in which def() this new constant will be used.

elulcao (changes requested):

Issue #5890, opened 2020-11-13
Couldn't follow 'Username'! ~user is inaccessible


v0lumehi:
## Expected Behavior
User is being followed
## Current Behavior
Error is being thrown: 
`Couldn't follow 'Username'! ~user is inaccessible`
It seems like selenium isn't able to get the button element. There was an attempted fix in https://github.com/timgrossmann/InstaPy/issues/5298, but that doesn't fix it for me. Maybe Instagram changed their code.
## Possible Solution (optional)
This was the presented solution in https://github.com/timgrossmann/InstaPy/issues/5298, but it's not working for me. The change was made in xpath_compile.py:
```
xpath["get_buttons_from_dialog"] = {
    "follow_button": "//div/button[text()='Follow']",
    "unfollow_button": "//div/button[text() = 'Following']",
}
```
## InstaPy configuration
Doesn't seem to be a config issue

v0lumehi:
It appears to work sometimes and sometimes not:
```
INFO [2020-11-13 13:07:15] Like# [6/95]
INFO [2020-11-13 13:07:15] https://www.instagram.com/p/CHiAL2Cjdp-/
INFO [2020-11-13 13:07:19] Image from: b'flame_keed'
INFO [2020-11-13 13:07:19] Link: b'https://www.instagram.com/p/CHiAL2Cjdp-/'
INFO [2020-11-13 13:07:19] Description: b'(FREE) KIZARU x YG WOODS Type Beat - Vibe (Prod. FLAME KEED) #kizaru #ygwoods #jabo #kizarutypebeat #borntotrap #jabotypebeat #kizarubeat #typebeat #freetypebeat #typebeatfree #rocket #vibebeat #trap #beat #beatfeee #freebeat #typebeat2020'
INFO [2020-11-13 13:07:24] User: 'flame_keed' |> followers: 1775 |> following: 2794 |> relationship ratio: 0.63
INFO [2020-11-13 13:07:54] --> Found 'Play' button for a video, traying to like it
INFO [2020-11-13 13:07:56] --> Video liked!
INFO [2020-11-13 13:08:01] --> Not commented
INFO [2020-11-13 13:08:53] Last follow is verified after reloading the page!
INFO [2020-11-13 13:08:53] --> Followed 'b'flame_keed''!
INFO [2020-11-13 13:08:53] Attempting to find user ID: Track: post, Username flame_keed
[...]

PR #5893, opened 2020-11-13
Fix for Issue #5890 

v0lumehi:
Problem: Couldn't follow 'Username'! ~user is inaccessible 
Solution: introduced new method that clicks on the user profile button in the comment page


v0lumehi:
I messed up sorry

PR #5896, opened 2020-11-14
#5895 - possible fix for incorrect xpath used by interact_by_comments

### LINKED ISSUE #5895: [BUG] Invalid xpath causes interact_by_comments to land in non exception else statement.
## Expected Behavior
When using the `interact_by_comments` method, the app should iterate on a post's comments identifying the like and dislike buttons in search for replies.
## Current Behavior
The current logic gap causes the app to not be able to identify the like/dislike button hence failing to the else statement located at https://github.com/timgrossmann/InstaPy/blob/master/instapy/comment_util.py#L314-L315
## Possible Solution (optional)
Fix the xpath in https://github.com/timgrossmann/InstaPy/blob/master/instapy/xpath_compile.py#L79-L80 to the new paths.
## InstaPy configuration
example [here](https://instapy.org/actions/#usage) :
```
session.set_use_meaningcloud(enabled=True, license_key='', polarity="P")
session.set_use_yandex(enabled=True, API_key='', match_language=True, language_code="en")

session.set_do_comment(enabled=True, percentage=14)
session.set_comment_replies(replies=[u"😎😎😎", u"😁😁😁😁😁😁😁💪🏼"], media="Photo")

session.set_user_interact(amount=2, percentage=70, randomize=False, media="Photo")
session.set_do_like(enabled=True, percentage=100)

session.interact_by_comments(usernames=["somebody", "other.buddy"], posts_amount=10, comments_per_post=5, reply=True, interact=True, randomize=True, media="Photo")```

### PR DESCRIPTION
**Pull request by luismsousa**
it seems that this path is not valid, causing the logic in https://github.com/timgrossmann/InstaPy/blob/master/instapy/comment_util.py#L237 to lead the following test https://github.com/timgrossmann/InstaPy/blob/master/instapy/comment_util.py#L279 to an else statement in https://github.com/timgrossmann/InstaPy/blob/master/instapy/comment_util.py#L305. 
because the unlike button xpath is also incorrect, the program ends up printing this message: https://github.com/timgrossmann/InstaPy/blob/master/instapy/comment_util.py#L314

### DIFF
--- a/instapy/xpath_compile.py
+++ b/instapy/xpath_compile.py
@@ -76,8 +76,8 @@ xpath["get_comment_input"] = {
 xpath["get_comments_on_post"] = {
     "commenter_elem": "//h3/a",
     "comments_block": "//div/div/h3/../../../..",
-    "like_button_full_XPath": "//div/span/div/button/div/span[*[local-name()='svg']/@aria-label='Like']",
-    "unlike_button_full_XPath": "//div/span/div/button/div/span[*[local-name()='svg']/@aria-label='Unlike']",
+    "like_button_full_XPath": "//*[*[local-name()='svg']/@aria-label='Like']",
+    "unlike_button_full_XPath": "//*[*[local-name()='svg']/@aria-label='Unlike']",
 }
 
 xpath["get_cord_location"] = {"json_text": "//body"}
--- a/instapy/xpath_compile.py
+++ b/instapy/xpath_compile.py
@@ -76,8 +76,8 @@ xpath["get_comment_input"] = {
 xpath["get_comments_on_post"] = {
     "commenter_elem": "//h3/a",
     "comments_block": "//div/div/h3/../../../..",
-    "like_button_full_XPath": "//div/span/button/span[@aria-label='Like']",
-    "unlike_button_full_XPath": "//div/span/button/span[@aria-label='Unlike']",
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: instapy/xpath_compile.py [hunk 1]
```diff
@@ -76,8 +76,8 @@
 xpath["get_comments_on_post"] = {
     "commenter_elem": "//h3/a",
     "comments_block": "//div/div/h3/../../../..",
-    "like_button_full_XPath": "//div/span/button/span[@aria-label='Like']",
-    "unlike_button_full_XPath": "//div/span/button/span[@aria-label='Unlike']",
+    "like_button_full_XPath": "//div/span/div/button/div/span[*[local-name()='svg']/@aria-label='Like']",
```
elulcao: Thats correct, the xpath is not the correct one,
could you check If the new xpath proposed can be reduced a with a wild card,
`//div/span/div/button/div/` seems to be a valid expression for reduction.

IG could change the xpath again adding or reducing levels, so we should pursue more generic xpaths.
ie:
`//*[*[local-name()='svg']/@aria-label='Like']`

check format with `black -t py34`

elulcao (changes requested):

instapy/xpath_compile.py
luismsousa: Edit made.

elulcao (approved):

breuerfelix (approved):

PR #5918, opened 2020-11-20
Check if user's account is private - Fixed conflicts

IreneArapogiorgi:
This PR contains all changes of PR #5855 that was merged, but then reversed due to conflicts with the `unfollow_users` method. This conflict has been fixed in this new PR.

**To explain the use of the `self_username` parameter in methods `get_followers`, `get_following` etc:**

_Before --_ These methods grabbed followers/following/nonfollowers etc of a given ig account without checking whether we follow that account in order to reach its data.
_Now --_ These methods now check whether the targeted account is private or not in order to proceed.

When both parameters `self_username` and `username` are found, `self_username` stands for our personal ig account, while `username` for the targeted one. This is how they've been used inside all these methods (`get_followers`, `get_nonfollowers` etc) to find whether session's ig account is following the targeted account or not.

**The conflict we missed:**

Inside file `unfollow_util.py`, a specific method called `unfollow_users` did not pass the `self_username` parameter when calling methods `get_followers`, `get_following` and `get_nonfollowers`. I fixed that by parsing the session's ig username instead to correspond to `self_username` parameter. Afterall, the `unfollow_users` method aims to reach followers of the user's personal account which is both the session's account (self_username) and the targeted account (username).

Hope everything makes sense to you. For any more information, please comment below.
@elulcao @timgrossmann


instapy/unfollow_util.py
```diff
@@ -251,7 +251,14 @@ def unfollow(
 
             # Unfollow only the users who do not follow you back
             unfollow_list = get_nonfollowers(
-                browser, username, relationship_data, False, True, logger, logfolder
+                browser,
```
elulcao, line 254: The problem I see is the following,

A user want to unfollow users, then they create the object like:
```python
session.unfollow_users(
        amount=5,
        nonFollowers=False,
        instapy_followed_enabled=True,
        instapy_followed_param="nonfollowers",  <== This is the variation that cause issues
        style="FIFO",
        unfollow_after=90 * 60 * 60,
        sleep_delay=60,
    )
```

`unfollow_users(...)` will call `unfollow()` with only these parameters; see below. 
```python
try:
            unfollowed = unfollow(
                self.browser,
                self.username,
                amount,
                (custom_list_enabled, custom_list, custom_list_param),
                (instapy_followed_enabled, instapy_followed_param),
                nonFollowers,
                allFollowing,
                style,
                self.automatedFollowedPool,
                self.relationship_data,
                self.dont_include,
                self.white_list,
                sleep_delay,
                self.jumps,
                delay_followbackers,
                self.logger,
                self.logfolder,
            )
``` 

The most important to keep in mind is `self.username` because latter on the other methods will call it (ie `get_followers()`). 
Then, they are two args used, arguments like `self_username` and `username`. So, this is the issue, since both variables were mixed in some point; the old logic will use the args mixed.

Using the new two args `self_username ` and `username ` in `def get_followers()` you will notice that now 
`self_username` becomes `username` and `username` becomes empty.

```
 251 INFO [2020-11-18 12:54:21] [internet]  Unfollowing the users followed by InstaPy
 252
 **253 INFO [2020-11-18 12:54:21] [internet]  Retrieving live 'Followers' data of None at "full" range
 254 ERROR [2020-11-18 12:55:13] [internet]  Sorry, an error occurred: 'NoneType' object has no attribute 'encode'** <== HERE 
 255 INFO [2020-11-18 12:55:14] [internet]  Sessional Live Report:
 256         |> No any statistics to show
```

So, just adding the new argument `self_username` is not enough because the positional arguments lost meaning 
for the current logic of the code. 

I verified the txn, the reverted one, and the methods compromised returns the same error reported in 5914, where `logfolder` lost their context due to the new argument. 

```bash
ERROR [2020-11-18 15:13:35] [xxxxxxxxx] Sorry, an error occurred: get_followers() missing 1 required positional argument: 'logfolder'
```

Check the logic for the following, to determine if the new `self_username` does not leave the old one `username` on blank:

```bash
get_followers(
get_following(
get_unfollowers(
get_nonfollowers(
get_mutual_following(
get_fans(
get_mutual_following(
```

elulcao (changes requested):

instapy/unfollow_util.py
IreneArapogiorgi: I think that this exact issue with the missing argument is now fixed by the changes I applied in this PR. I tested the `session.unfollow_users()` method and everything seems fine. Do you mind running my code by yourself as well, in order to confirm that?

elulcao: sure no problem,
Could you add the changes reverted again into this PR, 
this one is only having one part

IreneArapogiorgi: I'm not sure how to do that, however my entire implementation is included in [this branch](https://github.com/IreneArapogiorgi/InstaPy/tree/private_account_check), so you just have to [git clone this specific one](https://stackoverflow.com/questions/1911109/how-do-i-clone-a-specific-git-branch/4568323#4568323).

In case of merging, I think that the entire implementation of the [private_account_check](https://github.com/IreneArapogiorgi/InstaPy/tree/private_account_check) branch will be included in InstaPy's master branch, not just the latest fixes of this PR.

elulcao: You can try fetching the changes from your local, then send again the PR using this one. This is better to have all your changes in one transaction.

elulcao: Let me try the changes I saved in my local plus this new one, 
I will let you know my branch, then you can fetch from it.

elulcao:
Hello @IreneArapogiorgi 

Pushed the changes to this PR, could you kindly confirm all of them are the correct ones.
I just fixed the typo from `dissmiss_save_information` to `dismiss_save_information`. 
👍

IreneArapogiorgi:
Hello @elulcao 

Everything seems fine to me! Did you tested the functionality of the `session.unfollow_users()` method now with the current changes as well?

elulcao:
Hello @IreneArapogiorgi 
used `session.unfollow_users()` with `all` and `nonfollowers` 
I didn't see any issue

```python
    session.unfollow_users(
        amount=5,
        nonFollowers=False,
        instapy_followed_enabled=True,
        instapy_followed_param="nonfollowers",
        style="FIFO",
        unfollow_after=90 * 60 * 60,
        sleep_delay=60,
    )
```

elulcao (approved):

IreneArapogiorgi:
@elulcao that's great! I didn't have any issues while testing the code either. So, it seems like it's ready for merge.

PR #5929, opened 2020-11-27
PR: Update xpath for `likers_from_photo`, `get_followers` return for empty list, windows size to `iPhone XS Max: 414, 896`

### LINKED ISSUE #5924: Can't follow likers

elulcao:
Hello all

There is another issue for videos, at the moment the number of views in a video
does not behave like the number of likes in a photo. Clicking on the number of
views only shows the same number in a floating window.
In logs we should see something like `Couldn't find liked counter button. May be a video.`
this is normal and nothing to be fixed by now.

*instapy.py: Added logger to get_photo_urls_from_profile(), users_liked(), extract_information(). This is because new logger arg in `commenters_util.py `
*relationship_tools.py: Fixed return in get_followers(), it was returning False instead of a list. 
*commenters_util.py: Converted print() to logger(), Some print() were redirected to STDOUT instead of the general.log making hard to search failures.
*browser.py: Updated Window size to Apple iPhone XS Max: 414, 896, Window size for some pics or videos is not enough, with this change, the new image and video formats will be displayed on the screen allowing elements to be found without problem.
*xpath_compile.py: Updated likers_from_photo xpath, generic form.


timgrossmann (approved):

timgrossmann:
@elulcao Love the comments and the updates of you, thank you! 😊

Issue #5932, opened 2020-11-30
Commenting with '@{}' inserts my own username instead of the target's username


sharmanshah:
## Expected Behavior
@{} should insert the target's username in comments

## Current Behavior
@{} inserts my own username

## InstaPy configuration
0.6.12 - latest unreleaded

elulcao:
Comments under some circumstance were posted with InstaPy user instead of target user.
Now, comments are using target user when following format is used`@{}` 
Next PR will propagate the changes.

```bash
    6 INFO [2020-11-30 22:52:20] [internet]  Location: b'Destiny'
    5 INFO [2020-11-30 22:52:23] [internet]  User: 'target'  |> followers: 103133  |> following: 2  |> relationship ratio: 51566.5
    4 INFO [2020-11-30 22:53:18] [internet]  --> Image liked!
    3 INFO [2020-11-30 22:53:37] [internet]  --> Commented: b'@target I like your stuff, so giving a like '
    2 INFO [2020-11-30 22:53:39] [internet]  --> Not following
    1 INFO [2020-11-30 22:53:40] [internet]  Like# [13/14]
```

PR #5938, opened 2020-12-03
Problem: Added Two Factor Authentication feature, intended for protected

### LINKED ISSUE #2118: [request] Two-step authentication
Hey guys, so I'm thinking that it would be a great security improvement to allow us using two-step autentication with InstaPy. As of today, if you have this option on, instapy fails. However, since instagram gives us a few security codes, wouldn't it be possible to feed instapy those codes so that it could work having this feature enabled?

Thanks! :)

elulcao:
Hello all,
Hope you doing well,

Added Two Factor Authentication feature, intended for protected
accts. If the acct is not protected this feature will be ignored and will
be unnoticed. This feature basically should be used to bypass the 
extra security configuration used by the InstaPy user.

- comment_util.py: replaced unused variables in process_comments()
- instapy.py: added `security_codes` arg in constructor and `login_user()`
            `process_comments()` now uses `user_name` for commenting a
            target user; #5932
            `comment_by_locations()` now has the same code for `process_comments()`
- login_util.py: Added random import due to login_user() feature with Two Factor
               Authentication. `security_codes`, must be a list with codes
               and InstaPy will chose only one element randomly.
               Added new `def two_factor_authentication()`
- xpath_compile.py: Updated `get_comments_on_post` xpath with a generic form
- Added `pull_request_template.md`: to have a new format in the PR.
- instance-settings.md updated with steps for Two Factor Authentication
- Updated CHANGELOG with commit changes

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,26 @@
+<!-- Did you know that we have a Discord channel ? Join us: https://discord.gg/FDETsht -->
+<!-- Is this a Feature Request ? Please, check out our Wiki first https://github.com/timgrossmann/InstaPy/wiki -->
+# Pull Request Template
+
+## Description
+
+Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.
+
+Do not include any personal data.
+
+Fixes # (issue)
+
+## How Has This Been Tested?
+
+Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration.
+
+- [ ] Test
+
+## Checklist:
+
+- [ ] I have commented my code, particularly in hard-to-understand areas
+- [ ] I have made corresponding changes to the documentation
+- [ ] I have checked my code and corrected any misspellings
+- [ ] I have performed a self-review of my own code
+- [ ] My code follows the style guidelines of this project, `black -t py34`
+- [ ] My changes generate no new warnings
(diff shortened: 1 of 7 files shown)


timgrossmann (approved):
Looks really good again!
Also. love the clear explanations on the PRs.

Thank you very much for this update 😍

breuerfelix:
i just got one question to this:  
isn't the instagram two factor auth just like google auth ? when you login you have to look on your smartphone and get a time based token you have 30 seconds time to enter ?  
if yes, then isnt it impossible to know these beforehand ?

i could be totally wrong since i never used two factor auth at instagram but i am just curious :D  
cause if it is like that, it would be better to prompt the user for the code and not make him provide it via constructor :)

timgrossmann:
> i just got one question to this:
> isn't the instagram two factor auth just like google auth ? when you login you have to look on your smartphone and get a time based token you have 30 seconds time to enter ?
> if yes, then isnt it impossible to know these beforehand ?
> 
> i could be totally wrong since i never used two factor auth at instagram but i am just curious :D
> cause if it is like that, it would be better to prompt the user for the code and not make him provide it via constructor :)

Looks like one can use the recovery codes to have a fixed value for this: 
Security codes can be found in: `Settings` -> `Security` -> `Two-Factor-Authentication` -> `Backup Codes`

breuerfelix:
> > i just got one question to this:
> > isn't the instagram two factor auth just like google auth ? when you login you have to look on your smartphone and get a time based token you have 30 seconds time to enter ?
> > if yes, then isnt it impossible to know these beforehand ?
> > i could be totally wrong since i never used two factor auth at instagram but i am just curious :D
> > cause if it is like that, it would be better to prompt the user for the code and not make him provide it via constructor :)
> 
> Looks like one can use the recovery codes to have a fixed value for this:
> Security codes can be found in: `Settings` -> `Security` -> `Two-Factor-Authentication` -> `Backup Codes`

awesome! thanks for the clarification :) let's merge it!

PR #6109, opened 2021-03-07
Add verified_only flag to get_followers()

pschroeder89:
## Description

I use this package to pull verified users for my business accounts' followers. This seems to be a common use case, so I figured I'd add these changes to the main repo.

## How Has This Been Tested?

The following command now returns only verified users for username:
```
followers_list = session.grab_followers(username=targetAccount, amount=amount, live_match=True, store_locally=False, verified_only=True)
```

Checked for 5 different accounts, all worked.

## Checklist:

- [X ] I have commented my code, particularly in hard-to-understand areas
- [X ] I have made corresponding changes to the documentation
- [X ] I have checked my code and corrected any misspellings
- [X ] I have performed a self-review of my own code
- [X ] My code follows the style guidelines of this project, `black -t py34`
- [X ] My changes generate no new warnings

### DIFF
--- a/instapy/relationship_tools.py
+++ b/instapy/relationship_tools.py
@@ -186,7 +186,6 @@ def get_followers(
             # get followers
             page_info = data["user"]["edge_followed_by"]["page_info"]
             edges = data["user"]["edge_followed_by"]["edges"]
-
             for user in edges:
                 if verified_only:
                     if user["node"]["is_verified"]:
--- a/instapy/relationship_tools.py
+++ b/instapy/relationship_tools.py
@@ -30,7 +30,7 @@ def get_followers(
     store_locally,
     logger,
     logfolder,
-    verified_only=False
+    verified_only=False,
 ):
     """ Get entire list of followers using graphql queries. """
 
@@ -318,6 +318,7 @@ def get_followers(
         return verified_followers
     else:
         return all_followers
+
 
 def get_following(
     browser,
--- a/instapy/relationship_tools.py
+++ b/instapy/relationship_tools.py
@@ -38,7 +38,6 @@ def get_followers(
     user_data = {}
     variables = {}
     all_followers = []
-    verified_followers = []
     sc_rolled = 0
     grab_notifier = False
     local_read_failure = False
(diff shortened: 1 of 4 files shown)


instapy/relationship_tools.py
```diff
@@ -185,6 +187,11 @@ def get_followers(
             page_info = data["user"]["edge_followed_by"]["page_info"]
             edges = data["user"]["edge_followed_by"]["edges"]
             for user in edges:
+                # If verified_only is True, determine if following user is verified and append to verified_followers
+                if verified_only:
+                    if user["node"]["is_verified"]:
```
elulcao, line 191: nit: could be squashed this double `if` in a single one?

elulcao (approved):
Hello @pschroeder89 
Looks good to me, I added @breuerfelix as reviewer.
Just one thing, could you update `relationship-tools.md` to reflect this change,
Thanks 👍

instapy/relationship_tools.py
pschroeder89: done!

breuerfelix:
yeah some notes in the readme would be awesome and also in the changelog :)  
also why can't we just use all_followers ? i think it would be better to use this list instead of the verified_list. cause if you pass in true, you just want verified ones.

@elulcao what about we merge this to a dev branch so we can add stuff like changelog and readme ourselfes before merging to master :)

pschroeder89:
> yeah some notes in the readme would be awesome and also in the changelog :)
> also why can't we just use all_followers ? i think it would be better to use this list instead of the verified_list. cause if you pass in true, you just want verified ones.
> 
> @elulcao what about we merge this to a dev branch so we can add stuff like changelog and readme ourselfes before merging to master :)

Ok we now use all_followers to store just the verified users if verified_only is True. 
Also updated the readme and changelog.

instapy/relationship_tools.py
pschroeder89: Needed to break this up again since the logic is a bit different now

breuerfelix (approved):
awesome! i guess it is better with all_followers since there is a lot of logic afterwards aswell. now things won't break in some weird scenarios :) thanks for the PR !

elulcao:
> yeah some notes in the readme would be awesome and also in the changelog :)
> also why can't we just use all_followers ? i think it would be better to use this list instead of the verified_list. cause if you pass in true, you just want verified ones.
> 
> @elulcao what about we merge this to a dev branch so we can add stuff like changelog and readme ourselfes before merging to master :)

Yeah, having a Dev Branch would be better for users. After a while we can push more fixes into the main branch. For the Dev branch, let me know if you have a chance to create it in the repo.

Issue #6134, opened 2021-03-29
Sometime IG block InstaPy on Likes


Tolmino:
## Current Behavior

When is detected a block on Likes, InstaPy continue to try to like until IG block the account asking a security code 

## Possible Solution (optional)

pause InstaPy

GitNees:
Is this bot still working? I have stopped using it for years as instagram detects it to quickly. I just came back to see if its still the case.

Im actually using another bot that does that what you want. https://github.com/GramAddict 
It actually does everything on your phone, so its harder to detect (needs a real android phone or emulator) But has some great features!

Tolmino:
Yes ... that bot is sill working

mydani:
Used it once with the quickstart settings which are meant to be easy for 1st-timers. Really conservative.
Took about 5 minutes to get blocked for a day.
Retried next time with minutes between actions - blocked after 2nd action.
So seems IG is good at detecting Instapy.

PS: did not use any bot before.

MarcoLavoro:
Any news on this?
I still have a problem that block my account

pedromoto4:
Here the same thing, IG block with very low InstaPy activity  :/

PR #6135, opened 2021-03-29
Pause Like_Util.py

Tolmino:
## Description

Pause Like Util when a BLOCK on Likes is detected
so IG don't Block account

Fixes # (issue)

https://github.com/timgrossmann/InstaPy/issues/6134

## How Has This Been Tested?

Yes

## Checklist:

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have checked my code and corrected any misspellings
- [x] I have performed a self-review of my own code
- [ ] My code follows the style guidelines of this project, `black -t py34`
- [x] My changes generate no new warnings

### DIFF
--- a/instapy/like_util.py
+++ b/instapy/like_util.py
@@ -818,7 +818,8 @@ def like_image(browser, username, blacklist, logger, logfolder, total_liked_img)
             logger.info(
                 "--> {} was not able to get liked! maybe blocked?".format(media)
             )
-            sleep(120)
+            input("******** Press Enter to continue ********")
+            sleep(10)
 
     else:
         liked_elem = browser.find_elements_by_xpath(unlike_xpath)


ErezAmihud:
Adding input function in a library is not a good idea at all.
There are automations that relay on this bot library, and by placing input you will make them crush/be unresponsive.
example: https://github.com/breuerfelix/instapy-gui

fintzd (approved):
LGTM, thought this changes the actions pattern from automated, and implements a need to manually check for the next step (hitting enter), so I'm not sure about what others think about this. 

Maybe increasing the sleep time would be better? (at least if the aim is to keep the script more automated)

breuerfelix:
using input here is a really bad ideas as @ErezAmihud already said since it would destroy any usecase in an automated process.

instead of using the input you could pass a callback function to instapy which gets called here. so users can (if they want to) provide and function they want in case they get blocked. for automated scripts we could pass a sys.exit and for manual approach we could use the input function.  
also this wouldnt change the current process :)

any ideas?

breuerfelix (changes requested):
allow users to pass a callback for this functionality

PR #6316, opened 2021-09-17
Feature/force firefox esr

ErezAmihud:
## Description

This pull request bring to light the solution to the `Hide Selenium Extension: error` problem.
The solution is installing firefox-esr instead of firefox, but for new users there is no description or any indication.

This code does 2 things:
1. Try to find and use the `firefox-esr` installation instead of the regular one.
2. Add description about how to install and what to do if there is already an installation.

!!!!!!!!!!! NOTE !!!!!!!!!!!!! There are 3 things that I don't know how to do:
1. where the install page is located to add a like to install firefox there.
2. Where to put the docs file
3. What to change "NO IDEA WHAT LINK I SHOULD PUT HERE" with

## How Has This Been Tested?

On windows and linux:

- [X] Run without firefox-esr installed
- [X] Run with firefox-esr installed
- [X] Run with firefox-esr from the `browser_executable_path` variable

## Checklist:

- [X] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [X] I have checked my code and corrected any misspellings
- [X] I have performed a self-review of my own code
- [X] My code follows the style guidelines of this project, `black -t py34`
- [X] My changes generate no new warnings


elulcao:
@ErezAmihud 

1. What will happen when FFESR got deprecated? pinning down FF is not a final solution, instead of that we need to fix the real problem for the extension. Im curious about the "Hide Selenium Extension: error" you are mentioning, could you please elaborate. 

2. Did you check `browser_executable_path` InstaPy already have the flag for custom FF and gecko driver. NO other changes are needed for custom paths in Windows, nor macOS or linux.


```
### Choose the browser version
If you have more than one Firefox version on your system or if you are using a portable version you can instruct InstaPy to use that version using the `browser_executable_path` argument in the class initializer.

Specifying the Firefox executable path can also help you if you are getting the following error message:

`selenium.common.exceptions.SessionNotCreatedException: Message: Unable to find a matching set of capabilities`

Example on a Windows machine (with the right path also works on Linux and macOS)

```python
                  password=insta_password,
                  browser_executable_path=r"D:\Program Files\Mozilla Firefox\firefox.exe")
```

ErezAmihud:
@elulcao 
1. Ok so by default, there is evidence that the browser is using selenium to operate. Those evidence make instagram block users/ask for captchas/phone confirmation. The check that makes this error appear is that `window.navigator.webdriver` in the browser. Basically all the extension does is changing this property and a few other to return other values ([good example is 3rd line here](https://github.com/timgrossmann/InstaPy/blob/master/instapy/firefox_extension/content.js)).

FFESR is the firefox extended support versions - which was designed specifically for applications that run on a browsers but cannot take the time to support every change new versions will throw on you, so they update only for major changes. We are those exact people, before you pushed updates yesterday (appreciate it a lot by the way) the project was not touched for 2-3 months, while there were errors that were fixed in unmerged prs.

Right now, we are running after firefox updates to fix this error and possible new errors that appear, while all the users are suffering. If we switch to the ffesr we need to handle only the major changes of firefox and only once every few months or more (firefox 52 was released in 2017 and still works good).

Beside that, every user wants a smooth experience, they want to install the bot, run the bot, enjoy the bot. They don't want to sit for a few hours just to find a solution to a problem they don't know is a problem (because failing to hide selenium is a warning). Especially because there are ways for ffesr to live with regular firefox on the same computer (portable version-windows, install package - linux)
As an example of how hard it is, I actually found this was the solution after more then 15 hours of developing to instapy-gui and because I accidently opened the instapy docker.

Also, thinking about reproducing problems on the local machine - it can be tricky to do if the browser version is unknown.

[Example where it helped](https://discord.com/channels/510385886869979136/510386000330096640/888697111422599168)

2. As for the second point, the main point of this pr is telling the user he needs to install a firefox that is compatible with instapy. I am using the parameter (and the behavior of using this parameter is not changed), but the main problem is letting the user know what the hell to do with it. A good example of the average user is someone I helped, I tried to explain to him that 2 bots cannot run on the same workdir, and sent him a line to fix it, and it tool more then 1 hour for him to find it.

The sooner we make instapy stable for everyone, the more people using, and the more people contributing.

elulcao:
@ErezAmihud 
Kept it simple, enhance the documentation to use FF version lower than 88 to avoid the issue with the bot detection. As you mentioned the one is working is the ESR. Current code is able to handled custom FF and geckodriver location, and should not be changed; btw, it seems that the change proposed is more for windows users a general change.

ErezAmihud:
@elulcao
Ok so I removed most of the code, except for one line that tells the user the error is because of low firefox version and the user can check the docs for more data.
I insist on the line because it will save a lot of trouble for a lot of people

breuerfelix:
@ErezAmihud thanks for the doc update! this will help a lot :). 
one questions tho: why did you remove all the spaces at the line endings? those mean "add a newline" in markdown and without them, everything will be rendered in one line :)

i hate this markdown syntax because its kinda hidden (just a few people show invisible chars in their editor) but it is what it is.

have a look at your readme in a markdown preview window and you will see what i mean :) maybe you can try to readd all those spaces at the end and we can merge it!

ErezAmihud:
Closed because It was written 3 years ago, I thought it was merged, I don't know if this is still relevant.

PR #6317, opened 2021-09-17
Feature/add message to follow success

ErezAmihud:
When following someone there is a chance the `Message` will be the thing appearing.
This should fix it.

https://cdn.discordapp.com/attachments/510386000330096640/888568967747223592/unknown.png


breuerfelix (approved):
LGTM

PR #6356, opened 2021-10-15
switch from travis to github actions

breuerfelix:
this should reflect all tests from the old travis pipeline.  
since they are down somehow, the switch is needed in order to publish new versions to pypi :)

the problem: i can't really test them since some only triggers on master and some only trigger once they are merged ... a little weird but yeah, i wrote a lot of em already, they SHOULD work :D
Signed-off-by: Felix Breuer <fbreuer@pm.me>


elulcao (approved):
👍

elulcao:
Merging, 
Next changes will be using this pipeline

